PR #1, opened 2020-12-16
Suggestion/Request: add husky + lint-staged

StanislavSt:
A suggestion/request to add [husky](https://github.com/typicode/husky) and [lint-staged](https://github.com/okonet/lint-staged) to the project.

I have added:

- husky + lint-staged to  **devDependencies**.
- husky pre-commit hook that runs `lint-staged`

### Type of change
- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,9 +32,23 @@
     "eslint-config-prettier": "^7.0.0",
     "eslint-plugin-prettier": "^3.3.0",
     "eslint-plugin-react": "^7.21.5",
+    "husky": "^4.3.6",
+    "lint-staged": "^10.5.3",
     "prettier": "^2.2.1",
     "stylelint": "^13.8.0",
     "stylelint-config-sass-guidelines": "^7.1.0",
     "typescript": "^4.1.3"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{ts,tsx,js}": [
+      "yarn run lint",
+      "yarn run format"
+    ],
+    "*.scss": "yarn run lint:styles"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -32,9 +32,23 @@
...
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{ts,tsx,js}": [
+      "yarn run lint",
+      "yarn run format"
```
kachar, line 50: Awesome integration! :rocket: 

Do you mind adding `yarn type-check` as well to the pre-commit hook as the integrity of the ts files is not verified at the moment.

While the most simple approach would be to add it to the list of `lint-staged` it seems a bit hard as the `tsc --project xxx` option will run against the `tsconfig.json` files instead of the passed staged files.

I've tried that and got an error:
```
error TS5042: Option 'project' cannot be mixed with source files on a command line.
```

You can find more info in https://github.com/okonet/lint-staged/issues/468

---

Another approach might be to include it in the husky command, with the downside of running it even for scss files.

```json
"pre-commit": "yarn type-check && yarn lint-staged"
``` 

What do you think about all that?

kachar (changes requested):
@StanislavSt Thanks for the contribution! :bow: 

Please review the comment on the ts checks

package.json
kachar: On other hand we have a dedicated workflow to build the ts/tsx files. Running that on commit might be overkill

kachar (approved):
No changes needed, sorry for the noise around.

Thanks again for the contribution!

StanislavSt:
Thank you for updating the readme :)_

Issue #3, opened 2020-12-17
Add i18n support


kachar:
**Viable options:**

* https://lingui.js.org/
* https://github.com/flayyer/next-rosetta/ (https://github.com/lukeed/rosetta)
* https://www.i18next.com/ but keep in mind https://github.com/isaachinman/next-i18next/issues/869

---

https://ora.pm/project/277933/kanban/task/4480228

PR #9, opened 2020-12-20
Add i18n support

StanislavSt:
## Problem
_What problem are you trying to solve?_

Add support for i18n, that works with Next.js built-in support for internationalized (i18n) routing (v10.0.0+)
- github - #3 
- ora.pm - https://ora.pm/project/277933/kanban/task/4480228

## Solution
A simple solution using only `i18next` and `react-i18next` in combination with the new build-in support from next.js for [internationalized routing](https://nextjs.org/docs/advanced-features/i18n-routing).

Current implemented strategy is [sub-path-routing](https://nextjs.org/docs/advanced-features/i18n-routing#sub-path-routing)

**No HOC is being used,** a `i18next` instance is being initialized in the custom `_app` function. This allows for the use of [useTranslation](https://react.i18next.com/latest/usetranslation-hook) hook

Multiple namespaces are supported:
- If no namespace is specified for the page , only the default one will be loaded `common.json`
- If a page requires specific namespaces, they will have to be passed using `getStaticProps`.
- Only the namespaces required by the page will be loaded.

## Motivation and Context
The only thing that is not being handled by `next.js` at the moment is the dynamic namespace loading. I have used the solution in this comment https://github.com/isaachinman/next-i18next/issues/869#issuecomment-746762163 in the PR. Which I believe is more than enough for the current state of the project.

## How Has This Been Tested?
Manually, directly in the application.
```shell
yarn type-check
yarn build
```
**New dependencies**:
- `i18next` : https://github.com/i18next/i18next
- `react-i18next` : https://github.com/i18next/react-i18next

## Screenshots (if appropriate):
image

image


StanislavSt:
First time implementing i18n with next.js. Any suggestions or more elegant solutions are greatly appreciated.
A language switch component will be added in a separate PR if the current solution is approved.

src/i18n.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  if (router.locale && i18next.language !== router.locale) {
+    i18next.changeLanguage(router.locale)
+  }
+
+  for (const ns of namespaces) {
+    if (router.locale && !i18next.hasResourceBundle(router.locale, ns)) {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const data = require(`../public/locales/${router.locale}/${ns}.json`)
```
StanislavSt, line 29: Could not make this work with `await import()` :(

kachar (approved):
@StanislavSt Thanks for the contribution! 
After the last addition I think we're good to go :rocket: 

Works perfectly!

PR #16, opened 2021-01-24
Fix timeout issue on yarn install

Nikola-Andreev:
## Fix timeout issue on yarn install

Because of the `@material-ui/icons` package size, the downloading could fail with:
`There appears to be trouble with your network connection. Retrying...`

## Solution

Increase network-timeout
https://github.com/mui-org/material-ui/issues/12432


kachar (approved):
I've never experienced the problem personally and I hope after the fix that nobody will.

Thanks for the research and contribution!

Issue #26, opened 2021-02-03
Integration of Snackbar + MobX store POC


kachar:
We need a POC on the integration of MobX state store and Snackbar component

Sample approach: https://medium.com/@kofno/snackbar-logic-in-mobx-a27ab504fde1

- https://mobx.js.org/
- https://material-ui.com/components/snackbars/

PR #27, opened 2021-02-04
#26 Integration of Snackbar with MobX - POC

StanislavSt:
## Motivation and context
This is a POC , to see if MobX would be suitable for displaying a Snackbar.

- Create Mobx store for Alerts/Snackbars
- Add `Snackbar` to `Layout`.
- Display a snackbar from anywhere in the application using the AlertStore.

## Screenshots:
Feb-04-2021 22-33-08


src/stores/AlertStore.ts
```diff
@@ -0,0 +1,53 @@
...
+      show: true,
+      message,
+      type,
+      duration,
+    }
+    this.alerts.push(alert)
+  }
+
+  hide() {
+    if (this.alerts.length > 0) this.alerts[0].show = false
```
kachar, line 41: We might pass an `id` to the hide method, so we hide a specific alert, instead of the first one in the list.

```suggestion
  hide(id: number) {
    if (this.alerts.length > 0) this.alerts[id].show = false
```

src/components/layout/Snackbar.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <>
+      {getAlerts.map(({ id, show, duration, type, message }) => {
+        return (
+          <Snackbar
+            key={id}
+            open={show}
+            autoHideDuration={duration}
+            onClose={handleClose}
+            onExited={handleExited}>
+            <Alert severity={type} onClose={handleClose}>
```
kachar, line 23: At the moment we tend to hide all alerts on click via the [`clickaway` listener](https://material-ui.com/api/snackbar/#props) of `<Snackbar />`

We might discard the clickaway hiding by checking the `reason: SnackbarCloseReason` of the handler.

Another question: Do you think each individual alert should clear all alerts when it exits it's transition via `onExited`?

How do you find such approach @StanislavSt ?

```tsx
import React from 'react'
import { observer } from 'mobx-react'
import { Snackbar, SnackbarCloseReason } from '@material-ui/core'
import { Alert } from '@material-ui/lab'

function SnackBar() {
  const { getAlerts } = AlertStore

  const handleSnackBarClose = (id: number) => (
    _event: React.SyntheticEvent,
    reason: SnackbarCloseReason,
  ) => {
    if (reason === 'clickaway') {
      return
    }
    AlertStore.hide(id)
  }
  const handleClose = (id: number) => () => AlertStore.hide(id)

  return (
    <>
      {getAlerts.map(({ id, show, duration, type, message }) => {
        return (
          <Snackbar
            key={id}
            open={show}
            autoHideDuration={duration}
            onClose={handleSnackBarClose(id)}>
            <Alert severity={type} onClose={handleClose(id)}>
              {message}
            </Alert>
          </Snackbar>
        )
      })}
    </>
  )
}

export default observer(SnackBar)
```

kachar (commented):
@StanislavSt Awesome work on this POC

From my perspective it's worth going into the direction of MobX for the state management. :rocket: 

There are couple of questions around the implementation of the alerts logics, but nothing too drastic. 

I'll be happy to start a discussion around the review

src/components/layout/Snackbar.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import { observer } from 'mobx-react'
+import { Snackbar } from '@material-ui/core'
+import { Alert } from '@material-ui/lab'
+import { AlertStore } from '../../stores/AlertStore'
```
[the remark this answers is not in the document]
StanislavSt: Cool, did not realize that.

src/components/layout/Snackbar.tsx
StanislavSt: @kachar  Thank you for the fast review! 

I agree that we should not close the alert on `clickaway`.
I agree that `onExited` is not needed, I used it for an animation problem - but now when I tested again it is not needed.

Using the `id` approach will cause problems, such as after the `autoHideDuration` has expired you will close the current  alert. For example if I show multiple alerts one after another, the last one might close much earlier before the 3000ms has expired, because a callback comes from a previous alert. 

Using `id = this.alerts.length` to hide an alert
Alert_using_id

Without using  id to hide an alert:
alert_without_id


We could use an unique id and hide based on that, something like this:
```
 hide(id: number) {
    const index = this.alerts.findIndex((alert) => alert.id === id)
    this.alerts[index].show = false
  }
```
But I think the result will be the same, since we always clear the Alerts array , before pushing a new one.

kachar: I think we're good to go with this one.
You're right about the multiple alerts storage handling in the case we show a single alert at all times.

If the expectations for the behavior evolve we'll update the code accordingly.

Thanks again for the energy well spent on this PR! :zap:

kachar (approved):

PR #31, opened 2021-02-08
Improve formik form handling

uzuntonev:
## Motivation and context

 - Fix: properly translate the validation error message. Example below.
 - Made custom hook to remove complexity from components which will use formik
 - Move validation schema declaration on a separate file

## Screenshots:

Befor|After
---|---
| kJ978ViAlz-output | Aw6oKQggYz-output |


kachar (approved):
Nice extraction of common logics! Custom formik hook is top :100: 

We might improve it further if we extract the form from the page component and make the form model closer to the form component. I'll try to do an iteration as poc.

Issue #33, opened 2021-02-11
Generic form field component


kachar:
An iteration over the `<TextField />` components so they interact with the formik context and get nested with ease.

Issue #7, opened 2020-12-18
Create common components


kachar:
## Motivation

In order to have fully working frontend project we'll need couple of common components that will ease the development.

This issue is used to track the progress on their individual development cycle.

https://ora.pm/project/277933/task/4495226

## Components

- [x] `AppBar` (#38, thanks to @kachar)
- [x] `Footer` (#12, thanks to @ani-kalpachka)
- [x] `Nav`
- [x] `MobileNav`
  > Responsive layout of the navigation menu
- [x] `Link` (@kachar)
  > Combine interfaces of https://nextjs.org/docs/api-reference/next/link and https://material-ui.com/components/links/
- [x] `LinkButton` (@kachar)
  > Combine interfaces of https://nextjs.org/docs/api-reference/next/link and https://material-ui.com/components/buttons/
- [x] `GenericForm` (#8, thanks to @Nikola-Andreev)
  > Component to be used as base in order to allow easy replacement of underlying form library
- [x] `EmailSubscriptionForm`
  > Reusable email subscription form. Might be used in the footer or in the body
- [x] `LoginForm` (#12, thanks to @ani-kalpachka)
- [x] `RegisterForm` (#12, thanks to @ani-kalpachka)
- [x] `ResetEmailForm` (#12, thanks to @ani-kalpachka)
- [x] `ConfirmationDialog` (#17, thanks to @Nikola-Andreev)
  > Title, Description, Actions (Yes/No, Confirm/Cancel) https://material-ui.com/components/dialogs/#dialog

:information_source: Feel free to pick a component and extract the implementation in separate issue or PR.

sotir-ls:
Have we got any designs (breakpoints) to work against?

kachar:
@sotir-ls Not yet as far as I know

Anyhow the idea is to create the most common usecases for a web app - the must-haves

Feel free to extend the list in the issue description

kachar:
Closing this one as all the points above are done :heavy_check_mark: 

Steady foundation for the app! Thanks to everyone involved in the process :100:

PR #42, opened 2021-02-20
[7, 33] Create generic form components

Nikola-Andreev:
* Create GenericForm component, wrapper for formik form
* Create FormTextField component, wrapper for TextField
* Apply GenericForm and FormTextField in all app forms

Resolves:
#33 #7(Generic form component) 

No visual changes, only functionality.


src/components/common/form/FormTextField.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { TextField } from '@material-ui/core'
+import { FormContextConsumer } from 'components/common/form/FormContext'
+import { translateError } from 'common/form/useForm'
+
+export type RegisterFormProps = {
+  type: string
+  label: string
+  name: string
+  translate: any
+}
```
kachar, line 11: @Nikola-Andreev What's the idea behind passing the `translate` function every time as required prop, instead of using the hook `useTranslation()` inside the component to get it ?

It's much better to strictly type the props, having in mind that we can reuse the `TextFieldProps` straight away from MUI

src/components/common/form/GenericForm.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { formik } = useForm({ initialValues, onSubmit, validationSchema })
+
+  return (
+    <form onSubmit={formik.handleSubmit}>
+      <FormContextProvider value={formik}>{children}</FormContextProvider>
+    </form>
+  )
+}
+
+export default GenericForm
```
kachar, line 23: We can improve the typing by passing the `FormikConfig` to the component. This will gives several advantages combined with `...formProps`

- We don't have to type them manually
- We can pass more props to the `useForm` hook without changing this component
- We pass specific type to `useForm` and in this way the form knows what kind of values we drill down

By using `React.PropsWithChildren<{}>` we skip the need to specify the `children` type manually

I would make the context the wrapper of the actual `<form />` component to keep the semantics.

```suggestion
import { FormikConfig } from 'formik'
import React, { PropsWithChildren } from 'react'

import useForm from 'common/form/useForm'
import { FormContextProvider } from 'components/common/form/FormContext'

export type GenericFormProps<T> = PropsWithChildren<FormikConfig<T>>

export default function GenericForm<T>({ children, ...formProps }: GenericFormProps<T>) {
  const { formik } = useForm(formProps)

  return (
    <form onSubmit={formik.handleSubmit}>
      <FormContextProvider value={formik}>{children}</FormContextProvider>
    </form>
  )
}
```

src/components/common/form/FormContext.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+const { Provider, Consumer } = React.createContext({})
+
+export type FormContext = {
+  value: any
```
kachar, line 5: Let's give those an expected type

kachar (changes requested):
@Nikola-Andreev I've added some notes, mainly on the types being used in the components.

This PR helped me reconsider the decision to disable the eslint rule about `any` types in the app, so I've added a PR to turn them back on, at a warning level with https://github.com/daritelska-platforma/frontend/pull/43

Please let me know what's your vision on the topic and we can discuss further

src/components/common/form/FormTextField.tsx
Nikola-Andreev: Performance mainly. Why to call `useTranslation()` in each field, when it's already called outside on each form.
But if you think it is better i will add it inside `FormTextField`

kachar (approved):
Nice one :rocket:

Issue #49, opened 2021-02-22
Replace dropdown in support form with a list of checkboxes


kachar:
- easy to maintain and validate
- less complication in terms of code
- similar UX to other selection options

kachar:
Partially connected to #60

kachar:
Closed with #60

PR #61, opened 2021-02-23
Changed dropdown volunteer roles to checkboxes and fixed validation

vikinatora:
Regarding issue #49 

Additional questions steps:
1. Changed dropdown roles to checkboxes
2. Fixed error messages appearing before submitting the text

Info step: 
1. Fixed validation on GDPR checkbox appearing before submitting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,4 +35,3 @@ coverage
 
 # Env
 .env
-package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ coverage
 
 # Env
 .env
+package-lock.json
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -35,3 +35,4 @@ coverage
 
 # Env
 .env
+package-lock.json
```
uzuntonev, line 38: I don't think that is a good idea to have two different lockfiles.

uzuntonev (approved):
Great job! 💯 Perfect error handling. I think that improves user experience. According to me is better to have only one lockfile

.gitignore
kachar: I understood this ignore as a way to disregard the npm lockfile as the lockfile should always be stored in git as we do for `yarn.lock`

vikinatora: You are right, removed the line from .gitignore.

PR #70, opened 2021-02-25
Github actions update / prod/staging separation + automatic deployment

miraclebg:
## Motivation and context

It's about time we release the website to the wide public.

This update provides automatic production building and deployment to the prod env/host.

Just push a tag and 💥 that should be it!

Also - .env is now commited to the repo with defaults for all. It should be kept here.
If needed to override some of the parameters one should create a separate .env.local file and override the changes within it.


.env
```diff
@@ -0,0 +1,31 @@
...
+
+## API ##
+#########
+
+API_URL=https://api.podkrepi.bg/ # https://api.dp.localhost/
+
+## APP ##
+#########
+
+APP_URL=http://app.podkrepi.bg # http://localhost:3040
```
kachar, line 16: If this is the `.env.prod` file we need to add `httpS` to the url

miraclebg:
Several updates:

- dotenv npm package added do load the variables from the .env default file. Library initialized in next.config.js - probably not the right place but if it is not let me know so it can be corrected (and specify where to put it @).
- The required secrets to run the app are now passed along the docker-compose container init so they work properly (checked and they exist in the live container after deployment):

```
env:
          JWT_SECRET: ${{ secrets.APP_JWT_SECRET }}
          DISCORD_CLIENT_SECRET: ${{ secrets.APP_DISCORD_CLIENT_SECRET }}
```

So basically the idea here is to load all defaults from .env and to override the secret ones with the local env variables.

package.json
```diff
@@ -36,6 +36,7 @@
     "@types/yup": "^0.29.11",
     "@typescript-eslint/eslint-plugin": "^4.13.0",
     "@typescript-eslint/parser": "^4.13.0",
+    "dotenv": "^8.2.0",
```
kachar, line 39: We have to check if `dotenv` needs to be in dev deps or deps

kachar:
@miraclebg Pushed some commits to ensure we pass the discord client id as well as the secret and to place the secret envs in the server side config.

The location of `dotenv` in `next.config.js` seems to be a proper choice.

docker-compose.prod.yml
```diff
@@ -0,0 +1,22 @@
+version: '3'
+services:
+  app:
+    image: ghcr.io/daritelska-platforma/frontend/web:deployment-changes
```
kachar, line 4: @miraclebg We only need to find a way to define the newly built version

```suggestion
    image: ghcr.io/daritelska-platforma/frontend/web:${DEPLOY_TAG}
```

And pass `DEPLOY_TAG` from the `TapTap21/docker-remote-deployment-action` 

Without such mechanism we the tag `deployment-changes` will act the same as tag `latest`

Any other ideas are welcome

PR #77, opened 2021-02-26
add heading image, change animated texts, style icons

ani-kalpachka:
- Added heading image.
- Changed animated text values in the heading.
- Increased typewriter's typing and deleting speed.
- Changed sizes and alignment of the icons.

Video:
https://user-images.githubusercontent.com/14351733/109385197-c78d1080-78fa-11eb-9a1a-9b2bfda8f9b2.mp4


src/components/layout/Layout.tsx
```diff
@@ -32,7 +32,7 @@ const useStyles = makeStyles((theme) =>
     },
     offset: {
       ...theme.mixins.toolbar,
-      marginBottom: theme.spacing(6),
+      marginBottom: theme.spacing(0),
```
kachar, line 35: As we reduce the margin here we might want to introduce a margin to all other pages as the title is stitched to the app bar.



image

kachar: We can place a bool option to `<Layout disableOffset />`

```tsx
{!disableOffset && <div className={classes.offset} />}
```

And we get rid of the margin specific for the page

kachar (approved):
:clap: Well done!

Issue #37, opened 2021-02-16
SSR doesn't work for i18n


kachar:
The server renders plain non-translated objects on first request. Then the client renders the i18n which causes a flicker

```html
<title>nav.login</title>
```

kachar:
Waiting for the i18n + next.js 10 to be supported at https://github.com/isaachinman/next-i18next/issues/869

kachar:
There's a new beta of `next-18next` we we might bump this task up

https://github.com/isaachinman/next-i18next/blob/2640dc4fc1234b286e7b690bcb338c23329c0438/README.md

StanislavSt:
@kachar Looks like next-i18next v8.0.0 is out  - https://github.com/isaachinman/next-i18next/releases/tag/v8.0.0.
Time to fix this bug ?

PR #78, opened 2021-02-26
#37 SSR doesn't work for i18n

StanislavSt:
Remove the old custom implementation of i18n and use [next-i18next](https://github.com/isaachinman/next-i18next) instead.

Closes https://github.com/daritelska-platforma/frontend/issues/37
## Screenshots:

| srr-before | srr-now

**New dependencies**:
- `next-i18next` : v8.0.2


kachar:
The build problem seems to be reported at https://github.com/isaachinman/next-i18next/issues/995

StanislavSt:
Will convert this to a draft for now until there are more news.

kachar:
I guess we can avoid the issue if we import the types for @types/hoist-non-react-statics package temporary 

Do you think it's worth it ?

StanislavSt:
If we do that, another set if issues will follow :(
image

kachar (approved):
@StanislavSt Awesome integration! Smooth as jazz :saxophone:

PR #80, opened 2021-02-27
[75] Replace nivo pie with highcharts pie

Nikola-Andreev:
## Motivation and context

+ Animation added
+ Responsive on mobile(phone view)
+ Run animation only when component is visible
- Hides most of the labels in phone portrait mode, but user can see the info clicking on the section

Feel free to change the colors (generated them randomly)

## Screenshots:
Screenshot (684)
Screenshot (685)
Screenshot (687)
captured


kachar (approved):
@Nikola-Andreev Looks awesome! Nice and simplistic integration of the lib

Allowed myself to update the colors and change the sorting of the data so the biggest numbers are on the left side.

Before|After
---|---
image|image

PR #96, opened 2021-03-04
Homepage styling fixes

ani-kalpachka:
- Changed header image height - 800px below 1600px display and 950px above this display breakpoint.
- Fixed spacings as discussed with designer.
- Removed vertical scroll on mobile devices.
- Added discord team image in Team section.

Screenshots:
Annotation 2021-03-04 195212

Annotation 2021-03-04 200225

Annotation 2021-03-04 202816


src/components/index/sections/Jumbotron.tsx
```diff
@@ -10,23 +10,27 @@ import Typewriter from '../helpers/Typewriter'
...
       backgroundSize: 'cover',
       color: theme.palette.common.white,
-      height: theme.spacing(81),
       [theme.breakpoints.up('md')]: {
-        paddingTop: theme.spacing(35),
         height: theme.spacing(100),
+        paddingTop: theme.spacing(31),
+      },
+      [theme.breakpoints.up(1600)]: {
+        height: theme.spacing(119),
```
kachar, line 26: When we're dealing with sizes larger than `spacing(20)` i think we can go directly to the expected size in `rem` or `px` 

The cognitive understanding of how much `spacing(119)` is might be too hard as we need to multiply 119 * 8 to get the figure 952 pixels. We might shoot directly for `950px`

The spacing fn is more useful when we're dealing with gaps, paddings and margins or square objects.

src/components/index/sections/Jumbotron.tsx
```diff
@@ -36,8 +40,9 @@ const useStyles = makeStyles((theme) =>
     podkrepiButton: {
       color: theme.palette.common.white,
       borderColor: theme.palette.common.white,
-      padding: theme.spacing(2, 4),
+      padding: theme.spacing(1.5, 5),
       fontWeight: 600,
+      fontSize: theme.spacing(2),
```
kachar, line 45: For font sizes we can use `fontSize: theme.typography.pxToRem(15)` which gives us similar result as spacing with more responsive

src/components/index/sections/TeamSection.tsx
```diff
@@ -27,14 +32,17 @@ export default function TeamSection() {
...
         {t('index:team-section.heading')}
       </Typography>
-      <Grid item>
-        <Typography variant="body2">{t('index:team-section.content')}</Typography>
+      <Grid container>
+        <Grid item>
+          <Typography variant="body2">{t('index:team-section.content')}</Typography>
+        </Grid>
+        <Grid item>
+          <img src={`img/discord-team-image.png`} className={classes.teamImage} />
```
kachar, line 43: It's better to load the image with absolute route such as `/img/discord-team-image.png` as this url will break if we use the same component deeper in the route structure. Now it works because we use it on `/` but that's special case.

src/components/index/sections/TeamSection.tsx
```diff
@@ -27,14 +32,17 @@ export default function TeamSection() {
       justify="center"
       component="section"
       className={classes.container}>
-      <Typography variant="h5" component="h2" className={classes.heading}>
```
kachar, line 30: Please undo this change as it was requested by the SEO team. We should only have H1 and H2 on the homepage. If we don't specify it on the `<Typography variant="h5">` we'll get `<h5>` instead.

We can also adjust that globally, but I'm still not convinced we should 
https://material-ui.com/components/typography/#changing-the-semantic-element

kachar (changes requested):
@ani-kalpachka Nice iteration, well done

Bonus points for clearing the horizontal scroll on mobile devices :top: :100: 

There are couple of discussion points, let's tackle them prior the merge

kachar (approved):
:clap:

PR #121, opened 2021-03-15
Removed horizontal scrolling on homepage on mobile devices

ani-kalpachka:
Removed horizontal scrolling on homepage on mobile devices. 
The issue was caused by the larger font-size of the page title. Now it is 45px for mobile devices (under 600px):

image


kachar (approved):
Nice catch

PR #154, opened 2021-04-06
Added infographic section image with <Image /> tag

dimitur2204:
## Motivation and context

To show the path infographic on the Activity Section

https://github.com/daritelska-platforma/frontend/issues/124, only addresses part of the issue

https://user-images.githubusercontent.com/75036745/111613622-9fe0e800-87d6-11eb-950e-c453f40d64ba.jpg

Added the activity info graphic to the homepage so that users can visually see the process of creating and financing a campaign

I added the graphics as static images in the public folder and rendered them based on the viewport and language

## Screenshots:
Before|After
---|---
There was nothing belowe the text on the activity section | Added the activity infographic
image|image

## Testing

Tested on all resolutions widths from 1920 to 320 to check if it doesn't break

The main homepage below the banner

Local dev environment

Added the two images to the Activity Section file
image


kachar:
@dimitur2204 Looks awesome!

Maybe the first font from the infographic is different than the rest? If that's in our control it's gonna be awesome to update it so the same font is used everywhere

image

dimitur2204:
> @dimitur2204 Looks awesome!
> 
> Maybe the first font from the infographic is different than the rest? If that's in our control it's gonna be awesome to update it so the same font is used everywhere
> 
> image
> 
> 

Yes I actually noticed that but forgot to mention it. I will see what the problem is tomorrow.

kachar:
@swolf86 Can you please advice on the font used in the infographics

As it seems there are some text elements that are not converted to curves so the font is actually required.

If possible can you reupload the svg with all font elements as paths?

kachar (approved):
Iteration looks smooth :100: 

Let's merge this one and extract a dedicated issue for the font problem in the svg

https://github.com/daritelska-platforma/frontend/issues/157

PR #161, opened 2021-04-08
157 Changed svgs to the ones with the correct fonts

dimitur2204:
## Motivation and context

Previously the one of the sections in the infographic had a wrong font

https://github.com/daritelska-platforma/frontend/issues/157

Added the new fixed svgs in place of the old ones

## Screenshots:

|image | image
 |

## Testing

Tested on all the resolutions and on both locales for any bugs

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:


kachar (approved):
Awesome! Fixes the issue

Issue #159, opened 2021-04-08
[Bug] Support form is not submitted when newsletter question is confirmed


kachar:
**Describe the bug**
Support form is not submitted when newsletter modal question is confirmed. The final step is shown without the actual submission to the backend

**To Reproduce**

https://user-images.githubusercontent.com/893608/114022480-7d446b00-987a-11eb-87b9-7df3369d06e8.mp4

**Expected behavior**

https://user-images.githubusercontent.com/893608/114022574-95b48580-987a-11eb-83fa-ea624ea86e43.mp4

PR #162, opened 2021-04-08
Submitting the support form through the newsletter dialog

dimitur2204:
## Motivation and context

To be able to submit the form immediately after you either confirm or cancel the popup dialog.

#159

Previously when clicking the dialog buttons only the steps were advancing and no request was sent

Added a reference to the form with useRef() and used it to submit the form through the handleNewsletterDialogCancel/Confirm methods. There is a newer way to access the form through useFormikContext() hook, but that only works if the component you are accessing the hook from is a child of <Formik /> which the <SupportForm /> is not so I did it the old school way with innerRef.

## Screenshots:
| No request was sent | A request is now sent to the backend |
|https://user-images.githubusercontent.com/893608/114022480-7d446b00-987a-11eb-87b9-7df3369d06e8.mp4 | image
 |

## Testing

Tested it both without checking the box and clicking confirm and cancel. And with a checked box previously and all request were sent correctly. 
Checked Box - newsletter:true 
Confirm - newsletter:true 
Cancel - newsletter:false 

/support (last step)

Local Dev Environment

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/components/support-form/SupportForm.tsx
+++ b/src/components/support-form/SupportForm.tsx
@@ -148,7 +148,7 @@ const NewsletterDialog = ({ isOpen, handleConfirm, handleCancel }: NewsletterDia
 export default function SupportForm() {
   const { t } = useTranslation()
   const classes = useStyles()
-  const formRef = useRef<FormikProps<SupportFormData>>() as RefObject<FormikProps<SupportFormData>>
+  const formRef = useRef<FormikProps<SupportFormData>>(null)
   const form = formRef?.current
   const [loading, setLoading] = useState(false)
   const [maxStep, setMaxStep] = useState<Steps>(Steps.ROLES)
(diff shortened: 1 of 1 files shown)


src/components/support-form/SupportForm.tsx
```diff
@@ -148,6 +148,8 @@ const NewsletterDialog = ({ isOpen, handleConfirm, handleCancel }: NewsletterDia
 export default function SupportForm() {
   const { t } = useTranslation()
   const classes = useStyles()
+  const formRef = useRef<FormikProps<SupportFormData>>() as RefObject<FormikProps<SupportFormData>>
```
kachar, line 151: You can avoid setting the second `as` part if you provide a default value of `null` for the ref

```suggestion
  const formRef = useRef<FormikProps<SupportFormData>>(null)
```

This changes the way TS infers the type of the ref from 

```ts
function useRef<T = undefined>(): MutableRefObject<T | undefined>;
```
to
```ts
function useRef<T>(initialValue: T|null): RefObject<T>;
```

Issue #160, opened 2021-04-08
[Bug] Homepage Banner About us button text wrapping


dimitur2204:
**Describe the bug**
The text on the hero banner "About the project version on the EN version wraps on two lines"

**To Reproduce**
Steps to reproduce the behavior:\
1. Go to '/en'
2. See the about button

**Expected behavior**
The text in the button does not wrap and it looks as the support us button

**Screenshots**
![image](https://user-images.githubusercontent.com/61479393/114070997-49cd0500-98a9-11eb-8c73-6c23ced0e7df.png)

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: Chrome 
- Version Version 89.0.4389.114

**Additional context**

kachar:
Will be fixed with #164

kachar:
On master

PR #164, opened 2021-04-09
Fix Support section and Jumbotron buttons bug

ani-kalpachka:
- Fixed Support us section.
Desktop: 
image
Mobile:
image

- Fixed 'About Project' button bug with EN translation:
image

### DIFF
--- a/src/components/index/sections/Jumbotron.tsx
+++ b/src/components/index/sections/Jumbotron.tsx
@@ -47,7 +47,7 @@ const useStyles = makeStyles((theme) =>
       borderRadius: theme.spacing(3),
       fontWeight: 500,
       fontSize: theme.typography.pxToRem(15),
-      width: theme.spacing(27),
+      minWidth: theme.spacing(27),
       height: theme.spacing(7),
       margin: theme.spacing(2),
       [theme.breakpoints.up('sm')]: {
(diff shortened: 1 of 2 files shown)


src/components/index/sections/Jumbotron.tsx
```diff
@@ -45,25 +45,33 @@ const useStyles = makeStyles((theme) =>
...
     },
     podkrepiButton: {
       color: theme.palette.common.white,
       backgroundColor: theme.palette.primary.main,
       border: `2px solid ${theme.palette.primary.main}`,
       borderRadius: theme.spacing(3),
-      padding: theme.spacing(1.5, 5),
       fontWeight: 500,
       fontSize: theme.typography.pxToRem(15),
+      width: theme.spacing(27),
```
kachar, line 64: Adding fixed width is not the best way to handle the buttons as they will wrap if the text is changed.
Maybe we can replace it with `minWidth` of that's the visual expectation.

src/components/index/sections/SupportUsSection.tsx
```diff
@@ -10,42 +10,45 @@ const useStyles = makeStyles((theme) =>
...
     },
     clickHereButton: {
       color: theme.palette.common.white,
       backgroundColor: theme.palette.primary.main,
       border: `2px solid ${theme.palette.primary.main}`,
       borderRadius: theme.spacing(3),
-      padding: theme.spacing(1, 5),
-      margin: theme.spacing(4, 0),
-      fontWeight: 500,
+      width: theme.spacing(25),
```
kachar, line 42: Fixed height and width will be troublesome at some point, we better restrict those with `minWidth` and `minHeight` if really necessary.

kachar (changes requested):
Please adjust the fixed values `height` and `width` so we don't end up in the same situation again as #160

kachar (approved):

PR #188, opened 2021-05-09
[Feature] module campaigns project reorganization

xenry:
## Motivation and context

Currently campaigns is more or less a dummy module that provides a few endpoints which return hardcoded data. 

##  What is included?

- Module campaigns was reworked to follow "clean architecture" principles. More on that can be found [here](https://www.ssw.com.au/rules/rules-to-better-clean-architecture)
- Reworked service's request/response contracts (campaign.proto)
- Added new endpoints - _GetCampaignDetails_, _CreateCampaignType_, _ListCampaignTypes_
- Adjusted the script for initial sql tables creation - tables are now in the _campaign_ scheme, names are snake_case, adjusted types and mandatory fields
- Wired up some of the endpoints - _CreateCampaignType_, _CreateCampaign_, _GetCampaignDetails_ are now implemented end-to-end and are reading/writing to db
- Added integration tests

## Testing

- Integration tests were written. In order to execute them, one will need to have local running cockroach db (or postgres) with initial migration applied (e.g. tables created)


db/migrations/campaign/V4.0.0.1__table_campaigns.sql
```diff
@@ -1,48 +1,51 @@
 BEGIN TRANSACTION;
 
-CREATE TABLE IF NOT EXISTS "CampaignTypes" (
-    "Id" uuid NOT NULL DEFAULT (gen_random_uuid()),
-    "Name" character varying(50) NOT NULL,
-    "DescriptionKey" character varying(50) NULL,
-    CONSTRAINT "PK_CampaignTypes" PRIMARY KEY ("Id")
+CREATE TABLE IF NOT EXISTS campaign.campaign_types (
```
kachar, line 3: @xenry I'm not sure if editing the same migration will have the expected/desired effect when delivered to production.

We might need to drop the tables and create them in a separate migration like `V4.0.0.2__table_campaigns.sql

kachar (approved):
:clap:

mayapeneva (approved):

db/migrations/campaign/V4.0.0.1__table_campaigns.sql
xenry: yep, good idea, will fix it later today

xenry: @kachar fixed :)

PR #210, opened 2021-06-02
Add English translation and keys to "what is done" section on about page.

arthas168:
## Screenshots:

| image | image |

## Testing

Dev testing only.

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:


arthas168:
The one thing I'm not sure of is the holder name (с титуляр...). Should it stay as is and say "with holder Сдружение Подкрепи БГ" or should the name also be translated in English e.g.  "with holder Association Podkrepi BG" ?

frontend/src/components/about-project/sections/WhatIsDone.tsx
```diff
@@ -65,69 +65,64 @@ const CheckedLine = ({ label, href }: CheckedLineProps) => {
...
+      <br />
+      <strong>IBAN: {ibanNumber}</strong>
+      <br /> {t('about-project:with-holder')} <strong>{t('about-project:association-name')}</strong>
+      .
+    </div>
+  )
+
+  const leftColumnLabels: CheckedLineProps[] = [
+    {
+      label: t('about-project:work-organisation'),
```
kachar, line 85: You don't have to translate the items here, as they are getting translated inside `CheckedLine` once again. You might just pass the keys inside the `leftColumnLabels` and `rightColumnLabels`

frontend/src/components/about-project/sections/WhatIsDone.tsx
```diff
@@ -65,69 +65,64 @@ const CheckedLine = ({ label, href }: CheckedLineProps) => {
...
-    label: 'Приготвили сме High-level архитектурнo/организационна карта на платформата',
-    href: 'https://docs.podkrepi.bg/general/arkhitektura/architecture',
-  },
-]
-
 export default function WhatIsDone() {
   const { t } = useTranslation()
   const classes = useStyles()
 
+  const BankAccount = () => (
```
kachar, line 72: In order to translate this component we might use `const { t } = useTranslation()` inside of it, instead of moving it inside the body of `WhatIsDone`. This will ease the separation between them

arthas168: thanks! I'll separate them

frontend/src/components/about-project/sections/WhatIsDone.tsx
arthas168: Seemed like magic how the `CheckedLine` handles it, but that's what I get for not looking at it beforehand :D

kachar (approved):
Awesome contribution! :top: :100: 

Thanks for improving the codebase

Issue #208, opened 2021-06-02
[Bug] Support Request endpoint is not storing the JSON data in the DB


kachar:
The support request marshaler is not doing it's job properly and all data becomes `null` when stored in the db.

https://github.com/daritelska-platforma/frontend/blob/340a15b17e4210faded7f502735ab76ed818ca7b/backend/contact/database/datatypes.go#L12-L24

```shell
curl --request POST \
  --url http://localhost:5000/api/v1/support-request \
  --header 'Content-Type: application/json' \
  --data '{
	"person": {
		"email": "test@example.com",
		"name": "John Doe",
		"phone": "+35983333333333",
		"comment": "test",
		"terms": true,
		"gdpr": true,
		"newsletter": true
	},
	"support_data": {
		"roles": {
			"benefactor": false,
			"partner": false,
			"associationMember": true,
			"company": false,
			"volunteer": false
		},
		"benefactor": {
			"campaignBenefactor": false,
			"platformBenefactor": false
		},
		"partner": {
			"npo": false,
			"bussiness": false,
			"other": false,
			"otherText": ""
		},
		"volunteer": {
			"backend": false,
			"frontend": false,
			"marketing": false,
			"designer": false,
			"projectManager": false,
			"devOps": false,
			"security": false,
			"financesAndAccounts": false,
			"lawyer": false,
			"qa": false
		},
		"associationMember": {
			"isMember": true
		},
		"company": {
			"sponsor": false,
			"volunteer": false,
			"other": false,
			"otherText": ""
		}
	}
}'
```

PR #213, opened 2021-06-02
[#208] fix storing JSON into DB

cupakob:
## Motivation and context

bugfix issue #208 

## Testing

unit tests added

### DIFF
--- a/backend/contact/database/datatypes_test.go
+++ b/backend/contact/database/datatypes_test.go
@@ -1,0 +1,88 @@
+package database_test
+
+import (
+	"fmt"
+	"github.com/daritelska-platforma/v2/database"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+const payload = `{
+	"person": {
+		"email": "test@example.com",
+		"name": "John Doe",
+		"phone": "+35983333333333",
+		"comment": "test",
+		"terms": true,
+		"gdpr": true,
+		"newsletter": true
+	},
+	"support_data": {
+		"roles": {
+			"benefactor": false,
+			"partner": false,
+			"associationMember": true,
+			"company": false,
+			"volunteer": false
+		},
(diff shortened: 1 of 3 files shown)


backend/contact/database/datatypes_test.go
```diff
@@ -0,0 +1,87 @@
...
+    // given
+    jsonb := &database.JSONB{}
+    _ = jsonb.Scan(payload)
+
+    // when
+    value, err := jsonb.Value()
+
+    // then
+    assert.NoError(t, err)
+    assert.NotNil(t, value)
```
kachar, line 85: @cupakob Maybe we can add a test to make sure the output is the same as the input payload

```suggestion
    assert.NotNil(t, value)
    assert.JSONEqf(t, payload, fmt.Sprintf("%v", value), "payload-differs")
```

kachar:
@cupakob I've tested this PR locally, it does the convertion to string properly, tests coonfim that, but the actual pain of not storing the data in the table still persists.

Even with this fix it seems that we're missing something else to complete the issue

Do you have any idea what else might cause it ?

backend/contact/database/datatypes_test.go
cupakob: then `assert.NotNil` is not needed anymore. I've never tested, if the entry is correct persisted. I just read the comment, that the marshaler is not doing it's job properly. And that was my approach to fix the problem.

cupakob:
I've never tested, if the entry is correct persisted. I just read the comment, that the marshaler is not doing it's job properly. And that was my approach to fix the problem.

kachar:
@cupakob Do you think it will be more appropriate to merge & close this one and open a new issue?

cupakob:
yes, i think, it will be better to create another one, becaus the problem from the description is solved.

PR #216, opened 2021-06-03
[FIX] Refactor Privacy and Policy and Terms of Service

ani-kalpachka:
- Fixed the console errors on PrivacyPolicyPage.tsx.
- Reused common styles for PrivacyPolicyPage.tsx and TermsOfServicePage.tsx.
- Changed stylelint max-nesting-depth rule.
- Refactor some styles.


ani-kalpachka:
ESLint does not allow more than one selector nesting (global.scss).

kachar (approved):
Good call for extending the config to suit better our goals

Swift transition to global rules

PR #235, opened 2021-06-06
[Feature][231] Migrate husky v4 to v6

ani-kalpachka:
* https://github.com/typicode/husky-4-to-6
* https://typicode.github.io/husky/#/?id=custom-directory


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm test
```
kachar, line 4: We might need to update this so it's `yarn lint-staged`

frontend/package.json
```diff
@@ -47,7 +48,7 @@
     "eslint-config-prettier": "8.3.0",
     "eslint-plugin-prettier": "3.4.0",
     "eslint-plugin-react": "7.24.0",
-    "husky": "4.3.8",
+    "husky": "^6.0.0",
     "lint-staged": "11.0.0",
     "prettier": "2.3.0",
     "stylelint": "13.13.1",
```
kachar, line 54: We don't need `husky` config in `package.json` anymore

kachar: Also please use fixed version `"6.0.0"` instead of `"^6.0.0"

kachar (approved):
:100:

Issue #209, opened 2021-06-02
[Bug] GTM only tracks the initial load of the page


kachar:
We send the view events only on initial load of the SPA website

We should send `page_view` event on each change of internal routing

```tsx
  useEffect(() => {
    const onRouteChange = (url: string) => {
      trackEvent({
        event: 'page_view',
        language: i18n.language,
        page_title: document.title,
        page_pathname: url,
        page_location:
          document.location.protocol +
          '//' +
          document.location.hostname +
          document.location.pathname +
          document.location.search,
      })
    }

    router.events.on('routeChangeComplete', onRouteChange)
    return () => router.events.off('routeChangeComplete', onRouteChange)
  }, [i18n.language])
```

Additional read:

- https://dev.to/ornio/add-google-analytics-through-gtm-google-tag-manager-on-next-js-5e4f

PR #239, opened 2021-06-06
[#209] Bugfix: GTM only tracks initial page load

RRoussev88:
bugfix issue #209

### DIFF
--- a/frontend/src/common/util/useGTM.ts
+++ b/frontend/src/common/util/useGTM.ts
@@ -3,7 +3,7 @@ import { default as TagManager, TagManagerArgs } from 'react-gtm-module'
 
 const gtmId = getConfig().publicRuntimeConfig.GTM_ID
 
-export type EventType = 'event_1' | 'change_language'
+export type EventType = 'event_1' | 'change_language' | 'page_view'
 
 export type DataEvent = { event: EventType; [key: string]: unknown }
 
--- a/frontend/src/common/util/useGTM.ts
+++ b/frontend/src/common/util/useGTM.ts
@@ -3,7 +3,7 @@ import { default as TagManager, TagManagerArgs } from 'react-gtm-module'
 
 const gtmId = getConfig().publicRuntimeConfig.GTM_ID
 
-export type EventType = 'event_1' | 'change_language' | 'page_view'
+export type EventType = 'change_language' | 'page_view'
 
 export type DataEvent = { event: EventType; [key: string]: unknown }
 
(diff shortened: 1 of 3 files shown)


frontend/src/common/util/useGTM.ts
```diff
@@ -3,7 +3,7 @@ import { default as TagManager, TagManagerArgs } from 'react-gtm-module'
 
 const gtmId = getConfig().publicRuntimeConfig.GTM_ID
 
-export type EventType = 'event_1' | 'change_language'
+export type EventType = 'event_1' | 'change_language' | 'page_view'
```
kachar, line 6: I think we can delete `event_1` as it's not currently used

frontend/src/pages/_app.tsx
```diff
@@ -13,20 +14,42 @@ import 'styles/global.scss'
...
     // Remove the server-side injected CSS.
     const jssStyles = document.querySelector('#jss-server-side')
     if (jssStyles && jssStyles.parentElement) {
       jssStyles.parentElement.removeChild(jssStyles)
     }
+  }, [])
 
-    // Init GTM
-    useGTM().initialize({
+  useEffect(() => {
```
kachar, line 29: Good call to use dedicated effect for the gtm :rocket: 

As I see we're going to `initialize` gtm on every language change now.

I can't evaulate if this is okay or not, do you have a view on that?

If it's not okay, we can use a 3rd effect just for the init method with no dependencies

frontend/src/pages/_document.tsx
```diff
@@ -1,15 +1,29 @@
...
 import FaviconMetadata from 'components/brand/FaviconMetadata'
 
+const GTM_ID = getConfig().publicRuntimeConfig.GTM_ID
+
 export default class CustomDocument extends Document {
   render() {
     return (
       <Html lang="en">
         <Head>
+          {/* Google Tag Manager */}
```
kachar, line 16: I don't get why we have this placed in the document as the module we're using `react-gtm` is doing exactly this job

https://github.com/alinemorelli/react-gtm/blob/b1257405a82e900c174032a773a8ba70a8628081/src/Snippets.js#L12-L22

When we call `initialize` this is what is injected in our `<Head>` dynamically, so we don't have to do it manually.

RRoussev88: Looks like you are right. I should have read the `react-gtm-module` documentation :)

frontend/src/pages/_app.tsx
RRoussev88: I decided to put GTM initialization back to the first useEffect after all. It is better to call it only once.

kachar (approved):
Awesome! :rocket: :100: 

Thanks for the contribution @RRoussev88!

PR #244, opened 2021-06-07
Feature/#193 create beneficiary

mayapeneva:
Added CreateBeneficiary endpoint.


kachar:
@mayapeneva Is this one ready for review / merge ?

xenry:
looking very good. Lets just add an organizerId to the beneficiary so that we know who created it.

backend/protos/campaign.proto
```diff
@@ -80,6 +83,22 @@ message CampaignTypesResponse {
 	repeated CampaignTypeDetails campaign_types = 1;
 }
 
+message CreateBeneficiaryRequest {
+
+	string name = 1;
+	common.nomenclatures.BeneficiaryType type = 2;
+	common.nomenclatures.ISO2CountryCode country_iso_code = 3;
+  string city = 4;
```
xenry, line 91: formatting here can be fixed, seems that some of the fields are indented with tabs

db/migrations/campaign/V4.0.0.4__table_campaigns.sql
```diff
@@ -0,0 +1,21 @@
...
+
+CREATE TABLE IF NOT EXISTS campaign.beneficiaries (
+    id uuid NOT NULL DEFAULT (gen_random_uuid()),
+    name character varying(100) NOT NULL,
+    type integer NOT NULL,
+    iso2country_code integer NOT NULL,
+    city character varying(50) NOT NULL,
+    email character varying(100) NULL,
+    phone character varying(50) NULL,
+    website character varying(500) NULL,
```
xenry, line 11: I am wondering, maybe better store these additional properties like email, phone, website in a jsonb column? What do you think @kachar @igoychev ?

db/migrations/campaign/V4.0.0.4__table_campaigns.sql
```diff
@@ -0,0 +1,21 @@
+BEGIN TRANSACTION;
+
+CREATE TABLE IF NOT EXISTS campaign.beneficiaries (
+    id uuid NOT NULL DEFAULT (gen_random_uuid()),
+    name character varying(100) NOT NULL,
```
kachar, line 5: We can split the `name` in `first_name`, `last_name` to ensure that we can use them separatelly.

db/migrations/campaign/V4.0.0.4__table_campaigns.sql
kachar: I think `email` is quite important to be present as indexable column with strict type check.

It's going to be nice to have additional props as jsonb column as you say to collect metadata about the user that requires no special db structure as we can't properly type the elements of a json field.

An example might be the `website` field which may have a type check for `http://` or `https://` which will guarantee db integrity.

db/migrations/campaign/V4.0.0.4__table_campaigns.sql
```diff
@@ -0,0 +1,21 @@
...
+    type integer NOT NULL,
+    iso2country_code integer NOT NULL,
+    city character varying(50) NOT NULL,
```
kachar, line 8: Using string `city` might lead to data duplication for free form entries like `Sofia`, `София`, etc

db/migrations/campaign/V4.0.0.4__table_campaigns.sql
```diff
@@ -0,0 +1,21 @@
...
+    email character varying(100) NULL,
+    phone character varying(50) NULL,
+    website character varying(500) NULL,
+	
+    CONSTRAINT "PK_Beneficiaries" PRIMARY KEY (id)
+);
+
+ALTER TABLE campaign.campaigns
+ADD CONSTRAINT "FK_Campaigns_Beneficiaries_BeneficiaryId" FOREIGN KEY (beneficiary_id) REFERENCES campaign.beneficiaries (id) ON DELETE CASCADE;
```
kachar, line 17: We can add `unique` contstraint to the `email` column so we don't have duplicate beneficiary records for the same person

backend/protos/campaign.proto
mayapeneva: Done.

db/migrations/campaign/V4.0.0.4__table_campaigns.sql
```diff
@@ -0,0 +1,21 @@
+BEGIN TRANSACTION;
+
+CREATE TABLE IF NOT EXISTS campaign.beneficiaries (
+    id uuid NOT NULL DEFAULT (gen_random_uuid()),
+    name character varying(100) NOT NULL,
+    type integer NOT NULL,
+    iso2country_code integer NOT NULL,
```
tanyogeorgiev, line 7: We can store country code as is (string representation). This number is not very self-explanatory.

db/migrations/campaign/V4.0.0.4__table_campaigns.sql
mayapeneva: Done.

db/migrations/campaign/V4.0.0.4__table_campaigns.sql
kachar: Of course these ideas above are not mandatory for this iteration, we can implement them in the next sprints as well.

backend/module-campaigns/src/Podkrepibg.Campaigns.Application/Beneficiaries/Commands/CreateBeneficiary/CreateBeneficiaryValidator.cs
```diff
@@ -0,0 +1,31 @@
...
+    using FluentValidation;
+
+    public class CreateBeneficiaryValidator : AbstractValidator<CreateBeneficiaryCommand>
+    {
+        public CreateBeneficiaryValidator()
+        {
+            RuleFor(b => b.Request.FirstName)
+                .NotEmpty();
+
+            RuleFor(b => b.Request.LastName)
```
DeyanAtanasov, line 13: Should we not use MaximumLength to validate whether the last name is longer than the database column? If we decide to do it/use it we should probably do that for all of the string values.

PR #245, opened 2021-06-08
Windows compatibility.

tanyogeorgiev:
- [x] **New environment variables for development**

- `COMPOSE_PATH_SEPARATOR` :   Customize path separator. COMPOSE_FILE supports multiple Compose files separated by a path separator ( on Linux and macOS the path separator is :, on Windows it is ; ). If set to ":" Windows based devs can compose without additional configs.

 
- [x] **Add Symlink  command for Windows**

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,7 @@
 ## General ##
 #############
 
+COMPOSE_PATH_SEPARATOR=":"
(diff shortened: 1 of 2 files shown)


kachar (approved):
@tanyogeorgiev Awesome contribution! This will solve many cross platform issues for many volunteers :100:

Issue #229, opened 2021-06-04
[Bug] Skip (leave empty) input sections (Role and Additional questions) in Support form.


tanyogeorgiev:
1. Go to [https://podkrepi.bg/en/support](https://podkrepi.bg/en/support)
2. Check some option
3. Click Next
4. Click Back
5. Uncheck option
6. Click on upper Additional Section Image (not Next button)
7. Click Next Button

kachar:
Code reference: https://github.com/daritelska-platforma/frontend/blob/0f074f5695c6516b4ec6adfad38cd71c53bb3baf/frontend/src/components/support-form/SupportForm.tsx#L166-L173

PR #258, opened 2021-06-10
[#229] Bugfix: Turned off stepper icons functionality in support form

RRoussev88:


kachar (approved):
The best way to avoid futher issues with step validation is to take out the jumps between steps and to use the Back / Next buttons to validate the form on each step.
Well done

Issue #148, opened 2021-04-04
Fix the /bin/true healthcheck in docker compose


miraclebg:
Within the 'roach-lb' container the following healthcheck must be corrected to fulfil the check:

`
healthcheck:
      test: /bin/true
      interval: 3s
      timeout: 10s
      retries: 10
`

This check is needed to make sure other dependent containers on 'roach-lb' start only when 'roach-lb' has fully started and it provides a stable connection to the database behind it.

This is needed to prevent a scenario where flyway db migrations tries to run before the database is live.

PR #276, opened 2021-06-18
removed cockroach-lb from env.dev

igoychev:
Removed cockroach-lb from env.dev, now lb is used for staging and prod only.

## Motivation and context
The change was needed to because in dev mode lb is not needed and also used to clear some of the code that was passing through the LB while the rest was going directly to the node.

details: 
 - The golang contact service was not loading environment variables correctly, only the default hardcoded ones
 - golang reference updated from 1.15 to 1.16
 - not-fixed: campaign service (.Net) still uses hardcoded db host in appsettings

Tested:
Using 'docker compose up' on local machine and manual check the web site works.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -46,14 +46,3 @@ services:
       dockerfile: ./flyway/Dockerfile
     volumes:
       - ./db/migrations:/flyway/sql
-
-  roach-0:
-    image: cockroachdb/cockroach:latest
-    container_name: ${COMPOSE_PROJECT_NAME?}-roach-0
-    restart: always
-    hostname: roach-0
-    command: start-single-node --insecure --listen-addr=roach-0:${DB_PORT} 
-    volumes:
-      - roach-0-data:/cockroach/cockroach-data
-    networks:
-      - backend-net
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -40,15 +40,20 @@ services:
     ports:
       - "${CAMPAIGN_GRPC_PORT?}:5001"
 
-  roach-lb:
-    ports:
-      - "${DB_PORT?}:26257"
-      - "3080:8080"
-      - "3081:8081"
-
   flyway:
     build:
       context: .
       dockerfile: ./flyway/Dockerfile
     volumes:
       - ./db/migrations:/flyway/sql
+
+  roach-0:
+    image: cockroachdb/cockroach:latest
+    container_name: ${COMPOSE_PROJECT_NAME?}-roach-0
+    restart: always
+    hostname: roach-0
+    command: start-single-node --insecure --listen-addr=roach-0:${DB_PORT} 
+    volumes:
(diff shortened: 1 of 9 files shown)


docker-compose.prod.yml
```diff
@@ -11,6 +11,54 @@ services:
...
+    # command: start-single-node --cluster-name=dp-crdb-cluster --logtostderr=WARNING --log-file-verbosity=WARNING --listen-addr=roach-0:26257 --advertise-addr=roach-0:26257 --certs-dir=/certs
+    volumes:
+      - roach-0-data:/cockroach/cockroach-data
+    #   - certs-roach-0:/certs
+    # depends_on:
+    #   - roach-cert
+    networks:
+      - backend-net
+
+  roach-lb:
```
kachar, line 37: @igoychev It seems we'll keep `roach-lb` on prod as it's still in the `docker-compose.prod.yml`

docker-compose.yml
```diff
@@ -125,7 +108,8 @@ services:
       DATABASE_NAME: ${DB_NAME?}
       DATABASE_USER: ${DB_USER?}
       DATABASE_PASSWORD: ${DB_PASS?}
-      COCKROACH_HOST: roach-0:${DB_PORT?}
+      COCKROACH_HOST: roach-0
```
kachar, line 111: Maybe we can use `${DB_HOST?}` instead of hardcoded `roach-0` and set that once in the env files

docker-compose.dev.yml
```diff
@@ -40,15 +40,20 @@ services:
...
-      - "3081:8081"
-
   flyway:
     build:
       context: .
       dockerfile: ./flyway/Dockerfile
     volumes:
       - ./db/migrations:/flyway/sql
+
+  roach-0:
```
kachar, line 50: If we have this configuration in `docker-compose.yml` we don't need it here as they are inherited

igoychev: yep, now moved back to the main docker-compose.yml - see commit: d3613d3
The start command for dev is "start-single-node", while in the docker-compose.prod.yml is replaced with "start"

docker-compose.yml
igoychev: yep, fixed in d3613d3

docker-compose.prod.yml
igoychev: I thought on this a bit more and for prod I actually created a proper cockroach cluster with having roach-0 and roach-1 nodes - see d3613d3. This also fixes #148 where the haproxy healtcheck was not working anyway.

miraclebg:
I believe the load balancer should be kept - if we run 1+ nodes (as i believe this is the intention in PROD) we will need it.

So i would propose to:

- Move the LB into the PROD docker-compose file
- Move / create the second node roach-1 in the PROD docker-compose file
- Add an alias roach-lb in the DEV docker-compose file - so clients can use the same alias to connect to from both envs

This should resolve the situation.

Although if this is to be moved to K8S soon probably there isn't a need for a second roach node now :)

Issue #214, opened 2021-06-03
[Feature] Add ids to heading sections


kachar:
In order to link directly to a heading please add `id` attributes and maybe a small anchor `#`

`#` to be shown on heading hover

to be done as component wrapping the `Typography` component passing rest of props to it

PR #278, opened 2021-06-20
Add Typography component with ID and link button

skilldeliver:
Implemented higher order component of Typography that supports optional link button. 

## Motivation and context

This change provides an easy way to link section (and share it)
Issue #214 
Edits are made on the styling of heading elements (margin is removed and padding is added because otherwise the heading hides under the navbar)

P.S:
Sorry if my implementation is not following best practices coming from different background
but I am ready to improve it :/

## Screenshots:

| image | image |


frontend/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }),
+)
+
+type TypographyParams = PropsWithChildren<TypographyProps> & {
+  id?: string
+  linkable?: boolean
+  component?: string
+}
+
+export default function Typography({ children, id, linkable, ...props }: TypographyParams) {
```
kachar, line 30: I'd suggest that we rename the component from `Typography` to `<AnchorTypography />` or `<Heading />` as it currently overlaps with the generic component with the same name coming from MaterialUI

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
kachar, line 1: This file is not needed as we don't have `package.json` in the root

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
kachar, line 1: This file is not needed as we don't have `package.json` in the root

frontend/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,49 @@
...
+    hideLinkIcon: {
+      visibility: 'hidden',
+    },
+  }),
+)
+
+type TypographyParams = PropsWithChildren<TypographyProps> & {
+  id?: string
+  linkable?: boolean
+  component?: string
```
kachar, line 27: This looks like a leftover that we don't use

frontend/src/components/common/Typography.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
```
kachar, line 28: Here's my proposal for these props

If user sets `linkable` to `true`, then the property `id` becomes mandatory

```suggestion
type Linkable = {
  id: string
  linkable: true
} | {  linkable?: false }
type TypographyParams = PropsWithChildren<TypographyProps & Linkable>
```

node_modules/.yarn-integrity
skilldeliver: Removed with [commit](https://github.com/podkrepi-bg/frontend/pull/278/commits/d72ade00d9c390f0756265dc65d1fb8e319bf89d)

yarn.lock
skilldeliver: Removed with [commit](https://github.com/podkrepi-bg/frontend/pull/278/commits/d72ade00d9c390f0756265dc65d1fb8e319bf89d)

frontend/src/components/common/Typography.tsx
skilldeliver: Thank you for the suggestion! Implemented with this [commit](https://github.com/podkrepi-bg/frontend/pull/278/commits/2acb40725303920a7176445428506f95466b2249)

frontend/src/components/common/Typography.tsx
skilldeliver: Hello, @kachar , that it is interesting one, it look like `component` is not part of `TypographyProps`, it does not work without adding manually `component` prop.  You can check how I implemented it [here](https://github.com/podkrepi-bg/frontend/pull/278/commits/2acb40725303920a7176445428506f95466b2249)

kachar: I was observing the `component` prop at their api docs https://material-ui.com/api/typography/

It's defined in a parent component of `Typography` - `OverridableComponent`

You can see the definition at:

https://github.com/mui-org/material-ui/blob/14d95615dd0b31c616abb760a29ec30a8fa9005f/packages/material-ui/src/OverridableComponent.d.ts#L12-L18

Maybe we have a way to inherit it

skilldeliver: Found that other people are experiencing this issue too - https://github.com/mui-org/material-ui/issues/19512

TLDR;
`component` is not defined in `TypographyProps` issue is closed - waiting for a new report on v5 of Material-UI.

What do you think about using `component: React.ElementType` and leaving it for now?

kachar: Let's leave it for now `component: React.ElementType`, it's a good solution without making it a strict `string` type

Thanks for the research :man_scientist: :microscope:

kachar (approved):

PR #283, opened 2021-06-23
Fix graphql-gateway build

imilchev:
- Fixed building of graphql-gateway
  + Removed auto-generated files from the repo
  + Referenced go-types module
  + Added a Taskfile with lint and build steps
- Added a workflow for building the graphql-gateway
- Added a short README
- Added Dockerfile
- Split workflows for back-ends to 1 per module

**New dependencies**:

- [go-types](https://github.com/podkrepi-bg/types/tree/master/go-types) : Podkrepi.bg types


backend/graphql-gateway/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ghcr.io/podkrepi-bg/go-devcontainer:v1.0.0 AS build
+WORKDIR /app
+COPY . /app
+
+RUN task build
+
+FROM alpine:3.14
```
imilchev, line 7: I would like to add some sensible defaults for the 2 env variables that this module currently has - `PORT` and `ACCOUNT_SERVICE_URL`. @kachar what should those be?

imilchev:
After this PR is approved I will update the contact module as well. Such that it has a similar structure. At that point I think we can delete the `Taskfile.yml` from the `backend` folder and all the `proto` files

backend/graphql-gateway/Dockerfile
kachar: This url should be the way graphql gateway contacts the grpc services like account, campaigns, etc

Probably it will be provided from the api gateway

Not sure there will be a default value for it as it depends on the stack setup, not on this individual component

imilchev: Then I will not mention it then and will just add it to the README

imilchev:
Rebased on master

PR #286, opened 2021-06-24
Update contact build

imilchev:
- Added a Taskfile with lint, build, test and watch steps
- Added a short README
- Renamed module `github.com/podkrepi-bg/v2` -> `github.com/podkrepi-bg/contact`
- Updated workflow for contact module to re-use devcontainer
- Updated Dockerfile

### DIFF
--- a/backend/contact/api/config/config.go
+++ b/backend/contact/api/config/config.go
@@ -1,8 +1,8 @@
 package config
 
 import (
-	"github.com/spf13/viper"
 	"github.com/gofiber/fiber/v2"
+	"github.com/spf13/viper"
 )
 
 type Config struct {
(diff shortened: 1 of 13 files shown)


backend/contact/api/config/config.go
```diff
@@ -42,17 +42,16 @@ func (config *Config) setDefaults() {
 
 	// Set default App configuration
 	config.SetDefault("APP_ADDR", ":5000")
-	config.SetDefault("APP_ENV", "local")
```
imilchev, line 45: @kachar was there a reason for this setting? I saw it's not used and deleted it

kachar: `APP_ENV` is just a way to differentiate the running environment. It's not used anywhere as far as i recall.

kachar (approved):

PR #292, opened 2021-06-25
Update account build

imilchev:
- Added devcontainer for module-accounts
- Added workflow for module accounts


imilchev:
Added this PR just to test the workflow that I added. No review needed here.

PR #300, opened 2021-06-30
Add (404) not found page 

skilldeliver:
This PR introduces custom 404 error page with message supporting locales and button back to home.

Screenshot of the page:
image


frontend/src/components/errors/assets/NotFoundIllustration.tsx
```diff
@@ -0,0 +1,141 @@
+import React from 'react'
+
+import theme from 'common/theme'
+
+export default function NotFoundIllustratio() {
```
kachar, line 5: There's a typo in the name of the component

```suggestion
export default function NotFoundIllustration() {
```

frontend/src/components/errors/NotFoundPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+)
+
+export default function NotFoundPage() {
+  const { t } = useTranslation()
+  const router = useRouter()
+
+  const classes = useStyles()
+
+  return (
+    <Grid container className={classes.root}>
```
kachar, line 29: Maybe we can introduce the header navigation while we're showing the 404 page

kachar (approved):
@skilldeliver Kudos for the quick integration, looks awesome! 

Added some notes, let's discuss if needed - we can extract that as separate task if you think it will bloat the scope of this one.

frontend/src/components/errors/NotFoundPage.tsx
skilldeliver: Yup, we can do that!

What do you think only the header navigation or the whole layout component (header + footer)?

kachar: There's a catch with the whole layout  - sometimes when it depends on dynamic component it cannot be rendered during Static Site Generation (SSG) phase and `_error.tsx` is always rendered during that phase.
Anyway I don't we have any dynamic elements in the layout so far, so it wont be a problem to render it in the error page. 

You might try it and see if it works with no extra effort.

skilldeliver: Yeah, works fine with the current state of the layout.

If we include the error pages with the Layout maybe we need to leave somewhere note to mention that introducing dynamic component will break the error pages?

image

frontend/src/components/errors/NotFoundPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <Grid container className={classes.root} xs={12}>
+        <Grid container direction="column" alignItems="center" justify="center" spacing={4} xs={12}>
+          <Grid container item xs={8} sm={6} lg={4}>
+            <NotFoundIllustration />
+          </Grid>
+          <Grid item>
+            <Typography variant="body2">{t('errors:404Message')}</Typography>
+          </Grid>
+          <Grid container item alignItems="center" justify="center">
+            <Button
```
kachar, line 39: Maybe we can use `<LinkButton />` from `components/common/LinkButton.tsx` instead of plain button with `onClick` handler as it will render a more standard `<a href />` component

At the moment the back button is not functional until the page is completely rendered.

frontend/src/components/errors/NotFoundPage.tsx
kachar: Looks awesome! Let's keep it like that.

Don't worry about future updates if they break the app it will be shown during PR checks as we run `next build` during that phase

frontend/src/components/errors/NotFoundPage.tsx
skilldeliver: Good idea, first I tried with the `<LinkButtonIcon />` and I saw that it is rendering something that I don't need. :D

But now I realized that `<LinkButton />` will do the job, thanks! :)

kachar: Awesome! I think we're good to go :tada: 
Let's roll

PR #330, opened 2021-07-12
Replace using of appseting.json with env variables for BE campaigns

dguguchkov:
Replace using of appseting.json for getting db connection string with environment variables. 

Currently module campaign in BE reading db connection string from configuration which read it from appsetings.json. The PR replace reading of this JSON with reading env variables and build the string.

### DIFF
--- a/backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
+++ b/backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
@@ -1,0 +1,15 @@
+namespace Podkrepibg.Campaigns.Infrastructure.Configuration
+{
+    using System;
+
+    public class DbConfiguration
+    {
+        private static string DbHost = Environment.GetEnvironmentVariable("DB_HOST");
+        private static string DbPort = Environment.GetEnvironmentVariable("DB_PORT");
+
+        public static string GetConnectionString()
+        {
+            return $"Host={DbHost};Port={DbPort};Database=app;Username=root;Password=1234";
+        }
+    }
+}
--- a/backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DependencyInjection.cs
+++ b/backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DependencyInjection.cs
@@ -1,20 +1,19 @@
 namespace Podkrepibg.Campaigns.Infrastructure.Configuration
 {
     using Microsoft.EntityFrameworkCore;
-    using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.DependencyInjection;
     using Podkrepibg.Campaigns.Application.Data;
     using Podkrepibg.Campaigns.Infrastructure.Persistence;
 
     public static class DependencyInjection
     {
-        public static IServiceCollection AddInfrastructure(this IServiceCollection services, IConfiguration configuration)
+        public static IServiceCollection AddInfrastructure(this IServiceCollection services)
         {
             services.AddScoped<IApplicationDbContext>(provider => provider.GetService<CampaignsContext>());
             services.AddScoped<IApplicationReadOnlyDbContext>(provider => provider.GetService<CampaignsContext>());
 
             services.AddDbContext<CampaignsContext>(
-                options => options.UseNpgsql(configuration.GetConnectionString("CampaignDb")));
+                options => options.UseNpgsql(DbConfiguration.GetConnectionString()));
 
             return services;
         }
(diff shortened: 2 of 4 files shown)


backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
```diff
@@ -0,0 +1,15 @@
+namespace Podkrepibg.Campaigns.Infrastructure.Configuration
+{
+    using System;
+
+    public class DbConfiguration
+    {
+        private static string DbHost = Environment.GetEnvironmentVariable("DB_HOST");
```
preslavgerchev, line 7: You should get those from the `Configuration`, not the `Environment`.

backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
```diff
@@ -0,0 +1,15 @@
...
+        private static string DbPort = Environment.GetEnvironmentVariable("DB_PORT");
+
+        public static string GetConnectionString()
+        {
+            return $"Host={DbHost};Port={DbPort};Database=app;Username=root;Password=1234";
```
preslavgerchev, line 12: Can you also make the name and password configurable

backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
preslavgerchev: See [here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0#environment-variables). It will also make testing significantly easier as well as we can then mock the configuration, load it from a file.

preslavgerchev (changes requested):

backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
imilchev: In my opinion everything should be configurable. Even the user will be changing for the actual deployment

backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
dguguchkov: > You should get those from the `Configuration`, not the `Environment`.

In the story is written that the string values should be read from env : 
The connection string needs to be created from the DB_HOST and DB_PORT coming from .env files.

xenry: yep, I also support using the configuration infrastructure, it will be more flexible that way - for example we can override db_port via a commandline argument if we want to. You just need to add a `EnvironmentVariablesConfigurationProvider`.

backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DependencyInjection.cs
```diff
@@ -14,9 +14,10 @@ public static IServiceCollection AddInfrastructure(this IServiceCollection servi
             services.AddScoped<IApplicationReadOnlyDbContext>(provider => provider.GetService<CampaignsContext>());
 
             services.AddDbContext<CampaignsContext>(
-                options => options.UseNpgsql(configuration.GetConnectionString("CampaignDb")));
+                options => options.UseNpgsql($"Host={configuration.GetSection("DB_HOST").Value};Port={configuration.GetSection("DB_PORT").Value};Database={configuration.GetSection("DB_NAME").Value};Username={configuration.GetSection("DB_USER").Value};Password={configuration.GetSection("DB_PASS").Value}"));
```
dguguchkov, line 17: Can it be like this one or it is better to move the building the string into separate class ?

backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
dguguchkov: Now configuration.GetSection is used.

backend/module-campaigns/src/Podkrepibg.Campaigns.Infrastructure/Configuration/DbConfiguration.cs
dguguchkov: Now all the strings are read from env

preslavgerchev:
@dguguchkov is this PR still active?

preslavgerchev:
Closing this in favor of a new repo where it's fixed

PR #341, opened 2021-08-28
Replace existing migrations with postgresql

Christian-Vasilev:
Since we are migrating to PostgreSQL, all existing migrations except the accounts are now set to work with postgresql.
The new migrations can be found in `db/migrations/postgresql`


imilchev:
As I already mentioned on Discord these changes are probably not needed and you should make sure they are applied in the repo for module-contacts. Not sure if @kachar agrees...

imilchev:
ah, I now see that all migrations are in the same folder at the moment. Looks good to me but I would like @kachar  to have a look as well

imilchev (approved):

db/migrations/postgresql/util/V1.0.0.2__fn_set_deleted_at.sql
```diff
@@ -0,0 +1,25 @@
...
+LANGUAGE plpgsql;
+-- by default all functions are accessible to the public, we need to remove that
+-- and define our specific access rules
+REVOKE ALL PRIVILEGES ON FUNCTION campaign.soft_delete_record() FROM public;
+ALTER FUNCTION campaign.soft_delete_record() OWNER TO postgres;
+
+COMMIT;
+
+
+ALTER USER postgres SET search_path TO "$user", app, campaign, util;
```
imilchev, line 22: do we really want this? I would expect that in an ideal scenario each module will have its own database user which will be limited to a single database

db/migrations/postgresql/_init/V1.0.0.2__schemas.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN;
+
+-- SCHEMAS WILL BE CREATED INSIDE OF THE CURRENTLY UED DATABASE;
+CREATE SCHEMA IF NOT EXISTS "app";
```
imilchev, line 4: since we are now sort of re-creating the migrations, isn't it a good idea to split them per module? These modules will eventually be moved to different repos. We cannot keep all migrations in a single folder. It also doesn't make much sense since ideally deploy 1 module can potentially change only that module's database and not all others

imilchev (changes requested):

db/migrations/postgresql/_init/V1.0.0.2__schemas.sql
kachar: We can leave all tables in the public schema, this will help us split them in individual databases after that. At the moment we don't have or need any cross schema queries

db/migrations/postgresql/_init/V1.0.0.2__schemas.sql
```diff
@@ -0,0 +1,11 @@
...
+CREATE SCHEMA IF NOT EXISTS "campaign";
+CREATE SCHEMA IF NOT EXISTS "util";
+
+-- DROP THE DEFAULT SCHEMA IF IT EXISTS
+DROP SCHEMA IF EXISTS "public";
```
kachar, line 9: There are many db functions naturally existing in `public` schema. When we drop it we might miss those functions or types.

db/migrations/postgresql/_init/V1.0.0.2__schemas.sql
imilchev: Ideally each module should have its own database that is not accessed by others. If we change the migrations to PostgreSQL now, then later we need to re-write the migrations yet another time to split the databases. Not sure if that's worth it

preslavgerchev: I agree that splitting those per module would be ideal, it offers separation.

preslavgerchev: This is the schema for the contact module, right? I suggest we rename it to `contacts`, makes it more clear what it's related to.
```suggestion
CREATE SCHEMA IF NOT EXISTS "campaigns";
```

db/migrations/postgresql/app/V2.0.0.4__table_support_requests.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS app.support_requests (
```
preslavgerchev, line 3: Is there a specific reason why the schema/table names are not in quotes here (or are quoted in other places)?

db/migrations/postgresql/app/V2.0.0.4__table_support_requests.sql
```diff
@@ -0,0 +1,12 @@
...
+    "id"            UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    "person"        JSONB,
+    "support_data"  JSONB,
+    "created_at"    TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
preslavgerchev, line 7: I am not sure if it's not better to use `TIMESTAMP` and store/serve everything in UTC? 

Using the `TZ` part definitely makes it easier for the back-end but API consumers might have problems if we return the timestamps in the time zone of the server - what if the API only returns the datetime without the zone offset? Then the clients cannot reliably convert this to their own timezones.

To be safer it's probably just better to store it all in UTC 😄

db/migrations/postgresql/campaign/V4.0.0.4__table_campaigns.sql
```diff
@@ -0,0 +1,57 @@
...
+        REFERENCES "campaign"."campaign_types" ("id")
+          ON DELETE CASCADE,
+    CONSTRAINT "FK_Campaigns_Beneficiaries_BeneficiaryId"
+      FOREIGN KEY ("beneficiary_id")
+        REFERENCES "campaign"."beneficiaries" ("id")
+          ON DELETE CASCADE
+);
+
+-- Triggers
+CREATE TRIGGER trigger_set_deleted_at
```
preslavgerchev, line 38: Is there any reason why we can't just set the `deleted_at` timestamp from the application? Wiring this on a database level means it's harder to change down the line (should we need to).

kachar: @preslavgerchev I can imagine a scenario when the data is soft deleted from the db without properly set `deleted_at`. Having it automated ensures data consistency without hassle

db/migrations/postgresql/app/V2.0.0.4__table_support_requests.sql
kachar: That feels as a minor issue as both syntaxes are valid

preslavgerchev: Yes, that's definitely a nitpick, feel free to ignore it @Christian-Vasilev

db/migrations/postgresql/_init/V1.0.0.2__schemas.sql
kachar: If we're going to split databases it makes no sense to split also by schemas. 

It's not hard to send everything to `public` schema and send the module migration files to their own db once we have that.

db/migrations/postgresql/app/V2.0.0.4__table_support_requests.sql
imilchev: agreed

imilchev: `For timestamp with time zone, the internally stored value is always in UTC (Universal Coordinated Time, traditionally known as Greenwich Mean Time, GMT). An input value that has an explicit time zone specified is converted to UTC using the appropriate offset for that time zone. If no time zone is stated in the input string, then it is assumed to be in the time zone indicated by the system's timezone parameter, and is converted to UTC using the offset for the timezone zone.`

This is from the PostgreSQL manual (see [here](https://www.postgresql.org/docs/9.1/datatype-datetime.html)). It seems like this is the correct type and we just need to make sure that the module is actually providing the timestamp when saving data

db/migrations/postgresql/app/V2.0.0.1__table_contacts.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS app.contacts (
+    "id"            UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```
imilchev, line 4: This is only supported in PostgreSQL > 13.0. Please document that somewhere

db/migrations/postgresql/_init/V1.0.0.2__schemas.sql
imilchev: I spent some time reading about what is best - 1 db and multiple schemas or multiple dbs. There were mixed answers but I am convinced that we can make 1 db with multiple schemas - 1 per module and everything will work out just fine. I will implement that for the contact module and add it to its own repo

db/migrations/postgresql/_init/V1.0.0.2__schemas.sql
Christian-Vasilev: Drop was removed, thanks.

db/migrations/postgresql/app/V2.0.0.4__table_support_requests.sql
Christian-Vasilev: We should keep it consistent. I will update and replace everything in quotes, thanks for noticing it.

db/migrations/postgresql/app/V2.0.0.1__table_contacts.sql
Christian-Vasilev: Readme will be updated with the minimum required version using badge

db/migrations/postgresql/app/V2.0.0.4__table_support_requests.sql
imilchev: You can ignore the migration for the contact module. I already fixed it in the other repo https://github.com/podkrepi-bg/module-contact/tree/main/database/migrations

db/migrations/postgresql/_init/V1.0.0.2__schemas.sql
kachar: One db, multiple schemas:
- will ease the backup and restore procedures
- good separation
- if we need to do multi-schema reports - will be possible
- running one db instance with one db

Multiple dbs in single instance:
- true separation
- somewhat easy backup/restore procedures
- cross-db queries are possible (via `postgres_fdw`), but not advisable

Multiple dbs in separate instances:
- true separation
- backup separately, restore separately
- no cross-db queries

I'm okay we continue with approach :one: as long as it fits the needss of separation.

If we decide to go :two: or :three: , then it makes sense to use the `public` schema.

Christian-Vasilev: If we go with 2 or 3 we will be very limited in future. We neeed to design everything very well so we wont have the need of any db queries

preslavgerchev:
Are the contact migrations still needed here if @imilchev has already moved them to their own repo?

Else lgtm!

imilchev:
> Are the contact migrations still needed here if @imilchev has already moved them to their own repo?
> 
> 
> 
> Else lgtm!

Depends on how the current deployment will be migrated to the new module. @kachar should double-check what is a convenient way to fix it.

kachar:
@imilchev I'd say we can merge this and have the migraitons on master for a bit, before completely shut down contacts module in frontend repo. There's no point in hanging PRs

This PR looks like it may land on master now :+1:

imilchev:
Fine with it!

Issue #328, opened 2021-07-06
Create kubernetes deployments for all modules


kachar:

imilchev:
I created a deployment file for the front-end in #342. I think we should rename this issue to only be relevant for the front-end since we are going to be splitting the modules in their own repos. The contact module is also already done (see [here](https://github.com/podkrepi-bg/module-contact/tree/main/manifests)).

Issue #233, opened 2021-06-06
[Bug][DevOps] Frontend build fails


marchev:
As the issue title states, the frontend build fails on the `master` branch.

**Steps to reproduce:**
1. `git clone git@github.com:daritelska-platforma/frontend.git && cd frontend`
2. `ln -s .env.example .env`
3.  `docker-compose up -d `

**Expected result:**
The project is built successfully.

**Actual result:**
The `frontend` module build fails.

**Environment:**
- OS: **macOS Big Sur**
- Docker Desktop: **v3.3.3 (64333)**
- Docker Engine: **v20.10.6**
- Docker Compose: **v1.29.1**

**Logs:**
<details>
  <summary>Click to expand!</summary>

```
Docker Compose is now in the Docker CLI, try `docker compose up`

Building frontend
[+] Building 34.5s (9/10)
 => [internal] load build definition from Dockerfile                                                                                                     0.0s
 => => transferring dockerfile: 37B                                                                                                                      0.0s
 => [internal] load .dockerignore                                                                                                                        0.0s
 => => transferring context: 34B                                                                                                                         0.0s
 => [internal] load metadata for docker.io/library/node:14-alpine                                                                                        1.4s
 => [base 1/4] FROM docker.io/library/node:14-alpine@sha256:f07ead757c93bc5e9e79978075217851d45a5d8e5c48eaf823e7f12d9bbc1d3c                             0.0s
 => [internal] load build context                                                                                                                        0.0s
 => => transferring context: 14.12kB                                                                                                                     0.0s
 => CACHED [base 2/4] WORKDIR /app                                                                                                                       0.0s
 => CACHED [base 3/4] COPY package.json yarn.lock /app/                                                                                                  0.0s
 => CACHED [base 4/4] RUN apk --no-cache add curl                                                                                                        0.0s
 => ERROR [dependencies 1/1] RUN yarn install --production &&   cp -R node_modules /prod_node_modules &&   yarn install --production=false              33.0s
------
 > [dependencies 1/1] RUN yarn install --production &&   cp -R node_modules /prod_node_modules &&   yarn install --production=false:
#9 0.381 yarn install v1.22.5
#9 0.435 [1/4] Resolving packages...
#9 0.566 [2/4] Fetching packages...
#9 23.66 info fsevents@2.3.2: The platform "linux" is incompatible with this module.
#9 23.66 info "fsevents@2.3.2" is an optional dependency and failed compatibility check. Excluding it from installation.
[...]

PR #342, opened 2021-08-29
Added Kubernetes deployment file for front-end

imilchev:
- Added a Kubernetes deployment file for the front-end.
- Fixed a bug where apiUrl was hardcoded in front-end config

I have 2 points that I will need help with. @kachar perhaps you can help with them...
- The front-end image is 700MB. I see python, make g++ are being installed in the container which shouldn't be necessary in my opinion
- There are tons of ENV vars that do not really follow any convention and some of them seem to be duplicated. I think we should revise them, reduce them where possible and definitely document them in a README. I spent 15mins trying to figure out how to run the front-end on a port different from 3040 but couldn't figure it out. I don't think one should be aware of the implementation to be able to configure the module

P.S.: You can see the deployed version on https://dev.podkrepi.bg


imilchev:
Fixes #328 (if we consider it only for the front-end)

imilchev:
Moving the installation of python, g++ etc to the build container reduced the image size by ~200MB. The `node_modules` folder is 350MB and this is the output for the largest folders:
```
11M	./highcharts/modules
12M	./@material-ui/core
13M	./typeorm/browser
18M	./@sentry/cli
23M	./@material-ui/icons/esm
24M	./next/dist/compiled
26M	./typeorm
31M	./next/dist
32M	./@sentry
34M	./next
48M	./highcharts
69M	./@material-ui/icons
90M	./@material-ui
```

I guess if we agree to move the installation of the extra software we should be fine for now. However, I guess we install way too many things as modules. I highly doubt we are using the whole material UI, for example.

frontend/next.config.js
```diff
@@ -25,7 +25,7 @@ const moduleExports = {
     DISCORD_CLIENT_ID: process.env.DISCORD_CLIENT_ID,
     DISCORD_CLIENT_SECRET: process.env.DISCORD_CLIENT_SECRET,
     services: {
-      apiUrl: 'http://api.podkrepi.localhost:5000',
```
kachar, line 28: @imilchev That's actually not a bug of hardcoded url but alias to the api container in the local docker-compose network


https://github.com/podkrepi-bg/frontend/blob/f58120463810906c72c9b009d04bfc842f2f4f2b/docker-compose.yml#L47-L49

If we'll use another `API_URL` we might define which that is - at the moment it seems to be `CONTACT_API_URL` pointing to the contact module

If we merge this version to `master` and deploy to current prod stack the support requests won't be available as the url will not be set in the deployment  https://github.com/podkrepi-bg/frontend/blob/f58120463810906c72c9b009d04bfc842f2f4f2b/.github/workflows/deploy-prod.yml#L125

kachar (approved):

kachar:
>  I spent 15mins trying to figure out how to run the front-end on a port different from 3040 but couldn't figure it out.

It's defined in the dockerfile itself as exposed port
https://github.com/podkrepi-bg/frontend/blob/c1efdb47ab835f0f84f0fed06a7513cd93c180a6/frontend/Dockerfile#L10

And then we have it once again in the starting script which makes the node server listen to the same port

https://github.com/podkrepi-bg/frontend/blob/def1271054cd937362da6775839143d852f56c8e/frontend/package.json#L9

kachar:
> I see python, make g++ are being installed in the container which shouldn't be necessary in my opinion

I totally agree we don't need them at the moment. They were added for some reason which might not be valid anymore.

Check this out:
- https://github.com/podkrepi-bg/frontend/commit/c1efdb47ab835f0f84f0fed06a7513cd93c180a6
- https://github.com/podkrepi-bg/frontend/issues/233

imilchev:
> > I see python, make g++ are being installed in the container which shouldn't be necessary in my opinion
> 
> I totally agree we don't need them at the moment. They were added for some reason which might not be valid anymore.
> 
> Check this out:
> 
> * [c1efdb4](https://github.com/podkrepi-bg/frontend/commit/c1efdb47ab835f0f84f0fed06a7513cd93c180a6)
> * [[Bug][DevOps] Frontend build fails #233](https://github.com/podkrepi-bg/frontend/issues/233)
> 
> ```
> #9 30.24 Command: node-gyp-build
> #9 30.24 Arguments:
> #9 30.24 Directory: /app/node_modules/bufferutil
> #9 30.24 Output:
> #9 30.24 gyp info it worked if it ends with ok
> #9 30.24 gyp info using node-gyp@5.1.0
> #9 30.24 gyp info using node@14.17.0 | linux | arm64
> #9 30.24 gyp ERR! find Python
> #9 30.24 gyp ERR! find Python Python is not set from command line or npm configuration
> #9 30.24 gyp ERR! find Python Python is not set from environment variable PYTHON
> ```

This one I fixed here c93c260f8816a486221632d5b5d4dacd50bb0b8c

frontend/next.config.js
imilchev: It's still strange that there is some hard-coded URL in this file. What is it supposed to point to? This will not work for sure on the cluster as we don't have a route to this URL so I guess we need another setting for it.

imilchev:
> > I spent 15mins trying to figure out how to run the front-end on a port different from 3040 but couldn't figure it out.
> 
> It's defined in the dockerfile itself as exposed port
> https://github.com/podkrepi-bg/frontend/blob/c1efdb47ab835f0f84f0fed06a7513cd93c180a6/frontend/Dockerfile#L10
> 
> And then we have it once again in the starting script which makes the node server listen to the same port
> 
> https://github.com/podkrepi-bg/frontend/blob/def1271054cd937362da6775839143d852f56c8e/frontend/package.json#L9

I would expect that this port is truly configurable. It's strange that in the package json we hard-code it to 3040. Also my comment was more in general about all the settings that exist. I would expect the front-end to have a table similar to (this one)[https://github.com/podkrepi-bg/module-contact]. I already created several tables like this for the different modules but I don't think I can create one that is accurate for this module. Also there are strange namings like `API_URL` and then `REST_API_PORT`, which seem to be related to the same thing but don't really follow the same naming convention. If a table is created to document them, we will have a better overview of what is going on and on top of that we might see strange things like what I just mentioned.

frontend/next.config.js
kachar: @imilchev As I revised it there is an environment variable set in `.env.example` which refers to the `API_URL` with the correct value

https://github.com/podkrepi-bg/frontend/blob/bc8e1bb1867e1cb5d3ec172861fd485687bb1271/.env.example#L25-L26


The problem comes from the production env values of the same variable

https://github.com/podkrepi-bg/frontend/blob/bc8e1bb1867e1cb5d3ec172861fd485687bb1271/.env.prod#L24-L25

It points to `api.podkrepi.bg` which is invalid address at this point. If we want to use the `process.env.API_URL` we need to update it inside `.env.prod` to match the current hardcoded value so we don't break prod deployments.

kachar:
> I would expect that this port is truly configurable. It's strange that in the package json we hard-code it to 3040. Also my comment was more in general about all the settings that exist. 

I see two options:

1. Having in mind that we override the default port 3000 with custom port 3040 we can fallback to the default port for the node container and remap `80 -> 3000`
2. Provide custom port using env variable when we run `yarn start` like:

```json
{
  "scripts": {
    "start": "next start -p $APP_PORT",
  }
}
```

`APP_PORT=3045 yarn start` in bash
or
`CMD [ "APP_PORT=3040", "yarn", "start" ]` in `Dockerfile`


However I don't think having that port hardcoded in the container is a problem as long as we can remap it shen using the container in different environment. There are many many node containers that use the default port 3000 internally.


---


> I would expect the front-end to have a table similar to [this one](https://github.com/podkrepi-bg/module-contact).

That's gonna be an awesome addition and I'm totally up for it

igoychev (approved):

PR #357, opened 2021-09-18
Improvement. Creating of .env.local will be mandatory.

gonaumov:
## Motivation and context

We need to ensure creating of .env.local file. This simple package script take care of this. 
I am using **shx** in order to make in multi-platform.


kachar (approved):
Awesome increment :rocket:  No more broken environments and unnecessary investigations

PR #360, opened 2021-09-19
Disable sentry reports for localhost

gonaumov:
## Motivation and context

At the moment all localhost errors are getting reported to Sentry and this burns up our limits. We need to enable Sentry only for staging/prod env.


kachar:
The proposed behavior seems to be equal to the current one, I'm not sure it really fixes the problem. I liked the idea of blacklisting `localhost` and `127.0.0.1` from sentry reports. Let's roll with that ?

gonaumov:
@kachar I've just pushed the requested changes.

kachar (approved):

PR #396, opened 2021-10-30
Relocate Copy IBAN button

ani-kalpachka:
Align Copy IBAN button right next to the IBAN and on the same level of Next button.

## Screenshots:

| before| after |


kachar:
This task has been resolved recently in master
image

Let's figure out if we still need this PR or not

PR #405, opened 2021-11-06
Run unit tests during github PR workflows

StanislavSt:
## Motivation and context
Not sure against which node version we should test, so I'm testing against 12.x and 14.x using matrix strategy.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,4 +36,4 @@ jobs:
       if: steps.yarn-cache.outputs.cache-hit != 'true'
 
     - name: Run tests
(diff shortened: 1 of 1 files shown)


StanislavSt:
Also it's hitting the cache as expected. ~1min to 1,5mins without yarn-cache, 25s with yarn-cache. 
Maybe we should add yarn-cache step as well to `lint.yaml`. No idea what is our minutes limit in github.

kachar (approved):
Great addition :rocket:

> No idea what is our minutes limit in github.

No limits for OSS projects afaik

PR #407, opened 2021-11-07
404 feature flags

StanislavSt:
## Motivation and context
Use the `<Feature><Feature/>` component to wrap any component and add it to `next.config.js`.
This will only matter for production.

This solution is one of the possibly many ways to do this.
Another way to do it is with custom hook, which is more flexible(Can be used not only for components). Will have more code inside the components though.

Added `    FEATURE_ENABLED: {
      CAMPAIGN: false,
    },` to  `publicRuntimeConfig`.


StanislavSt:
Not really happy with the namings. If you have better ideas, please share them 👍 .

atanasster:
> Not really happy with the namings. If you have better ideas, please share them 👍 .

Naming looks fine to me, I have seen names that are playing with `Feature` and `Flags`  - `FeatureFlags`, `Flagged`, `FeatureToggle`. But I also struggle with names, so not the best feedback :)

The functionality looks great.

src/components/common/Feature.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  if (isFeatureEnabled || !isProduction) {
+    return <>{children}</>
+  }
+
+  //Add custom 'In development' page below
+  return (
+    <Layout
+      githubUrl="https://github.com/podkrepi-bg/frontend/tree/master/src/components/campaigns/CampaignsPage.tsx"
+      figmaUrl="https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=5100%3A21216">
```
kachar, line 25: We can get rid of these urls as they point to a different component

```suggestion
    <Layout>
```

kachar: @StanislavSt We need to consider hiding elements with the feature without rendering a whole layout.

As an example we need to hide the "Campaigns" nav link from the `MainNavMenu` component. Placing a Layout there will break the navigation.

Maybe we need a hook for that location as you've mentioned

next.config.js
```diff
@@ -42,6 +42,9 @@ const moduleExports = {
     APP_URL: process.env.APP_URL,
     GTM_ID: 'GTM-TWQBXM6',
     keycloakConfig,
+    FEATURE_ENABLED: {
+      CAMPAIGN: false,
```
kachar, line 46:
```suggestion
      CAMPAIGN: process.env.FEATURE_CAMPAIGN ?? false,
```

kachar:
@StanislavSt Added some commits to hide the nav menu item

I think we're good to go to prod with this one :rocket:

kachar (approved):

PR #418, opened 2021-11-27
Replace Keycloak integration with NextAuth Keycloak provider

kachar:
## Motivation and context

Keycloak creates a redirect loop in order to authenticate user on every server side render of the page.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "presets": ["next/babel"]
-}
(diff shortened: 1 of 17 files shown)


kachar:
Adding label "on hold" as we're exploring another solution with customized look'n'feel of the default keycloak login/register form instead of having the whole authentication mechanism inside the frontend. This might help us iterate faster and still keep certain functionalities like checking the nested permissions from the jwt token.

kachar:
Closing this PR as we'll stick with Keycloak authentication forms with custom design

PR #456, opened 2022-02-01
Added bootcampers layout

PepiPetrov:
This is the first task for bootcamp.


src/components/bootcamp/layout/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+      style={{
+        position: 'fixed',
+        left: '0',
+        bottom: '0',
+        width: '100%',
+        backgroundColor: '#46dbf2',
+        textAlign: 'center',
+        color: 'white',
+        height: '6.8%',
+      }}>
```
kachar, line 43: This style is inline and won't be cached during server side rendering.

You can do the following instead and move the styles to the container using `sx` property which is cached.

```tsx
<Container maxWidth="lg" component="footer" sx={...}>
```

src/components/bootcamp/search/email.tsx
```diff
@@ -0,0 +1,84 @@
...
+export type SearchFormProps = { initialValues?: SearchFormData }
+
+export default function SearchByPhone({ initialValues = defaults }: SearchFormProps) {
+  const [res, setRes] = React.useState([])
+
+  const onSubmit = async (values: SearchFormData, { resetForm }: FormikHelpers<SearchFormData>) => {
+    try {
+      setRes([])
+      const data = await axios.get(
+        'http://localhost:5010/api/bootcamp/search/email/' + values.keyword,
```
kachar, line 31: Links like that are way better when we manage them in a single location like `routes.ts` and use them as functions.

This helps us refactor the URL more easily later on

src/components/campaign-types/layout/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+function Copyright(props: CopyrightProps) {
+  return (
+    <Typography
+      variant="body2"
+      color="text.secondary"
+      align="center"
+      style={{ marginBottom: '0.5%', color: 'white' }}>
+      <Link color="inherit" style={{ fontSize: '13px', textDecoration: 'none' }}>
+        © {new Date().getFullYear()} Pordkrepi.bg Campaign types module {props.description}
```
kachar, line 18: This footer component seems to be duplicated and exists more than once. It's better to have them in a single place and reuse it whenever appropriate.

src/components/city/layout/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+function Copyright(props: CopyrightProps) {
+  return (
+    <Typography
+      variant="body2"
+      color="text.secondary"
+      align="center"
+      style={{ marginBottom: '0.5%', color: 'white' }}>
+      <Link color="inherit" style={{ fontSize: '13px', textDecoration: 'none' }}>
+        © {new Date().getFullYear()} Pordkrepi.bg Cities Module {props.description}
```
kachar, line 18: One more footer in this universe :)

kachar (approved):
Well done @PepiPetrov 

Kudos for the energy spent on this :zap: 

The code style and formatting are top :tophat: 

There are some minor ideas of improvement that I've tried to point out in the comments.

The major things we should be looking forward to reduce are:

- code duplication (multiple footer components with the same or similar contents)
- test entity `bootcamp` together with real-life entities `city` and `campaignTypes`
- hardcoded links to API endpoints (normally you don't have to specify the domain and it will still work
-  poor task/PR description and no screenshots or video attached

PR #473, opened 2022-02-06
Enable static pages with dynamic header

quantum-grit:
This change enables static pages generation at build time, while the header is still loaded dynamically at runtime. 
 - The dynamic header is imported as per the With No SSR section here: https://nextjs.org/docs/advanced-features/dynamic-import
 - The static pages are marked with full dots on the below screenshots. 

## Screenshots:
Screenshot 2022-02-06 at 16 00 29

## Testing
manual on localhost only


kachar (approved):
Bull's eye :dart: 

Straight to the point with adequate technical depth :fishing_pole_and_fish:

Issue #529, opened 2022-02-22
[Vault] Create dashboard for Vault entity


stilian-nikolaev:
Adding Vault model to the admin page

closed by [#572  ](https://github.com/podkrepi-bg/frontend/pull/572)

Caution! Only coordinators can manipulate the data.
In order to test the functionality at its current state you have to first make a coordinator entity with the same id of your profile
Check the following examples:
![Screenshot (329)](https://user-images.githubusercontent.com/74594156/156359751-88ae3e60-f149-4edb-8a39-871a3d494946.png)
![Screenshot (330)](https://user-images.githubusercontent.com/74594156/156359879-af869497-a75b-45dc-be7c-ea2c11645e93.png)

Furthermore, only the coordinators of the campaign connected to the vault can edit or delete it.
So you should also create a campaign in which the coordinator ID is the same as your profile ID
Check the following example:
![Screenshot (331)](https://user-images.githubusercontent.com/74594156/156360761-32a6d9ae-1896-47ce-b556-0c739388e0ad.png)
And when creating a vault just use the ID of the campaign you have created like this:
![Screenshot (333)](https://user-images.githubusercontent.com/74594156/156361279-9f21d21d-61cb-4967-ba1d-3f60b9b2d462.png)
![Screenshot (334)](https://user-images.githubusercontent.com/74594156/156361292-63d04888-6e0a-48c6-8436-24ea7ffe7a0b.png)

Only then will you be able to edit or delete the vault you have created.
That is because the server is intended to work that way in the future when more functionality is added.

Also when editing a vault only the name can be changed.
The amount is intended to be changed by donations and withdrawals/transactions but that functionality is not yet implemented. 

All the crud operations work properly and show the according alerts
![Screenshot (335)](https://user-images.githubusercontent.com/74594156/156369098-a2063f5a-5c5b-470c-ab40-016703815145.png)
![Screenshot (336)](https://user-images.githubusercontent.com/74594156/156369106-6606bf00-d126-4561-9044-ecd02e3429b7.png)
![Screenshot (337)](https://user-images.githubusercontent.com/74594156/156369109-e6fac4a0-b9b2-4a39-a291-433bb662aa10.png)
![Screenshot (338)](https://user-images.githubusercontent.com/74594156/156369113-94003802-6a5e-4ed4-b01e-36f91795a123.png)
[...]

PR #572, opened 2022-03-02
Vault dashboard for admin page

stilian-nikolaev:
Adding Vault model to the admin page

closes [#529 ](https://github.com/podkrepi-bg/frontend/issues/529)

Caution! Only coordinators can manipulate the data.
In order to test the functionality at its current state you have to first make a coordinator entity with the same id of your profile
Check the following examples:
Screenshot (329)
Screenshot (330)

Furthermore, only the coordinators of the campaign connected to the vault can edit or delete it.
So you should also create a campaign in which the coordinator ID is the same as your profile ID
Check the following example:
Screenshot (331)
And when creating a vault just use the ID of the campaign you have created like this:
Screenshot (333)
Screenshot (334)

Only then will you be able to edit or delete the vault you have created.
That is because the server is intended to work that way in the future when more functionality is added.

Also when editing a vault only the name can be changed.
The amount is intended to be changed by donations and withdrawals/transactions but that functionality is not yet implemented. 

All the crud operations work properly and show the according alerts
Screenshot (335)
Screenshot (336)
Screenshot (337)
Screenshot (338)
Screenshot (339)

I have put the mutation hooks with authentication configuration directly in the restRequests file
I have three different modals: Details, Delete and Delete all
I am temporarily using the modal store from the doucments folder, due to their similarities and reusability.
A common modal store is currently being implemented and I will migrate to it once it is done

Edit and Create form are in a single file and are dynamically chosen based on whether there is an id in the query
I have extracted the actions column from the grid in a separate file
I have also created prefetch hooks and vaults are prefetched on index page and the according vault is prefetched on the edit page


kachar (approved):
@stilian-nikolaev Nicely done

One more entity to the stack :books: 

Thanks for this upgrade of the app

PR #575, opened 2022-03-03
Small speedup of build and additional tooling

quantum-grit:
## Motivation and context
Tried to make the build a little bit faster

on slower machines the full build is veeeery slow

the case was that some contributors avoid running full build because it is slow and commit not compilable code

## Testing
full rebuild and manual testing

**New dependencies**:

- `sharp` : automated image resizing

**New dev dependencies**:

- `bundle-analyzer` : shows a heatmap with sizes of packages


kachar (approved):
Looks promising :medal_military:  Thanks for the research

PR #576, opened 2022-03-03
Added meta descrption tags for better SEO of pages

### LINKED ISSUE #565: [Bug] Static page html generation for SEO
**Describe the bug**
Currently static pages like www.podkrepi.bg/about are not indexed by Google. The issue could be because there is no static html generated for them.

**To Reproduce**
Search for podkrepi in google and notice that there is no indexed text from podkrepi.bg

quantum-grit:
Changes:
- removed hardcoded lang=en because it shows as en for bg pages
- added metaDescription for better SEO

## Motivation and context
Closes #565 - Google SEO was not showing description and not indexing well Podkrepi.bg 

## Screenshots:
Wrong language code:
Wrong language code
Fixed language code:
Fixed Language code
Added Meta Description:
MetaDescription Added
Lighthouse says SEO checks at 100%
Lighthouse OK

## Testing
Manual
Lighthouse Chrome Plugin
SEO Meta 1 click Chrome Plugin


public/locales/bg/common.json
```diff
@@ -9,7 +9,7 @@
     },
     "campaigns": {
       "index": "Кампании",
-      "create": "Предложи кампания",
+      "create": "Създай кампания",
```
kachar, line 12: Marketing and other teams requested this to be renamed to Suggest campaign to avoid awkward moments that the campaign will be created straight away without reviewing.

Anyway I'm fine with it

PR #578, opened 2022-03-04
Adding select fields for ids

PepiPetrov:
According to @kachar I replaced text fields for ID with select dropdowns.


kachar:
@PepiPetrov There are some unrelated changes regarding `.husky` again, let's clear them so there's no change in that part

PepiPetrov:
I can see it has only two files.

kachar:
@PepiPetrov Looks like the mode has been changed for the executable files from 755 to 644

image

PepiPetrov:
Strange. How can I fix it?

kachar:
You might check this topic https://stackoverflow.com/a/10516406/668245

It makes git detect the changes in the file modes, which will help you see when they are changed

As for these changed files it will be easier if I join and help you clear them up

PepiPetrov:
OK. I'll read it.

kachar:
To restore the mode I've used

```
chmod 755 .husky/pre-commit manifests/postrelease.sh manifests/release.sh
```

PepiPetrov:
Thank you for the info.

src/common/util/keycloak.ts
```diff
@@ -0,0 +1,55 @@
+import getConfig from 'next/config'
```
kachar, line 1: @PepiPetrov This file has been moved to `frontend/src/middleware/auth/keycloak.ts` on `master`

There's no point in having two copies of the same file, so please remove this sone and keep only the one from `middleware/auth/keycloak`

kachar (approved):
@PepiPetrov Marvelous upgrade of the app. Thanks for your contribution.

Can you please split your contributions in smaller PRs that are providing dedicated functionality. 

Also some overview of the changes would be nice! The PR description now explains that person pickers will be developed, but the actual code gives more than that. It will make us lost if we want one part of the code, but not the other. QA team won't be aware of what they should test after such a PR

:v:

PR #584, opened 2022-03-06
Show persons that is not already coordinators in dropdown

dplamenov:
Show persons that is not already coordinators in dropdown

## Screenshots:

| before | after |


kachar (approved):
Comparing the lists on the frontend is not efficient at all, but that's what we can do at this point I guess.

In the best-case scenario, the backend won't give an error when we try to create a coordinator profile if one already exists. It would return the existing record instead of an error.

PR #599, opened 2022-03-09
Fixed admin navigation items to be blue

PepiPetrov:
Fixed admin navigation items to be blue according to @TeDy-A-K's issue.


kachar (approved):
Nice extraction of the menu items to a list

Next time please include before/after screenshot or a video of the new functionality. Gives better understanding of what's changed and what needs to be tested

PR #601, opened 2022-03-10
Fixed text color on support us page to be blue

PepiPetrov:
Fixed text color on support us page to be blue.

Before: 

155538025-152f5906-b6ef-4d2b-ac0a-d0a7813e25d0

After:


src/components/support-form/steps/AdditionalQuestions.tsx
```diff
@@ -209,7 +212,7 @@ export default function AdditionalQuestions() {
   return (
     <Grid container spacing={6} justifyContent="center">
       <Grid item xs={12}>
-        <Typography variant="h4" align="center">
+        <Typography variant="h4" align="center" className={classes.subtitleText}>
```
kachar, line 215: This way of handling it is okay, but the better way of handling it is with a custom Title component that has those styles internally.

With dedicated components, we embrace composition and single-purpose and we guarantee same result for all instances of the title.

Now the title can be blue but if we want to adjust the font size or underlying components we need to tweak 5+ files

src/components/support-form/helpers/header-styles.ts
```diff
@@ -0,0 +1,11 @@
...
+import { lighten, Theme } from '@mui/material'
+
+export const useHeaderStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    subtitleText: {
+      color: lighten(theme.palette.primary.dark, 0.1),
+      width: '100%',
+    },
+  }),
+)
```
kachar, line 11: You can import the theme inside the body of `useHeaderStyles` in order to reduce the props that this hook needs.

Example

```suggestion
import theme from 'common/theme'

export const useHeaderStyles = makeStyles(() =>
  createStyles({
    subtitleText: {
      color: lighten(theme.palette.primary.dark, 0.1),
      width: '100%',
    },
  }),
)
```

PR #605, opened 2022-03-11
Change campaign carousel settings

ani-kalpachka:
- Added pagination:
Screenshot 2022-03-11 082509

- Hid the slider if there are no images to display.
- Hid nav arrows if there is only one image.
- Slowed down the autoplay time to 5 seconds.


kachar (approved):
Nice touch :balloon:

Issue #531, opened 2022-02-23
[Withdrawal] Create dashboard for Withdrawal entity


Yakimov1337:
Based on issue [#459 ](https://github.com/podkrepi-bg/frontend/issues/459) a layout with crud functionality for Withdrawal entity and integrate it into the admin panel.

PR #607, opened 2022-03-11
[Withdrawal] Create dashboard and CRUD functionalities 

Yakimov1337:
Create CRUD functionalities for Withdrawal entity based on https://github.com/podkrepi-bg/frontend/issues/531#issue-1148312835

Screenshot_20
Screenshot_21
Screenshot_22
Screenshot_23


src/common/hooks/withdrawals.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export const usePersonListForWithdrawal = () => {
+  const { keycloak } = useKeycloak<KeycloakInstance>()
+  const personList = useQuery<PersonResponse[]>(
+    endpoints.person.list.url,
+    authQueryFnFactory<PersonResponse[]>(keycloak?.token),
+  )
+  return personList.data
+}
```
kachar, line 66: Is there anything special about the person list for withdrawals? If it's plain person list we can reuse the person list hook

src/components/withdrawals/CreateForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+import SubmitButton from 'components/common/form/SubmitButton'
+import { WithdrawalStatus } from './WithdrawalTypes'
+
+export default function CreateForm() {
+  const router = useRouter()
+  const bankAccounts = useBankAccountsForWithdrawal()
+  const vaults = useVaultsForWithdrawal()
+  const campaigns = useCampaignsForWithdrawal()
+  const personList = usePersonListForWithdrawal()
+  const currencies = ['BGN', 'USD', 'EUR']
```
kachar, line 39: Better extract this line to module level outside of the component

src/components/withdrawals/CreateForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+  function handleSubmit() {
+    const data: WithdrawalInput = {
+      status: WithdrawalStatus.initial,
+      currency: currency,
+      amount: amount,
+      reason: reason,
+      sourceVaultId: vaultId,
+      sourceCampaignId: campaignId,
+      bankAccountId: bankAccountId,
+      documentId: '6061b425-34e6-45bf-b563-a85b933c9339',
```
kachar, line 74: If `documentId` is mandatory maybe we should make it optional

If it's not mandatory then we don't need to pass it at the stage of withdrawal creation as it comes later on in real life

src/components/withdrawals/CreateForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+      AlertStore.show('Моля изберете Валута!', 'error')
+    } else if (data.amount != undefined && data.amount < 1) {
+      AlertStore.show('Моля изберете Сума!', 'error')
+    } else if (data.documentId != undefined && data.documentId?.length < 1) {
+      AlertStore.show('Моля изберете Документ!', 'error')
+    } else if (data.approvedById != undefined && data.approvedById?.length < 1) {
+      AlertStore.show('Моля изберете полето "Одобрен от"!', 'error')
+    } else {
+      mutation.mutate(data)
+    }
```
kachar, line 99: This way of handling validation is not acceptable.

All these messages and validation rules should be part of `yup` schema with single error handler for all the errors

```
AlertStore.show(errorMessage, 'error')
```

src/components/withdrawals/CreateForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  return (
+    <Box>
+      <Typography variant="h5" component="h2" sx={{ marginBottom: 2, textAlign: 'center' }}>
+        {t('withdrawals:form-heading')}
+      </Typography>
+      <Grid container spacing={2} sx={{ width: 600, margin: '0 auto' }}>
```
kachar, line 107: We better use `GenericForm` to handle the form data and the validation

Handling it manually will lead to confusion and duplicate data.

The Formik forms can handle validation and can set values on their own without custom `useState` for each of the data fields

kachar:
@Yakimov1337 Let's make the necessary updated in a follow up PR not to hold this one too much on the pipeline

PR #609, opened 2022-03-14
Removed restRequests folder and changed file extensions in service folder

PepiPetrov:
Removed restRequests folder and changed file extensions in service folder.


kachar (approved):
Updates looks good as they provide better structure to the codebase

PR #619, opened 2022-03-16
Fixed campaign details page

PepiPetrov:
Fixed campaign details page according to figma design.


kachar (approved):
@PepiPetrov When implementing a figma design-related task it's really useful to provide before/after screenshots for historical purposes.


Thanks for your contribution! :small_airplane:

PR #621, opened 2022-03-16
Major admin city crashes because of too many re renders

PepiPetrov:
Major admin city crashes because of too many re renders. Countries crashes when trying to select a record.

### DIFF
--- a/src/components/countries/CountriesPage.tsx
+++ b/src/components/countries/CountriesPage.tsx
@@ -5,6 +5,7 @@ import AdminContainer from 'components/admin/navigation/AdminContainer'
 
 import HeadingSeparator from './HeadingSeparator'
 import CountryGrid from './grid/CountryGrid'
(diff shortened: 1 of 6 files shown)


src/components/countries/CountriesPage.tsx
```diff
@@ -5,13 +5,15 @@ import AdminContainer from 'components/admin/navigation/AdminContainer'
...
 import CountryGrid from './grid/CountryGrid'
+import GridAppbar from './grid/GridAppbar'
 
 export default function CountriesPage() {
   const { t } = useTranslation('countries')
 
   return (
     <AdminLayout>
       <AdminContainer title={t('headings.countries')}>
+        <GridAppbar></GridAppbar>
```
kachar, line 16: Self-closing syntax is prefered when the component has no children

```suggestion
        <GridAppbar />
```

Issue #613, opened 2022-03-15
[Feature] Create campaign types filter


Yakimov1337:
Create campaign types filter   for campaign page, edit heading and subheading, remove apply button.
![Screenshot_25](https://user-images.githubusercontent.com/92337248/158376072-5db7b866-3ab5-4d0e-9e7a-3ae3e47d06b1.png)
.

PR #625, opened 2022-03-17
[Feature] Create campaign types filter for campaigns page

Yakimov1337:
- [x] Created filter for campaign types for campaigns page based on #613
- [x] Changed styles and text for heading and subheading
- [x] Removed apply button
Screenshot_30

https://user-images.githubusercontent.com/92337248/158852120-b41cce2d-45e9-43e0-81fb-0574867ce3b9.mp4


src/components/campaigns/CampaignFilter.tsx
```diff
@@ -0,0 +1,255 @@
...
+            type.name != 'Treatment and rehabilitation' &&
+            type.name != 'Leukemia' &&
+            type.name != 'Nursing homes' &&
+            type.name != 'Transplantation' &&
+            type.name != 'Cancer' &&
+            type.name != 'Genetic diseases' &&
+            type.name != 'Rehabilitation'
+          ) {
+            othersCounter++
+          }
```
kachar, line 89: We should never depend on the actual name of the campaign as it would be different when translated in Bulgarian.

We need to find different way of differentiating those categories, even if we need to add an additional db field

kachar (changes requested):

kachar:
@Yakimov1337 The backend for this task is extracted in https://github.com/podkrepi-bg/api/issues/146

src/components/campaigns/CampaignFilter.tsx
```diff
@@ -0,0 +1,255 @@
...
+      campaignTypes?.map((type) => {
+        if (campaign.campaignTypeId == type.id) {
+          if (
+            type.name == 'Leukemia' ||
+            type.name == 'Transplantation' ||
+            type.name == 'Cancer' ||
+            type.name == 'Nursing homes' ||
+            type.name == 'Treatment and rehabilitation' ||
+            type.name == 'Rehabilitation' ||
+            type.name == 'Genetic diseases'
```
bikkimahato, line 66: Strict equality operator === should be used.

src/components/campaigns/CampaignFilter.tsx
```diff
@@ -0,0 +1,255 @@
...
+          }
+        })
+        setCampaignToShow(result)
+      })
+    }
+  }, [campaignType])
+
+  // async function handleClick(e: any) {
+  //   setCampaignType(e.target.textContent)
+  // }
```
bikkimahato, line 178: Uncommented should not be used.

bikkimahato (commented):
General feedback.

src/components/campaign-types/Categories.ts
```diff
@@ -0,0 +1,13 @@
+export enum CampaignTypeCategory {
```
kachar, line 1: This file better be lowercase as we don't export any React component from it, but it acts as utility function

Minor detail

kachar:
Desktop|Mobile
---|---
image|localhost_3040_campaigns(iPhone 12 Pro)

kachar (approved):

Issue #620, opened 2022-03-16
[Feature] Create Infographic for campaign page


Yakimov1337:
Create infographic for campaign page.
![Screenshot_26](https://user-images.githubusercontent.com/92337248/158590295-93eb6fd1-7dd9-404b-8331-12d402891b63.png)

PR #626, opened 2022-03-17
[Feature] Create Infographic for campaign page

Yakimov1337:
- [x] Added Infographic image for campaign page based on #620 
Screenshot_31


kachar (approved):
Nicely done @Yakimov1337 

I wonder how this would look like on mobile

Issue #632, opened 2022-03-19
[Feature] Add Responsive rotated infographic image 


Yakimov1337:
**Add Responsive rotated infographic image for Campaign page**

PR #633, opened 2022-03-19
[Feature] Add Responsive rotated infographic image 

Yakimov1337:
- [x] Add Responsive ROTATED infographic image #632
Screenshot_33


src/components/campaigns/CampaignsList.tsx
```diff
@@ -7,6 +7,10 @@ import CampaignCard from './CampaignCard'
 
 export default function CampaignsList() {
   const { data, isLoading } = useCampaignList()
+  let mobile = false
+  if (/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)) {
+    mobile = true
+  }
```
kachar, line 13: We better use the standard MUI approach to the responsive sizes and get it via the `useMediaQuery()` hook

```
import { useMediaQuery } from '@mui/material'
```

```suggestion
  const mobile = useMediaQuery('(max-width:900px)')
```

kachar (approved):

Issue #623, opened 2022-03-16
CampaignDocumentRole - admin interface


borislavstoychev:
The task is to create a Data grid which should contain the following attributes:
- [ ] **name:** ***Example (profile image)***
- [ ] **description:** ***Example (Image that is used as a profile picture for the campaign)***

1.Data grid
- [ ] Should have a title above it to say what we display in it 
- [ ]  Should display our entity loaded from the back-end
- [ ] Should have a + button/icon at the top which leads to a new page displaying the create form
- [ ] Should have a delete button/icon which opens a confirmation modal & deletes all selected items by the checkbox column 
- [ ] checkbox column
- [ ] Should have a last column with 3 icons aligned to the right :

-view icon (opens a modal that shows all the values of the entity clicked)
-edit icon (opens a new page with form in which you can edit all the form values, also on this form we should have 2 buttons -save/cancel)
-delete icon (a normal confirmation pop-up asking "Are you sure" and 2 answers yes/no)

2.Notifications
- [ ] Should show success/error notification after create/edit/delete

PR #640, opened 2022-03-21
Campaign document role #623

borislavstoychev:
## Screenshots:

## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:


borislavstoychev:
image
image
image
image
image
image
image
image
image
image
image

kachar (changes requested):
@borislavstoychev Nice overall contribution!

I've added some note that would require manual changes regarding the links and reusing the routes.

One thing I don't get at all - why the whole PR is regarding the campaign document role and not for campaign document.

We add/delete documents and not roles

Can you please correct that everywhere so the role is not part of the PR

src/service/apiEndpoints.ts
```diff
@@ -114,4 +114,13 @@ export const endpoints = {
     list: <Endpoint>{ url: '/company/list', method: 'GET' },
     viewCompany: (slug: string) => <Endpoint>{ url: '/company/' + slug, method: 'GET' },
   },
+  campaignDocument: {
+    listDocuments: <Endpoint>{ url: '/campaign-document-role', method: 'GET' },
```
kachar, line 118: Can you send me the PR where we have the API for endpoint `/campaign-document-role`

kachar: On the other hand we have backend for `/campaign-file`

borislavstoychev: I guess I should delete the role from [here]( https://github.com/podkrepi-bg/api/pull/149) too

kachar:
This PR has served its demo purpose so it can be closed now

PR #651, opened 2022-03-23
Added loading image from api for campaign card and details components

PepiPetrov:
Added loading image from api for campaign card and details components.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -46,6 +46,9 @@ const moduleExports = {
       CAMPAIGN: process.env.FEATURE_CAMPAIGN ?? false,
     },
   },
+  images: {
+    domains: [process.env.API_HOST],
+  },
   swcMinify: true,
 }
 
--- a/next.config.js
+++ b/next.config.js
@@ -47,7 +47,7 @@ const moduleExports = {
     },
   },
   images: {
-    domains: [process.env.API_HOST],
+    domains: [process.env.API_HOST || 'localhost'],
   },
   swcMinify: true,
 }
--- a/next.config.js
+++ b/next.config.js
@@ -47,7 +47,7 @@ const moduleExports = {
     },
   },
   images: {
-    domains: [process.env.API_HOST || 'localhost'],
+    domains: [process.env.IMAGE_HOST || 'localhost'],
   },
   swcMinify: true,
 }
--- a/src/components/campaigns/CampaignCard.tsx
+++ b/src/components/campaigns/CampaignCard.tsx
@@ -19,7 +19,7 @@ import {
 } from '@mui/material'
 import { Favorite } from '@mui/icons-material'
 import ArrowForwardIosIcon from '@mui/icons-material/ArrowForwardIos'
-import getConfig from 'next/config'
+import { campaignListPictureUrl } from 'common/util/campaignImageUrls'
 
 const useStyles = makeStyles((theme) => ({
   media: {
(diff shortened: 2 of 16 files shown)


src/components/campaigns/CampaignCard.tsx
```diff
@@ -118,14 +119,20 @@ type Props = { campaign: CampaignResponse }
 export default function CampaignCard({ campaign }: Props) {
   const classes = useStyles()
   const { t } = useTranslation()
+  const { publicRuntimeConfig } = getConfig()
   //const amounts = [20, 50, 100]
   //const [alignment, setAlignment] = React.useState<string | null>('left')
   // const handleAlignment = (event: React.MouseEvent<HTMLElement>, newAlignment: string | null) => {
   //   setAlignment(newAlignment)
   // }
+  const undefinedUrl = publicRuntimeConfig.API_URL + '/api/campaign-file/' + undefined
```
kachar, line 128: This row is very uncommon, can you please share why you needed such variable name and structure??

src/components/campaigns/CampaignDetails.tsx
```diff
@@ -66,8 +67,15 @@ type Props = {
...
   const classes = useStyles()
   const { t } = useTranslation()
+  const { publicRuntimeConfig } = getConfig()
 
-  const bannerSource = '/img/campaign-banner.png'
+  const undefinedUrl = publicRuntimeConfig.API_URL + '/api/campaign-file/' + undefined
+  const pictureUrl =
+    publicRuntimeConfig.API_URL +
+    '/api/campaign-file/' +
+    campaign.campaignFiles.find((x) => x.role == 'background')?.id
```
kachar, line 76: If we need to use these on a second place we need to extract them as utility function

src/components/campaigns/CampaignCard.tsx
PepiPetrov: We need it to check if there is no valid picture url.

kachar: It's going to be better to make a condition based on the result of 

```tsx
function campaignPicture(campaign): string | null {
  const background = campaign.campaignFiles.find((x) => x.role == 'background')
  if (background) {
    return background.id
  }
  return null
}
```

by encapsulating this functionality you can reuse it elsewhere and you don't have to do against checks undefined values

src/components/campaigns/CampaignCard.tsx
```diff
@@ -19,19 +19,20 @@ import {
 } from '@mui/material'
 import { Favorite } from '@mui/icons-material'
 import ArrowForwardIosIcon from '@mui/icons-material/ArrowForwardIos'
+import { campaignListPictureUrl } from 'common/util/campaignImageUrls'
 
 const useStyles = makeStyles((theme) => ({
   media: {
     backgroundSize: 'contain',
-    filter: 'grayscale(1)',
+    // filter: 'grayscale(1)',
```
kachar, line 27: If we don't need these lines please remove them instead of commenting them out

next.config.js
```diff
@@ -46,6 +46,9 @@ const moduleExports = {
       CAMPAIGN: process.env.FEATURE_CAMPAIGN ?? false,
     },
   },
+  images: {
+    domains: [process.env.API_HOST || 'localhost'],
+  },
```
kachar, line 51: Thoughtful addition :+1:

kachar (approved):
Well done @PepiPetrov :clap: 

Thanks for the contribution and the thoughtful experience

PR #656, opened 2022-03-24
Upload files with types

dplamenov:
Upload types of campaign files
## Screenshots:
Screenshot at Mar 24 17-27-16


src/components/file-upload/FileList.tsx
```diff
@@ -23,6 +44,22 @@ function FileList({ files, onDelete }: NewType) {
...
           <ListItemText primary={file.name} />
+          <FormControl>
+            <InputLabel id="choose-type-label">Избери тип</InputLabel>
+            <Select
+              labelId="choose-type-label"
+              id="choose-type"
+              value={(filesRole.find((f) => f.file === file.name)?.role as string) || 'Избери тип'}
+              label="Избери тип"
+              onChange={setFileRole(file)}>
+              <MenuItem value="Избери тип">Избери тип</MenuItem>
```
kachar, line 55: This menu item "Choose type" will count as a role too

src/components/file-upload/FileList.tsx
```diff
@@ -23,6 +44,22 @@ function FileList({ files, onDelete }: NewType) {
...
+              <MenuItem value="background">background</MenuItem>
+              <MenuItem value="coordinator">coordinator</MenuItem>
+              <MenuItem value="campaignPhoto">campaignPhoto</MenuItem>
+              <MenuItem value="invoice">invoice</MenuItem>
+              <MenuItem value="document">document</MenuItem>
```
kachar, line 60: Let's use an enum here and loop over the values/keys

Adding these manually will be considered bad practice as we might forget to do it

kachar: On master branch we have those values in a enum already

https://github.com/podkrepi-bg/frontend/blob/77540771be860821daf1cbabcc7867263cf32cd3/src/components/campaign-file/roles.ts#L1-L9

imilchev:
I leave this review up to @kachar as front-ends are not my cup of tea.

kachar (approved):

src/components/campaigns/CampaignForm.tsx
```diff
@@ -245,6 +245,10 @@ export default function CampaignForm({ initialValues = defaults }: CampaignFormP
             <FileUpload
               onUpload={(newFiles) => {
                 setFiles((prevFiles) => [...prevFiles, ...newFiles])
+                setFilesRole((filesRoles) => [
+                  ...filesRoles,
+                  ...newFiles.map((file) => ({ file: file.name, role: 'background' })),
```
kachar, line 250: We could use the enum value here instead of hardcoded string

kachar (approved):

PR #662, opened 2022-03-29
[expense] Created admin dashboard panel

### LINKED ISSUE #477: [Expense] Create admin dashboard panel

dpmarkov:
## Motivation and context

I've created the **UI and functionality** needed to operate with the _Expense_ entity in the **admin** dashboard.
Operations such as: create, edit, view and delete an _Expense_ record are now possible.

## Demo:

I think a **short demo** will be most sufficient.

https://user-images.githubusercontent.com/71282381/160602474-02253c3d-040e-42e0-bd84-0d9be1d0773b.mp4

close #477


src/components/expenses/ExpenseSelect.tsx
```diff
@@ -0,0 +1,75 @@
...
+      ? useDocumentsList().data?.map((record) => record.id)
+      : name == 'vaultId'
+      ? useVaultsList().data?.map((record) => record.id)
+      : name == 'type'
+      ? validTypes
+      : name == 'status'
+      ? validStatuses
+      : name == 'currency'
+      ? validCurrencies
+      : ['true', 'false'] //deleted
```
kachar, line 34: This part is not really well structured

Instead of having one `ExpenseSelect` component which combines `vaultId`, `type`, `status`, `currency` and etc we need to have separate components like:

- `VaultSelect`
- `ExpenseTypeSelect`
- `ExpenseStatusSelect`
- `CurrencySelect`
- `DeletedCheckbox`
- `DocumentSelect`
- `<PersonSelect name="approvedById" />`

This needs to be updated in order to accept the PR

src/gql/expenses.ts
```diff
@@ -0,0 +1,57 @@
...
+  pending = 'pending',
+  approved = 'approved',
+  canceled = 'canceled',
+}
+
+export enum ExpenseCurrency {
+  BGN = 'BGN',
+  EUR = 'EUR',
+  USD = 'USD',
+}
```
kachar, line 57: This needs to be removed as we already have the currencies in another file

src/components/expenses/grid/DetailsModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const dataConverted = [
+    { name: 'ID', value: `${data?.id}` },
+    { name: t('fields.type'), value: `${data?.type}` },
+    { name: t('fields.status'), value: `${data?.status}` },
+    { name: t('fields.currency'), value: `${data?.currency}` },
+    { name: t('fields.amount'), value: `${data?.amount}` },
+    { name: t('fields.vaultId'), value: `${data?.vaultId}` },
+    { name: t('fields.deleted'), value: `${data?.deleted}` },
+    { name: t('fields.description'), value: `${data?.description}` },
+    { name: t('fields.documentId'), value: `${data?.documentId}` },
```
kachar, line 28: Why do you convert all fields to string? This actions seems to be redundant and will lead to `undefined` strings shown to the end user

src/components/expenses/Form.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  if (id) {
+    id = String(id)
+    data = useViewExpense(id).data
+  }
+
+  const initialValues: ExpenseInput = {
+    type: data?.type || 'none',
+    status: data?.status || 'pending',
+    currency: data?.currency || 'BGN',
```
kachar, line 65: The default here should come from the predefined currency enum `Currency.BGN`

src/components/expenses/Form.tsx
```diff
@@ -0,0 +1,177 @@
...
+import GenericForm from 'components/common/form/GenericForm'
+import FormTextField from 'components/common/form/FormTextField'
+import SubmitButton from 'components/common/form/SubmitButton'
+import LinkButton from 'components/common/LinkButton'
+
+import ExpenseSelect from './ExpenseSelect'
+
+const validTypes = Object.keys(ExpenseType)
+const validStatuses = Object.keys(ExpenseStatus)
+const validCurrencies = Object.keys(ExpenseCurrency)
```
kachar, line 32: This needs to use another `Currency` enum which is already on master

kachar (changes requested):
@dpmarkov We need to do some changes in order for this PR to be accepted.

Please take care of them and ping us if you need any assistance

src/components/expenses/grid/DetailsModal.tsx
dpmarkov: It's because of how `<DetailsDialog/>` component expects data.
```javascript
export type Field = {
  name: string
  value: string | number | JSX.Element
}
```

If I modify it to (_by adding undefined and null for value type_)
```javascript
export type Field = {
  name: string
  value: string | number | JSX.Element | undefined | null
}
```
then this will work:
```javascript
const dataConverted = [
    { name: 'ID', value: `${data?.id}` },
    { name: t('fields.type'), value: data?.type },
    { name: t('fields.status'), value: data?.status },
    { name: t('fields.currency'), value: data?.currency },
    { name: t('fields.amount'), value: data?.amount },
    { name: t('fields.vaultId'), value: data?.vaultId },
    { name: t('fields.deleted'), value: data?.deleted?.toString() },
    { name: t('fields.description'), value: data?.description },
    { name: t('fields.documentId'), value: data?.documentId },
    {
      name: t('fields.approvedById'),
      value: approvedBy ? `${approvedBy.firstName} ${approvedBy.lastName} (${approvedBy.id})` : '',
    },
  ]

  return <DetailsDialog data={dataConverted} />
```

Null is required because **documentId** can be null.
Undefined, obviously because the rest of the properties of ExpenseResponse can be undefined.
Deleted property is of type Boolean | undefined so it need to be converted to string if set.

**TL;DR**:
I agree. I want to remove the redundant conversation to strings.
But to do so I need to change the possible types of value property in type Field (_DetailsDialog.tsx_) so Typescript can be happy.

src/gql/expenses.ts
dpmarkov: I didn't noticed. My bad.

src/components/expenses/grid/DetailsModal.tsx
kachar: It's best to define the `value` as `value: React.ReactNode` which includes all of the mentioned types

src/components/expenses/ExpenseSelect.tsx
dpmarkov: Is the `DeletedCheckbox` necessary on the form?

I don't receive the _soft_ deleted records because that's the way the API is set.

controller:
```javascript
@Public()
@Get('list')
async findAll() {
  return await this.expensesService.listExpenses()
}
```

service:
```javascript
async listExpenses(returnDeleted = false): Promise<Expense[]> {
  return this.prisma.expense.findMany({ where: { deleted: returnDeleted } })
}
```

I also cannot edit record which was _soft_ deleted because it just doesn't show on the grid.

dpmarkov:
@kachar Thanks for all the input and tips.
I did the required changes.

About the 'deleted' field. I changed it to checkbox and left it disabled.

form-edit

Additionally I removed the `DeleteCheckbox` from the create form. It makes no sense to be there.

form-create

Let me know if something needs improving.

src/components/expenses/ExpenseSelect.tsx
kachar: I've referenced that `DeleteCheckbox` as an example for the following piece of code in this branch

```tsx
<ExpenseSelect name="deleted" allowEmpty={false} disabled={true} />
```

kachar: Beside that we don't need the `deleted` field to be visible

src/components/expenses/Form.tsx
```diff
@@ -0,0 +1,181 @@
...
+              <DeletedCheckbox />
+            </Grid>
+          ) : null}
+
+          {id ? (
+            <>
+              <Grid item xs={6}>
+                <Button fullWidth variant="contained" type="submit" color="secondary">
+                  {t('btns.save')}
+                </Button>
```
kachar, line 160: Please use  `<SubmitButton fullWidth label="btns.save" />` instead of custom button

kachar: If the only difference is the label you can simplify this code to:

```tsx
<Grid item xs={6}>
  <SubmitButton fullWidth label={id ? 'btns.save' : 'expenses:btns.submit'} />
</Grid>
```

src/components/expenses/CurrencySelect.tsx
```diff
@@ -0,0 +1,39 @@
+import { useTranslation } from 'react-i18next'
+import { FormControl, MenuItem } from '@mui/material'
+import { useField } from 'formik'
+
+import { Currency } from 'gql/currency'
+import FormTextField from 'components/common/form/FormTextField'
+
+export default function CurrencySelect() {
+  const name = 'currency'
```
kachar, line 9:
```suggestion
export default function CurrencySelect({ name = 'currency' }) {
```

If we move the name to the passed props with a default value we make this component really reusable

If we need it in a different form we can use it twice by passing different form `name` prop:

```tsx
<>
  <CurrencySelect name="fromCurrency" />
  <CurrencySelect name="toCurrency" />
</>
```

src/components/expenses/Form.tsx
```diff
@@ -0,0 +1,181 @@
...
+            />
+          </Grid>
+          <Grid item xs={id ? 10 : 12}>
+            <PersonSelect name="approvedById" />
+          </Grid>
+          {id ? (
+            <Grid item xs={2}>
+              <DeletedCheckbox />
+            </Grid>
+          ) : null}
```
kachar, line 153: You can simplify to

```suggestion
          {id && (
            <Grid item xs={2}>
              <DeletedCheckbox />
            </Grid>
          )}
```

src/components/expenses/PersonSelect.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useField } from 'formik'
+
+import { usePersonList } from 'common/hooks/person'
+import FormTextField from 'components/common/form/FormTextField'
+
+type Props = {
+  name: string
+}
+
+export default function PersonSelect({ name }: Props) {
```
kachar, line 12: I think this might work as well

In any case the `name` prop can be optional when it has a default value

```suggestion
export default function PersonSelect({ name = 'personId' }) {
```

src/components/expenses/CurrencySelect.tsx
dpmarkov: Agree, and probably in the future should be moved to other place like `components/admin`. It'll be more _visible_ and we'll know that it's a component made to be shared across the admin forms.

I'll add the translation string to the passed props as well ```javascript
CurrencySelect({ name = 'currency', translation = 'expenses' })```. Just in case we decide to use different json file.

src/components/expenses/ExpenseSelect.tsx
dpmarkov: I see.

src/components/expenses/CurrencySelect.tsx
kachar: Let's rename the translation prop from `translation` to `label` or something more explicit so it's clear what's the purpose.

It will be hard if we need to pass two `translation` props for different things

kachar: @dpmarkov I do agree we can move this component to `components/currency` even now, no nee d to wait to a future moment

kachar: Actually I was already confused what the `translation` actually means - it's the translation domain :dagger: not the actual label.

Please change it to `translationDomain` or `ttranslationNamespace`

src/service/apiEndpoints.ts
```diff
@@ -82,6 +82,14 @@ export const endpoints = {
     postCoordinator: <Endpoint>{ url: '/coordinator', method: 'POST' },
     deleteCoordinator: (id: string) => <Endpoint>{ url: `/coordinator/${id}`, method: 'DELETE' },
   },
+  expenses: {
+    listExpenses: <Endpoint>{ url: '/expenses/list', method: 'GET' },
+    createExpense: <Endpoint>{ url: '/expenses/create-expense', method: 'POST' },
+    viewExpense: (id: string) => <Endpoint>{ url: `/expenses/${id}`, method: 'GET' },
+    editExpense: (id: string) => <Endpoint>{ url: `/expenses/${id}`, method: 'PATCH' },
+    deleteExpense: (id: string) => <Endpoint>{ url: `/expenses/${id}`, method: 'DELETE' },
+    deleteExpenses: <Endpoint>{ url: `/expenses`, method: 'DELETE' },
```
kachar, line 91: Maybe for the delete endpoint w need to pass `id` as well? @dpmarkov

kachar: Or this whole row `delteExpenses` is not needed

src/service/expense.ts
```diff
@@ -0,0 +1,53 @@
...
+  return async () => {
+    return await apiClient.delete<ExpenseResponse, AxiosResponse<ExpenseResponse>>(
+      endpoints.expenses.deleteExpenses.url,
+      {
+        headers: authConfig(keycloak?.token).headers,
+        data: idsToDelete,
+      },
+    )
+  }
+}
```
kachar, line 53:
```suggestion
```

src/service/apiEndpoints.ts
kachar, line 91:
```suggestion
```

src/components/expenses/CurrencySelect.tsx
dpmarkov: In that case should I remove `const { t } = useTranslation(translation)` and rely on just the label prop for that matter?

kachar: I think we can keep it as the field label is getting translated automatically as `fields.${name}`

Normally for most of the fields we can have translations in `common.json` with key `fields`

No need to translate `currency` in every entity, as it would result in multiple translations for the same thing.

However if we pass the translation namespace/domain to the component we can override it which is a good thing for reusing it.

Does that make sense?

dpmarkov: @kachar maybe adding currency in the `common` translation domain might be a good idea?
If you agree it could be in common -> admin -> fields -> currency..

dpmarkov: I'll put 'fields' in the base of the json object not nested in admin. I didn't noticed at first that 'admin' is in 'nav'.

dpmarkov:
I made a quick check that everything still works and translations are loaded correctly. It looks ok to me.
All done.

src/components/expenses/grid/GridHelper.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { GridRenderCellParams } from '@mui/x-data-grid'
+
+export const statusRenderCell = (cellValues: GridRenderCellParams): React.ReactNode => {
+  const style =
+    cellValues.value === 'pending'
+      ? { color: 'blue' }
+      : cellValues.value === 'approved'
+      ? { color: 'green' }
+      : cellValues.value === 'canceled'
```
kachar, line 10: Let's use enum values instead of string constants

kachar (approved):
Thanks, @dpmarkov for this tremendous progress for the app, a lot of effort has been set in and the results are outstanding.

dpmarkov:
> Thanks, @dpmarkov for this tremendous progress for the app, a lot of effort has been set in and the results are outstanding.

Thank you too. Your suggestions were very helpful along the way of improving this PR.

Issue #638, opened 2022-03-21
Extract common reusable component for `AppBar` functionality in admin


kachar:
Extract common reusable component for `AppBar` functionality in admin

Pass the needed title/actions/url params to that component

kachar:
## Screenshots:

| Before              | After              |
| ------------------- | ------------------ |
|![image](https://user-images.githubusercontent.com/893608/162612752-f0b833b9-0fe1-4302-bc09-3a1a8d772333.png)|![image](https://user-images.githubusercontent.com/893608/162612766-ca0d25ed-04eb-492e-bd5c-4df717688cef.png)|

PR #682, opened 2022-04-08
 Extract common reusable component for AppBar functionality in admin #638 

borislavstoychev:
## Motivation and context

Extracts common components and update style of sub menus

## Screenshots:

|image|image|


kachar (approved):
Nice update of the app @borislavstoychev 
Thanks for your invested time :clock1:

PR #688, opened 2022-04-09
Redirect user to same page before login page

### LINKED ISSUE #683: [Bug] User doesn't go back to the requested page after login
**Describe the bug**
When user visits https://dev.podkrepi.bg/admin without being logged they are sent to login, but then they are never sent back to `/admin`

**To Reproduce**

1. Go to https://dev.podkrepi.bg/admin or https://dev.podkrepi.bg/admin/supporters
2. End up on login page
3. Login as valid user
4. See that you always land on https://dev.podkrepi.bg/profile

**Expected behavior**
Land on the same page you initially requested

https://dev.podkrepi.bg/admin
or
https://dev.podkrepi.bg/admin/supporters

dplamenov:
Redirect user to same page before login page

## Motivation and context
Should close #683 

https://user-images.githubusercontent.com/30602242/162568877-f91a1fd7-9b1d-4baf-9cb1-f04e49983301.mp4

### DIFF
--- a/src/components/auth/login/LoginPage.tsx
+++ b/src/components/auth/login/LoginPage.tsx
@@ -3,7 +3,7 @@ import { KeycloakInstance } from 'keycloak-js'
 import { LinearProgress } from '@mui/material'
 import { useKeycloak } from '@react-keycloak/ssr'
 
-import { baseUrl } from 'common/routes'
+import { baseUrl, routes } from 'common/routes'
 import { useRouter } from 'next/router'
 
 export default function LoginPage() {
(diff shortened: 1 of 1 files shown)


kachar (changes requested):
@dplamenov Thanks for the rapid development of this task

We need to do some adjustments to the code because right now as I tested it with the standard login button in the nav bar we end up in bad URL

image


The `pathname` parameter is not always available in the query so it concatinates `undefined` to the response URL.

Please make sure we have a default `redirectUrl` if this param is not set. It's gonna be nice to redirect back to `routes.profile` in that scenario

dplamenov:
Ok. I will check.

dplamenov:
@kachar I think now is much better.

kachar (approved):
Nice update @dplamenov

PR #690, opened 2022-04-09
Add nvmrc file

LyuboslavLyubenov:
NVM is a utility that allows users to switch between nodejs version "on the fly".  If user executes `nvm use` in the projects folder and there is .nvmrc file, then NVM will automatically switch to the project nodejs version. 

This pr adds that file, with latest stable version of nodejs (until this moment). For sake of consistency i used the same version as podkrepi-bg/backend.


kachar (approved):
Awesome contribution @LyuboslavLyubenov 
Thanks for the upgrade

Issue #698, opened 2022-04-10
[Feature] Make sure the edit password link leads to Keycloak profile page


kachar:
**Is your feature request related to a problem? Please describe.**
Related to #608 

At the moment edit button on the password does nothing

![image](https://user-images.githubusercontent.com/893608/162607832-82756791-6111-409e-a134-d74ab6ab27d8.png)

**Describe the solution you'd like**
Please make sure this button leads the user to Keycloak profile page so they can edit their password

The final link is:

https://keycloak.podkrepi.bg/auth/realms/webapp-dev/account/password

Where `webapp-dev` is the realm `keycloakConfig.realm`

To create that link please use the default `keycloak-js` methods if it can lead to the password page directly.

```tsx
keycloak?.createAccountUrl({
  redirectUri: `${baseUrl}${routes.profile}`,
})
```
If you can't make it work then feel free to use the hardcoded links (according to the config)

```tsx
const link = `${keycloakConfig.url}/realms/${keycloakConfig.realm}/${keycloakConfig.clientId}/password`
```

dplamenov:
@kachar When user click `edit password` button should be redirected to `${keycloakConfig.url}/realms/${keycloakConfig.realm}/${keycloakConfig.clientId}/password`, right?

![Screenshot at Apr 10 11-50-21](https://user-images.githubusercontent.com/30602242/162610428-e25663ea-5eac-4cfc-8a3e-4146a136d244.png)

PR #702, opened 2022-04-11
Redirect user to keycloak to change password

dplamenov:
Redirect user to keycloak to change password from user profile

## Motivation and context

Should close issue #698


src/components/auth/profile/PersonalInfoTab.tsx
```diff
@@ -109,7 +120,7 @@ function PersonalInfoTab(props: { value: number; index: number }) {
               <p className={classes.bold}>парола:</p>
               <p>***********</p>
               <Box sx={{ position: 'absolute', right: '5px', top: '5px' }}>
-                <Link href="#">
+                <Link href="#" onClick={changePasswordHandler}>
```
kachar, line 123: Since `next/link` component is designed to do in-app routing here we use it not the way it's supposed to

I'd suggest that we use another component `<ExternalLink href={link}>` which will open the password page in a new tab.

PR #704, opened 2022-04-12
Admin: reduced unnecessary API calls

### LINKED ISSUE #624: [Bug] Major - App makes wrong requests
**Describe the bug**
Context: admin panel
Wrong GET requests are being sent to the api. This happens when `<DetailsModal />` component (_it doesn't matter for which entity_) is being rendered. It makes requests with either empty string or wrong id and fails to re-fetch data for the modal. This should be avoided.

This behavior causes the app to be not responsive.
For example: if you delete a record from the grid then sometimes the deleted row does not disappear instantly.

**To Reproduce**

1. Go to [admin](http://localhost:3040/admin)
2. Select any entity
3. Delete a record
4. Watch the console for errors

dpmarkov:
This PR fixes problems described in issue #624

The unnecessary API calls are **reduced** substantially.

I solved the problem by creating a **new instance** of `ModalStore` for every entity.
Also applied **conditional rendering** to `DetailsModal` so it doesn't make requests until is shown.
Considering that issue #653 is not completed I've chosen to stick with `ModalStore` for now.

Just a note that **sometimes the grid doesn't refresh** on time after a record is deleted.
I checked and it's because of expired token. But It does after a few seconds with a renewed token.
At the moment I don't have a solution for that.

Additionally I **removed some endpoints and hooks** for `delete many` since we don't use them anymore.

## Demo

https://user-images.githubusercontent.com/71282381/162940541-6d742786-977c-48ca-839a-ac025e8f739e.mp4

please close #624


src/components/admin/DeleteModal.tsx
```diff
@@ -33,5 +33,5 @@ export default observer(function DeleteModal({ mutationFn, translate, endpoint }
     deleteMutation.mutate(selectedRecord.id)
   }
 
-  return <DeleteDialog deleteHandler={deleteHandler} />
+  return <DeleteDialog modalStore={ModalStore} deleteHandler={deleteHandler} />
```
kachar, line 36: Only in this location we use the exported const `ModalStore` from `'stores/dashboard/ModalStore'`

Do you think we need to migrate it so all components that use `<DeleteModal />`  pass their `modalStore` instance?

src/components/admin/DeleteDialog.tsx
```diff
@@ -3,14 +3,15 @@ import { observer } from 'mobx-react'
...
+import { ModalStoreImpl } from 'stores/dashboard/ModalStore'
 
 type Props = {
+  modalStore: ModalStoreImpl
   deleteHandler: () => void
 }
 
-export default observer(function DeleteModal({ deleteHandler }: Props) {
-  const { isDeleteOpen, hideDelete, selectedRecord } = ModalStore
+export default observer(function DeleteModal({ modalStore, deleteHandler }: Props) {
```
kachar, line 13: Let's rename this component to match the filename - `DeleteDialog`

src/components/admin/DeleteModal.tsx
dpmarkov: Of course, I do.
I asked a question in our Discord group about that component. I don't see where it is used so I know where to define that instance.

kachar: @dpmarkov I think you need to pass the modal store to `<DeleteModal />` the same way we pass it to `<DeleteDialog />` 

In this way whoever uses delete modal will need to provide their local instance of modal store.

PR #708, opened 2022-04-14
Recurring donation

Marina-yoya:
All Recurring Donations
image

Create form
image

Details
image


src/components/recurring-donation/Form.tsx
```diff
@@ -0,0 +1,161 @@
...
+    initialValues = {
+      status: data?.status,
+      personId: data?.personId,
+      extSubscriptionId: data?.extSubscriptionId,
+      extCustomerId: data?.extCustomerId,
+      amount: data?.amount,
+      currency: data?.currency,
+      sourceVault: data?.sourceVault,
+    }
+  }
```
kachar, line 81: These values should be pass from the page

The idea is:

1. In page load the initial recoerds
2. Load the form with the initial records predefined
3. Work on the data

At the moment `CreatePage` and `EditPage` look the same, but they should not.

See as example 

https://github.com/podkrepi-bg/frontend/blob/621cb6bb8d1ab4f658effb08ee9e300b8dfc35e8/src/components/campaigns/grid/EditPage.tsx

dimitur2204:
What about having the fields for the ids be a `<Select />` like that.

recurring-donation-select-concept

In my opinion the current one with text fields is very user unfriendly.
**However the requests to get all the users and vaults would be expensive !!!**

kachar:
@dimitur2204 I do agree that we need a better approach for selecting a person.

Let's use the current `<PersonSelect />` for this task and fork the topic in another issue so we can do it properly on all locations at once.

It can be done within the `PersonSelect` component without involving the recurring donations context. It can be realized as a Button that opens a Modal for finding the person you need by typing their name, etc

kachar (approved):
Thanks, @Marina-yoya and @dimitur2204  for bringing this functionality to life

Issue #442, opened 2022-01-16
[Campaign] Implement Donation History for Campaign


igoychev:
On the Campaign Details Page create a history of donated amounts. 

See the Figma Design here: https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=6874%3A22174

Donations should be loaded via the backend API as per this issue https://github.com/podkrepi-bg/api/issues/67

dimitur2204:
Isn't that done or am I not understanding the issue?

Look here
<img width="1230" alt="image" src="https://user-images.githubusercontent.com/61479393/163349393-b6235d3e-c191-413e-adcc-5e145cf209ea.png">

dimitur2204:
Should we limit the amount of  donations that we show. Maybe the latest 5-10, because if there are a lot the UI wouldn't look good.

PR #709, opened 2022-04-14
Add getting campaign donations from backend

dimitur2204:
## Motivation and context

This is needed to provide information about the history of the donations on a given campaign.
Connected to #442.
Currently displaying only one donation by default and you need to click the button to see them all.

**It might be a good idea to somehow restrict the number of donations the backend gives us back. Because it theory it could be thousands and it that case that fetch would be huge.**

## Screenshots:

|image|donation-history-new|

## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:


src/common/util/date.ts
```diff
@@ -24,6 +24,36 @@ export const getRelativeDate = (value: Date | string) => {
...
+  }
+  if (duration.hours != 0) {
+    formattedString = `${duration.hours} hours ago`
+  }
+  if (duration.days != 0) {
+    formattedString = `${duration.days} days ago`
+  }
+
+  return formattedString
+}
```
kachar, line 55: @dimitur2204 In the next iterations you might consider using `formatRelative` fn from `date-fns`

https://date-fns.org/v2.28.0/docs/formatRelative

It can provide relative time in many locales

```js
formatRelative(addDays(new Date(), -6), new Date(), { locale: 'bg' })
```

src/components/campaigns/DonorsAndDonations.tsx
```diff
@@ -56,28 +38,55 @@ const useStyles = makeStyles((theme: Theme) =>
...
-export default function DonorsAndDonations() {
+export default function DonorsAndDonations({
+  donations,
+}: {
+  donations: CampaignDonation[] | undefined
+}) {
   const classes = useStyles()
   const { t } = useTranslation()
-
+  const [donationsToShow, setDonationsToShow] = useState<CampaignDonation[] | undefined>(undefined)
```
kachar, line 48: You don't have to explicitly specify `undefined`, not passing anything will effectively be the same

src/components/campaigns/InlineDonation.tsx
```diff
@@ -62,6 +63,7 @@ export default function InlineDonation({ campaign }: Props) {
   const summary = campaign.summary.find(() => true)
   const reached = summary ? summary.reachedAmount : 0
   const { data: prices } = useSinglePriceList()
+  const { data: donations, error: donationHistoryError } = useCampaignDonationHistory(campaign.id)
```
kachar, line 66: We might add  a spinner when history is loading

dimitur2204: Done in https://github.com/podkrepi-bg/frontend/pull/709/commits/f9d6300e738619a13847789b1f5b473401603fb6

src/common/util/date.ts
dimitur2204: Done in https://github.com/podkrepi-bg/frontend/pull/709/commits/f9d6300e738619a13847789b1f5b473401603fb6

kachar (approved):
Well done @dimitur2204 

Thanks for the updates!

PR #714, opened 2022-04-19
user to be redirected to a new page with info about FAQ on mobile device

Marina-yoya:
image


kachar (approved):
Nice catch @Marina-yoya

Issue #696, opened 2022-04-10
[Feature] Make sure the default birthday date is at least 18 years prior the current year


kachar:
**Is your feature request related to a problem? Please describe.**
We shall only allow people over 18 years old to be using the website

**Describe the solution you'd like**
For their convenience please make the default selection of birthday year be the current one minus 18 years.

![image](https://user-images.githubusercontent.com/893608/162607021-aec64c09-ded9-479d-9f9d-4fd62ae4e4a6.png)

PR #718, opened 2022-04-19
Georgy genchev 696 feature make sure the default birthday date is at least years prior the current year

GeorgyGenchev:
The task is to create functionality that ensures default birthday date is at least  18 years prior the current year

## Motivation and context

To make sure that the person who is registering is 18 years old

https://github.com/podkrepi-bg/frontend/issues/696

## Screenshots:

image
image
image


.prettierrc
```diff
@@ -5,5 +5,6 @@
   "trailingComma": "all",
   "printWidth": 100,
   "jsxBracketSameLine": true,
-  "arrowParens": "always"
+  "arrowParens": "always",
+  "endOfLine": "auto"
```
kachar, line 9: This should be `lf` if we want to specify it

kachar (approved):
:clap: :calendar: :cake:

PR #726, opened 2022-04-23
Add form select field common component

dimitur2204:
## Motivation and context

We didn't have a Select component to wrap around the `Formik` functionality which I think can be useful in forms around the app.

## Screenshots:

image


kachar (approved):
Nice abstraction @dimitur2204 

I hope we can take leverage of it in many components

Issue #697, opened 2022-04-10
[Feature] Create confirmation modal according to discussion


kachar:
**Is your feature request related to a problem? Please describe.**
We've discussed a specific approach for handling confirmations for deletions and now it's time to actually implement it

**Describe the solution you'd like**
Please develop the discussed functionality from https://github.com/podkrepi-bg/frontend/discussions/567#discussioncomment-2408626

ani-kalpachka:
Is this functionality already implemented?

PR #728, opened 2022-04-26
Feature/725 admin recurring donation person select

dimitur2204:
## Motivation and context

Improve #708 

Made the selection of a user for the recurring donation more user friendly and is now not a text field.

Added a `useConfirm` hook which according to the discussion here #697.

Added a `<FormFieldButton />` component which mimics the look of a material field, but acts as a button.

Added a `<PersonSelectDialog />` component which renders a dialog with an autocomplete where you can choose a person and get his information shown so that you are sure what you are selecting.

## Notes for the reviewer:

The API call for the `personList` would be a very heavy one and I don't think it's worth it just to populate an `Autocomplete`. It would be a better idea to only fetch like a list of the  `{id, firstName, lastName}` or paginate that call in some way. 

However this is an admin platform and maybe it's not that big of a deal. I am not sure :|

## Screenshots:

|choosing-a-person-old|choosing-a-person-new|


src/common/hooks/confirm.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+export type ConfirmHookProps = {
+  open: boolean
+  loading: boolean
+  openHandler: () => void
+  confirmHandler: () => void
+  closeHandler: () => void
+}
+
+const useConfirm = ({ onConfirm, onClose }: ConfirmProps): ConfirmHookProps => {
```
kachar, line 15: Awesome!

 Let's rename the filename to match the hook name - `useConfirm.ts`

src/components/person/PersonAutocomplete.tsx
```diff
@@ -0,0 +1,58 @@
...
+  onSelect: (person: PersonResponse | null) => void
+  autocompleteProps?: Omit<
+    AutocompleteProps<PersonResponse, undefined, undefined, undefined>,
+    'renderInput' | 'options' | 'getOptionLabel' | 'onChange' | 'loading'
+  >
+  showId: boolean
+}
+export default function PersonAutocomplete({
+  onSelect,
+  showId,
```
kachar, line 16: Maybe we can add a default value to the boolean field `showId` so we don't have to add it as `showId={false}`

```suggestion
  showId?: boolean
}
export default function PersonAutocomplete({
  onSelect,
  showId = false,
```

src/components/person/PersonInfo.tsx
```diff
@@ -0,0 +1,69 @@
...
+          </Typography>
+        </Box>
+      </Grid>
+      <Grid item xs={12} md={6}>
+        <Typography className={classes.infoHeading} variant="h6" color={theme.palette.primary.dark}>
+          {t('person:info.general')}
+        </Typography>
+        <Box className={classes.infoWrapper}>
+          <Typography>
+            {t('person:info.createdAt')}: {format(parseISO(person.createdAt), 'PPpp')}
```
kachar, line 55: There are couple of pre-existing formatting functions at https://github.com/podkrepi-bg/frontend/blob/master/src/common/util/date.ts

Let's reuse one of them if matches the expected format or add this format to the utils so it can be reused later on.

kachar (commented):
Thank you for the contribution @dimitur2204  :raised_hands: 

It really rocks the boat, full of reusable code and thoughtful dev experience. :boat: 

Added some minor points of discussion

src/components/person/PersonInfo.tsx
dimitur2204: Done in [95fedb9](https://github.com/podkrepi-bg/frontend/pull/728/commits/95fedb9d7a1c818ab2b7b06e662d78ad23679d4e)

src/components/person/PersonAutocomplete.tsx
dimitur2204: Done in [95fedb9](https://github.com/podkrepi-bg/frontend/pull/728/commits/95fedb9d7a1c818ab2b7b06e662d78ad23679d4e)

src/common/hooks/confirm.ts
dimitur2204: Done in [95fedb9](https://github.com/podkrepi-bg/frontend/pull/728/commits/95fedb9d7a1c818ab2b7b06e662d78ad23679d4e)

kachar:
:tada:

PR #734, opened 2022-04-28
720 feature add button to load more campaigns and set intially shown campaigns to 6.

GeorgyGenchev:
When campaigns are loaded initially only 6 campaigns should be shown.
A button should be added and when it is clicked it should show all of the campaigns. When it is clicked again it should show less campaings(6 campaigns).
When they are less than 6 campaigns there is no button.

image

image
image

## Testing

http://localhost:3040/campaigns


kachar (approved):
Nice one @GeorgyGenchev 

Thanks :eagle:

PR #735, opened 2022-04-28
727 bug campaign details page donors count

Bunny303:


src/components/campaigns/DonorsAndDonations.tsx
```diff
@@ -47,8 +47,11 @@ export default function DonorsAndDonations({
   const { t, i18n } = useTranslation()
   const [donationsToShow, setDonationsToShow] = useState<CampaignDonation[] | undefined>()
   const [all, setAll] = useState<boolean>(false)
+  const shownDonationsNumber = 1
   useEffect(() => {
-    all ? setDonationsToShow(donations) : setDonationsToShow(donations?.slice(0, 1))
+    all
```
kachar, line 52: It's gonna be better if we use `useMemo` instead of `useState + useEffect`

```tsx
const shownDonationsNumber = 5
const donationsToShow = useMemo(() => {
  if (all) {
    return donations
  }
  return donations?.slice(0, shownDonationsNumber)
}, [donations, all])
```

kachar (changes requested):
@Bunny303 Thanks for the enhancement, please check my suggestion below

PR #736, opened 2022-04-28
One time donation

borislavstoychev:
## Screenshots:

## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:

## One Time Donation
- [x] UI - all 4 steps done
- [x] First Step done
- [ ]  Second Step
     :x:  Login functionality
     :x: Google login button
     :x: New registration link
     :heavy_check_mark:  Anonymous fields 
- [ ] Third step
     :x:  Card payment 
     :heavy_check_mark: Bank payment
- [ ] Final step
     :x: On success step buttons and social icons not working
     :x: On fail step buttons not working
- [ ] Stepper information is not stored anywhere


src/components/one-time-donation/components/Steps.tsx
```diff
@@ -0,0 +1,118 @@
...
+  anonymousDonation: true,
+  personsFirstName: '',
+  personsLastName: '',
+  personsEmail: '',
+  personsPhone: '',
+  payment: 'bank',
+}
+
+export default function DonationStepper() {
+  const [status, setStatus] = React.useState(false)
```
kachar, line 39: What does `status: true` mean in this context?

Maybe we can bring more context if we rename it to `donationSuccess` or just `success` as it implies that it has only true or false as values.

Status fields normally have specific values like `pending`, `valid` and so on

src/components/one-time-donation/components/Steps.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const { t } = useTranslation('one-time-donation')
+  const router = useRouter()
+  const slug = String(router.query.slug)
+  const { data }: UseQueryResult<{ campaign: CampaignResponse }> = useViewCampaign(slug as string)
+  const valts = useVaultsList().data
```
kachar, line 44: The regular users does not have access to this endpoint, it's used only by administrators

You can find the campaign valult only by via the view campaign endpoint

If the data is not provided, the API should be changed

src/components/one-time-donation/OneTimeDonationPage.tsx
```diff
@@ -0,0 +1,72 @@
+import { Box, Grid, Theme, Typography } from '@mui/material'
+import { makeStyles } from '@mui/styles'
+import createStyles from '@mui/styles/createStyles'
+import Layout from 'components/layout/Layout'
+import { useTranslation } from 'next-i18next'
+import Image from 'next/image'
+import {} from 'path'
```
dimitur2204, line 7: Missclicked import

src/components/one-time-donation/components/AnonimusMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const [field] = useField('anonymousDonation')
+  const { t } = useTranslation('one-time-donation')
+
+  return (
+    <>
+      <CheckboxField label={t('anonimus-menu.checkbox-label') as string} name="anonymousDonation" />
+      <Collapse in={field.value} timeout="auto" unmountOnExit>
+        <List component="div" disablePadding>
+          <Typography>{t('anonimus-menu.info-start')}</Typography>
+          <Grid my={'35px'}>
```
dimitur2204, line 32: For all margins you should use `theme.spacing(amount)`

src/components/one-time-donation/components/ThirdStep.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const classes = useStyles()
+  const [field] = useField('payment')
+  const router = useRouter()
+  const slug = String(router.query.slug)
+  const { data }: UseQueryResult<{ campaign: CampaignResponse }> = useViewCampaign(slug as string)
+  const { t } = useTranslation('one-time-donation')
+  const bankAccountInfo = {
+    owner: t('third-step.owner'),
+    bank: t('third-step.bank'),
+    iban: 'BG66 UNCR 7000 1524 3490 32',
```
kachar, line 65: Please reuse the constant `ibanNumber` from https://github.com/podkrepi-bg/frontend/blob/1ba1b816db44ee1ebe7eb0fa39d5c850c6879496/src/common/iban.ts#L1

src/components/one-time-donation/components/ThirdStep.tsx
```diff
@@ -0,0 +1,137 @@
...
+              <CopyTextButton
+                label={t('third-step.btn-copy')}
+                text={bankAccountInfo.bank}
+                variant="contained"
+                size="small"
+                color="info"
+              />
+            </Grid>
+            <Grid mx={11} my={3} item display="flex" justifyContent="space-between" xs={9}>
+              <Typography> IBAN: BG66 UNCR 7000 1524 3490 32 </Typography>
```
kachar, line 108: Reuse `ibanNumber` from https://github.com/podkrepi-bg/frontend/blob/1ba1b816db44ee1ebe7eb0fa39d5c850c6879496/src/common/iban.ts#L1

src/components/one-time-donation/components/AnonimusMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+      maxWidth: '540px',
+      height: '70px',
+      background: '#FFFFFF',
+      borderRadius: '60px',
+      textAlign: 'center',
+    },
+  }),
+)
+
+export default function AnonimusMenu() {
```
kachar, line 21: This component and filename have typo, please rename them to `AnonymousMenu`

src/components/one-time-donation/components/Steps.tsx
```diff
@@ -0,0 +1,118 @@
...
+import SecondStep from './SecondStep'
+import ThirdStep from './ThirdStep'
+import Success from './Success'
+import Unsuccess from './Unsuccess'
+import {
+  DonationBankInput,
+  DonationResponse,
+  DonationStep as StepType,
+  OneTimeDonation,
+} from '../../../gql/donations'
```
kachar, line 12: Please use import from `'gql/donation'` and skip all `../../../`

src/service/donation.ts
```diff
@@ -30,6 +31,16 @@ export function useCreateDonation() {
...
+export function useCreateBankDonation() {
+  // const { keycloak } = useKeycloak<KeycloakInstance>()
+  return async (data: DonationBankInput) => {
+    return await apiClient.post<DonationResponse, AxiosResponse<DonationResponse>>(
+      endpoints.donation.createBankDonation.url,
+      data,
+    )
+  }
+}
+
```
kachar, line 43: This endpoint is only for administrators and regular users will receive `401 Forbidden` when they try to donate

src/components/one-time-donation/components/Steps.tsx
```diff
@@ -0,0 +1,118 @@
...
+        personsFirstName: values.personsFirstName,
+        personsLastName: values.personsLastName,
+        personsPhone: values.personsPhone,
+        extCustomerId: String(Math.random() * 5),
+        extPaymentIntentId: String(Math.random() * 5),
+        extPaymentMethodId: String(Math.random() * 5),
+        targetVaultId: vault?.id,
+      }
+      console.log(data)
+      // await mutation.mutateAsync(data)
```
kachar, line 76: This endpoint is only for administrators and regular users will receive `401 Forbidden` when they try to donate

A new endpoint needs to be created which only receives a subset of the params like `campaignId`, without knowledge for the `vaultId`


---


We may leave things here the way they are until we discuss and acknowledge the changes in the backend first.

public/locales/bg/one-time-donation.json
```diff
@@ -0,0 +1,59 @@
+{
+  "anonimus-menu": {
```
kachar, line 2: There's a typo here

```suggestion
  "anonymous-menu": {
```

kachar (changes requested):

src/components/one-time-donation/components/AnonimusMenu.tsx
kachar: An easy way to use the theme in `sx` prop is as follows:


```tsx
<Grid sx={theme => ({ my: theme.spacing(5) })}>
```

src/components/one-time-donation/components/Success.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { t } = useTranslation('one-time-donation')
+
+  return (
+    <Grid>
+      <Grid
+        container
+        justifyContent="center"
+        sx={{ '& .MuiSvgIcon-root': { fontSize: 132.5, color: '#4BD12A', marginTop: '58px' } }}>
+        <CheckCircleOutlinedIcon />
+      </Grid>
```
dimitur2204, line 44: You shouldn't be styling the children through a class on the parent. If you need these styles just passed them to the ` <CheckCircleOutlinedIcon />`

src/components/one-time-donation/components/Success.tsx
```diff
@@ -0,0 +1,76 @@
...
+          sx={{
+            '& .MuiSvgIcon-root': {
+              fontSize: '38px',
+              color: '#909090',
+              margin: 2,
+            },
+          }}>
+          <FacebookOutlinedIcon />
+          <InstagramIcon />
+          <LinkedInIcon />
```
dimitur2204, line 66: If you need that style for many of them like here make it a class in the styles

src/components/one-time-donation/components/ThirdStep.tsx
```diff
@@ -0,0 +1,137 @@
...
+      marginRight: 'auto',
+    },
+  }),
+)
+
+export default function ThirdStep() {
+  const classes = useStyles()
+  const [field] = useField('payment')
+  const router = useRouter()
+  const slug = String(router.query.slug)
```
dimitur2204, line 59: Shouldn't router.query.slug be a string already?

dimitur2204 (changes requested):

src/components/one-time-donation/components/FirstStep.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }),
+)
+const amounts = [
+  { values: '2', label: '2 лв.' },
+  { values: '20', label: '20 лв.' },
+  { values: '5', label: '5 лв.' },
+  { values: '50', label: '50 лв.' },
+  { values: '10', label: '10 лв.' },
+  { values: '100', label: '100 лв.' },
+]
```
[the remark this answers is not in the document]
kachar: We already have them defined with api call `useSignlePriceList()` in

https://github.com/podkrepi-bg/frontend/blob/09238e8ea0d5151610f6356c16b0b2763dc86e1e/src/components/campaigns/InlineDonation.tsx#L75

src/components/one-time-donation/components/FormCheckField.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export default function CheckboxField({ name, label }: CheckboxFieldProps) {
+  const classes = useStyles()
+  const { t } = useTranslation('one-time-donation')
+  const [field, meta] = useField(name)
+  const helperText = meta.touched ? translateError(meta.error as TranslatableField, t) : ''
+  return (
+    <FormControl required component="fieldset" error={Boolean(meta.error) && Boolean(meta.touched)}>
+      <FormControlLabel
+        label={typeof label === 'string' ? `${t(label)}` : label}
```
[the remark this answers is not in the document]
kachar: Unfortunately, this seems to be necessary as sometimes `label` is a `ReactElement` or `number` and `TFunction` does accept only strings.

src/components/one-time-donation/components/FormikStepper.tsx
```diff
@@ -0,0 +1,113 @@
...
+            </Button>
+            <Button
+              className={classes.btn}
+              startIcon={isSubmitting ? <CircularProgress size="1rem" /> : null}
+              disabled={isSubmitting}
+              variant="contained"
+              color="inherit"
+              type="submit">
+              {isSubmitting ? 'Потвърждение' : isLastStep() ? t('btns.end') : t('btns.next')}
+            </Button>
```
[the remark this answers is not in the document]
kachar: Most of these props can be also found in the `<SubmitButton />` component.

Please check if we can reuse it here with custom `label`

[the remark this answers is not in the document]
kachar: Also the label `потвърждение` is not ranslated as the rest

src/components/one-time-donation/components/Steps.tsx
```diff
@@ -0,0 +1,118 @@
...
+} from '../../../gql/donations'
+import { FormikStep, FormikStepper } from './FormikStepper'
+import { useTranslation } from 'next-i18next'
+import { useRouter } from 'next/router'
+import { useMutation, UseQueryResult } from 'react-query'
+import { CampaignResponse } from 'gql/campaigns'
+import { useViewCampaign } from 'common/hooks/campaigns'
+import { useVaultsList } from 'common/hooks/vaults'
+import { useCreateBankDonation } from 'service/donation'
+import { AxiosError, AxiosResponse } from 'axios'
```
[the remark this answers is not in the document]
kachar: Actually `react-query` is agnostic regarding the transport and here it is using `axios` to send the requests.

However we use the axios error and response just to validate if the error is an axios error.  I think we can have it as is

src/components/one-time-donation/components/Steps.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export default function DonationStepper() {
+  const [status, setStatus] = React.useState(false)
+  const { t } = useTranslation('one-time-donation')
+  const router = useRouter()
+  const slug = String(router.query.slug)
+  const { data }: UseQueryResult<{ campaign: CampaignResponse }> = useViewCampaign(slug as string)
+  const valts = useVaultsList().data
+  const vault = valts?.find((a) => a.campaignId === data?.campaign.id)
```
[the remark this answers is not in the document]
kachar: Vaults should be only an internal abstraction and should never reach the frontend. Why? Because bad actors may try to change the vault ID and send the money elsewhere. 

The donation money transfers to `vaults` should be done only on the backend or with administrative access.

src/components/one-time-donation/components/ThirdStep.tsx
kachar: The router type system of parsed query string is typed as `string | string[]` as the query can be `?name=one` or `?name[]=one&name[]=two`

The funny thing is this can never happen on the page where we've got `campaign/[slug].tsx` but the router type still remains. If Next.js didn't use `router.query` for the pathname params this wouldn't be the case

---

I see another bigger problem here than the types.

We don't really know if the component `ThirdStep` would be used on a page which has variable param `[slug].tsx` in it's path. So? So it would break if this component is used on another page - `router.query.slug` would end up empty.

The best way to handle this is to pass the slug down the stream from the page component.

src/components/one-time-donation/components/FirstStep.tsx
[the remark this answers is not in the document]
borislavstoychev: When i try to fetch then i get 401 unauthorized. Am i need STRIPE_SECRET_KEY?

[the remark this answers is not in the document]
dimitur2204: You need to have your own Stripe account and in the  API `.env.local` you need to set STRIPE_SECRET_KEY and STRIPE_WEBHOOK_SECRET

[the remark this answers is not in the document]
dimitur2204: This is a readme describing it
https://github.com/podkrepi-bg/api/blob/master/TESTING.md

src/components/one-time-donation/components/FormikStepper.tsx
```diff
@@ -0,0 +1,113 @@
...
+import React, { PropsWithChildren, useState } from 'react'
+
+export interface FormikStepProps
+  extends Pick<FormikConfig<FormikValues>, 'children' | 'validationSchema'> {
+  label?: string
+}
+
+export function FormikStep({ children }: FormikStepProps) {
+  return <>{children}</>
+}
```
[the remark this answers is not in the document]
borislavstoychev: I use this to submit field validation to each step separately. There are dependencies between the fields from which the stepper did not work. In this way, each step receives its own validation.

PR #742, opened 2022-04-30
Feature/price radio btn

dimitur2204:
## Changes

* Radio Buttons and Radio Group were in one component  and I think moving them to seperate ones would be better.
* Added the `useSinglePrices` hook to now fetch prices from Stripe.
* Cleaned up the styles for the radio button to now use the theme everywhere and removed fixed widths and heights.
* Added the container style to the whole form which handles the 4 steps.

## Important
* Changed the export of the `theme` to now be another object that we can add values to. I think it shouldn't be a problem, but we have to keep it in mind


src/common/theme.ts
```diff
@@ -137,7 +136,16 @@ export const themeOptions: ThemeOptions = {
...
-const theme = createTheme(themeOptions)
-export default responsiveFontSizes(theme)
+const theme: Theme = createTheme(themeOptions)
+const materialTheme = responsiveFontSizes(theme)
+const podkrepiTheme = {
+  borders: {
+    dark: colors.blue.dark,
+    light: colors.blue.main,
+    roundRadius: 60,
+  },
```
kachar, line 147: I'm not sure why we have this here instead of in `themeOptions` :thinking:

src/components/one-time-donation/components/ThirdStep.tsx
```diff
@@ -75,7 +74,7 @@ export default function ThirdStep() {
         <Typography className={classes.h2}>{t('third-step.title')}</Typography>
       </Grid>
       <Grid item mx={10}>
-        <RadioGroupFormik name="payment" options={options as []} />
+        <PriceRadioGroup name="payment" options={options as []} />
```
kachar, line 77:
```suggestion
        <PriceRadioGroup name="payment" options={options} />
```

src/components/one-time-donation/components/PriceRadioGroup.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useTranslation } from 'next-i18next'
+import { FormControl, FormHelperText, RadioGroup, Typography } from '@mui/material'
+import { TranslatableField, translateError } from 'common/form/validation'
+import PriceRadioButton from './PriceRadioButton'
+import Stripe from 'stripe'
+import { money } from 'common/util/money'
+
+export type RadioGroupProps = {
+  name: string
+  options: Stripe.Price[] | undefined
```
kachar, line 11: Maybe the type here is different

```tsx
options: { values: 'card' | 'bank', label: string }[]
```

Somewhere we have a mismatch

src/common/theme.ts
dimitur2204: `ThemeOptions` are a set type by MUI. I tried putting it there, but it requires a lot of gymnastics with typings and it's not worth the hassle
image

kachar: Alright, sounds reasonable, thanks for the explanation :+1:

src/components/one-time-donation/components/PriceRadioGroup.tsx
dimitur2204: This is happening, because the same radio group is used for the Third Step later. I will see how to make the whole group more reusable so we can use it later on too.

src/components/one-time-donation/components/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,68 @@
...
+    value: string | number
+  }[]
+  muiRadioGroupProps?: Partial<RadioGroupProps>
+  /**
+   * Props that get passed to the `<Grid>` that wraps around each rendered button
+   * @example
+   * <Grid {...propsYouPassed}>
+   *  <RadioButton>
+   * <Grid />
+   */
```
dimitur2204, line 28: @kachar  What do you think about putting JSDocs on all of the common components is it worth the work?

src/components/one-time-donation/components/RadioButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+          checked={checked}
+          onChange={onChange}
+          {...muiRadioButtonProps}
+        />
+      }
+    />
+  )
+}
+
+export default RadioButton
```
dimitur2204, line 66: Should we move this and the `<RadioButtonGroup />` to components/common?

kachar (approved):

src/components/one-time-donation/components/RadioButtonGroup.tsx
kachar: Nice idea, it will give good insight on how they are supposed to be used around

PR #744, opened 2022-04-30
Add rounded corners to inputs on the theme

dimitur2204:
## Motivation and context

Changed the styling of the inputs in the `muiTheme` we are using. 

The corners of the oneline input fields are now fully round and the corners of the mulitline input are more round than they were before

## Screenshots:

|image|image|

## Testing

Went through many of the pages and didn't find any bugs whatsoever 🚀!

### DIFF
--- a/src/common/theme.ts
+++ b/src/common/theme.ts
@@ -95,6 +95,30 @@ export const themeOptions: ThemeOptions = {
       styleOverrides: {
         root: {
           fontSize: '1rem',
+          borderRadius: 60,
+        },
+        multiline: {
+          borderRadius: 20,
+        },
+      },
+    },
+    MuiFilledInput: {
+      styleOverrides: {
+        root: {
+          borderRadius: 60,
+        },
+        multiline: {
+          borderRadius: 20,
+        },
+      },
+    },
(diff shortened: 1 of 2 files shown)


kachar (commented):
@dimitur2204 Do we need to revoke custom border radius styles from any inputs or buttons so the theme can kick in everywhere?

dimitur2204:
> @dimitur2204 Do we need to revoke custom boder radius styles from any inputs or buttons so the theme can kick in everywhere?

What exactly do you need

kachar:
@dimitur2204 If we use custom border style on some inputs or buttons, the newly added global values won't override the local styles.

So I wonder if we need to revise the usage of `borderRadius` in the whole project?

kachar (approved):
I've reviewed every usage of `borderRadius` and it seems none of them are related to inputs.

I'm not a particular fan of the so rounded inputs, but  let's roll with this update :rocket:

dimitur2204:
> 

I am not particularly a fan too, but design said that's how the UI Kit is. 

For the border radiuses being repeated everywhere I would change that once the changes to the theme from #742 get merged into master some time in the future. Hopefully I remember :).

PR #746, opened 2022-04-30
Added background to pdf

PepiPetrov:
Added background to pdf as it was missing.

Screenshot:


src/components/pdf/Certificate.tsx
```diff
@@ -71,7 +100,8 @@ export default function Certificate({ donation, person }: Props) {
   const formattedDate = formatDateString(donation.createdAt)
   return (
     <Document title="Дарение">
-      <Page size="A4" style={styles.page}>
```
kachar, line 74: Why do we change the page size from `A4` to `LETTER`?

src/components/pdf/Certificate.tsx
```diff
@@ -14,7 +14,15 @@ Font.load({ fontFamily: 'Roboto' })
...
   page: {
-    backgroundColor: '#E1E4E5',
+    backgroundColor: '#fff',
+  },
+  backgroundImage: {
+    position: 'absolute',
+    height: '100%',
+    width: '100%',
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
kachar, line 24: What's the error we ignore here?

Maybe it's better to use `@ts-expect-error` instead of `@ts-ignore`

src/components/pdf/Certificate.tsx
PepiPetrov: The size is being changed as if document size equals A4, background could not be set because content goes to the next page, instead of staying over background.

kachar (approved):
Nicely done @PepiPetrov

PR #748, opened 2022-05-01
Fix/step 1 styling

dimitur2204:
## Motivation and context

Improvements to #736:

* There were some issues with some of the styling on the first step page, but that also required many changes to the general `FormikStepper` and `OneTimeDonationPage`.
* Changed how the `Layout` component was implemented and now it's a similar implementation as it is in the Campaign details page, since the layout there also includes the photo in a circle with a banner photo on the back.
* Added the normal `CheckboxField` and `FormTextField` inputs and if they need changing we would have to change them globally if that is fine with the design team
* Now fetching the sources of the images from the actual campaign

### Old:
| Mobile              | Desktop              |
|first-step-sm-old|first-step-lg-old|

### New:
| Mobile              | Desktop              |
|first-step-sm-new|first-step-lg-new|


src/components/common/form/RadioButtonGroup.tsx
```diff
@@ -36,21 +36,28 @@ export default function RadioButtonGroup({
...
   return (
-    <FormControl required component="fieldset" error={Boolean(meta.error) && Boolean(meta.touched)}>
-      <RadioGroup value={field.value} name={name} {...muiRadioGroupProps}>
+    <FormControl
+      fullWidth
+      required
+      component="fieldset"
+      error={Boolean(meta.error) && Boolean(meta.touched)}>
+      <RadioGroup
+        onChange={(e, v) => {
```
kachar, line 48: Great touch for moving the change handler to the radio group instead of individual buttons

kachar (approved):
Great move changing the fields to the theme defaults. More customizable and the same everywhere.

Awesome progress with this PR @dimitur2204

PR #751, opened 2022-05-02
Updated getting of data for campaign grid

Bunny303:
## Motivation and context

At the moment campaigns are filtering by state (only active and comleted are shown). As admin I have to see all of the campaings and new route is used that returns all campaings with needed data.

### DIFF
--- a/src/common/hooks/campaigns.ts
+++ b/src/common/hooks/campaigns.ts
@@ -10,6 +10,10 @@ export function useCampaignList() {
   return useQuery<CampaignResponse[]>(endpoints.campaign.listCampaigns.url)
 }
 
+export function useCampaignAdminList() {
+  return useQuery<CampaignResponse[]>(endpoints.campaign.listAdminCampaigns.url)
+}
+
 export function useCampaignTypesList() {
   return useQuery<CampaignType[]>(endpoints.campaignTypes.listCampaignTypes.url)
 }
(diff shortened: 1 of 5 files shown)


src/common/hooks/campaigns.ts
```diff
@@ -4,12 +4,21 @@ import { useKeycloak } from '@react-keycloak/ssr'
...
+} from 'gql/campaigns'
 
 export function useCampaignList() {
   return useQuery<CampaignResponse[]>(endpoints.campaign.listCampaigns.url)
 }
 
+export function useCampaignAdminList() {
+  return useQuery<AdminCampaignResponse[]>(endpoints.campaign.listAdminCampaigns.url)
+}
+
```
kachar, line 21: We may need to send authorization JWT in the headers of this request, otherwise, the backend will throw 401 unauthorized error.

```suggestion
export function useCampaignAdminList() {
  const { keycloak } = useKeycloak<KeycloakInstance>()
  return useQuery<AdminCampaignResponse[]>(
    endpoints.campaign.listAdminCampaigns.url,
    authQueryFnFactory<CampaignResponse>(keycloak?.token),
  )
}

```

kachar (commented):
@Bunny303 Can you please verify if we really need to send the auth header in the request for admin campaigns

src/common/hooks/campaigns.ts
Bunny303: Added, route is not public so it's needed.

kachar (approved):
Top :top:

PR #755, opened 2022-05-03
Styling and structure step 2

dimitur2204:
## Motivation and context

Step 2 was not finished and style properly. I did not stick to the designs exactly as they were but I think it came out pretty good.

I am thinking of using that form for a base for the Login form we would have to actually login people instead of the keycloak implementation we have right now.

Changed the validation schema in a way that is probably not perfect right now, but the whole form would have to get reworked either way so it thing that would be a separate whole PR.

## Screenshots:

step-2-new


dimitur2204:
In the last commit I reverted some changes that I had made to the theme. They were regarding the outline buttons and the hover they had. I didn't like it much is that the way it is actually supposed to be?
outlined-hover

src/components/one-time-donation/components/AnonymousForm.tsx
```diff
@@ -0,0 +1,68 @@
+import { Box, Collapse, Grid, Typography } from '@mui/material'
+import * as React from 'react'
+import FormTextField from 'components/common/form/FormTextField'
+import CircleCheckboxField from 'components/common/form/CircleCheckboxField'
+import { useField } from 'formik'
+import { useTranslation } from 'next-i18next'
+import { ExpandLess, ExpandMore } from '@mui/icons-material'
+
+export default function AnonymousMenu() {
```
kachar, line 9: We may need to choose between `AnonymousForm` (filename) or `AnonymousMenu` (component name) and stick with one of those

src/components/one-time-donation/components/FormikStepper.tsx
```diff
@@ -50,10 +51,18 @@ export function FormikStepper<T>({ children, ...props }: GenericFormProps<T>) {
...
   function isLastStep() {
     return step === childrenArray.length - 2
   }
+
+  function isFirstStep() {
+    return step === 0
+  }
+
+  function isLogged() {
+    return currentPerson && currentPerson.status !== 'unauthenticated'
```
kachar, line 64: This is not really safe as the `status` can be `undefined` or empty string and they still would be considered `logged`

src/components/one-time-donation/components/Steps.tsx
```diff
@@ -63,14 +60,13 @@ export default function DonationStepper() {
...
-        personsFirstName: values.personsFirstName,
-        personsLastName: values.personsLastName,
-        personsPhone: values.personsPhone,
+        personsEmail: values.personEmail,
+        personsFirstName: values.personFirstName,
+        personsLastName: values.personLastName,
+        personsPhone: values.personPhone,
         extCustomerId: String(Math.random() * 5),
         extPaymentIntentId: String(Math.random() * 5),
         extPaymentMethodId: String(Math.random() * 5),
```
kachar, line 69: We don't need those at this point

src/components/one-time-donation/components/AnonymousForm.tsx
kachar:
```suggestion
export default function AnonymousForm() {
```

kachar (approved):

PR #758, opened 2022-05-09
Added bankHash and allowDonationOnComplete in Campaign types

Bunny303:
## Motivation and context

Properties bankHash and allowDonationOnComplete were added into the campaign model and they are returning to client api at the moment. In order to be used both should be added into the relevant types.


kachar (approved):
Precise update, thanks @Bunny303

Issue #759, opened 2022-05-10
[Bug] FAQ Tabs UI bug


dimitur2204:
**Describe the bug**
The tabs to change the questions hide when the screen is bigger.

**To Reproduce**

1. Go to '/faq'
2. See error

**Expected behavior**

**Screenshots**
#### How it should be:
![image](https://user-images.githubusercontent.com/61479393/167722058-73344850-807b-482b-9b7f-89f5dff61eb2.png)

#### How it is:
![image](https://user-images.githubusercontent.com/61479393/167722022-923873da-6b10-41d1-8ca3-4573e2075b7c.png)

**Desktop (please complete the following information):**

- OS: macOS
- Browser Chrome

Issue #555, opened 2022-02-25
[Feature] New page> Login/Register option before a donation


AlbenaGeleva:
Currently the Donate button on a campaign sends you to the online payment transaction site. Unless the user is already logged in, the donation will be anonymous. 

Proposal: from https://dev.podkrepi.bg/campaigns/for-the-children when clicking Donate, point to a new page offering the user to log in or register (not mandatory). See sample text and format attached:

[Podkrepi Campaigns - Landing page from Donate button.docx](https://github.com/podkrepi-bg/frontend/files/8141861/Podkrepi.Campaigns.-.Landing.page.from.Donate.button.docx)

Issue #450, opened 2022-01-25
FAQ page/ not working SSR support for better SEO


Polina1985:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

igoychev:
Tested with Lighthouse dev tool and SEO looks fine for the FAQ page. See screenshot below:
<img width="585" alt="Screenshot 2022-04-19 at 14 43 44" src="https://user-images.githubusercontent.com/7055304/163996275-4dd9e32e-5e32-452f-bda4-56f6fc564494.png">

kachar:
This issue is not resolved IMO

![image](https://user-images.githubusercontent.com/893608/163999157-66dafb8e-a72c-46eb-a027-c16da56c5d1d.png)

In the view-source:https://dev.podkrepi.bg/faq of the page we cannot see the actual text from the page, no matter the SEO score is 100%

If we can't see the text from the FAQ answers, neither can Google Search

I'll keep this issue open until we figure it out.

Issue #371, opened 2021-10-11
Create custom authentication


kachar:
The motivation here is to reduce the friction between different systems and provide smooth UX in a single application.

At the moment users are being sent to https://keycloak.podkrepi.bg/auth/realms/webapp-dev/protocol/openid-connect/auth in order to login to https://podkrepi.bg and this creates confusion.

Our API layer now supports authentication via `/login` and `/register` endpoints (Check [swagger](https://dev.podkrepi.bg/docs/))

What needs to be done:

- [ ] Replace Keycloak authentication screens (login & registration) with internal forms
  - [ ] Replace handler for login form
  - [ ] Replace handler for registration form
  - [ ] Handle cookie storage (either manually on the frontend or via XHR cookie by the backend)
- [ ] Recognize the session cookie and provide it to the `SSRKeycloakProvider`

atanasster:
I’ll give it a go

kachar:
Closing this issue as we'll stick with Keycloak authentication forms with a custom design.

All other solutions are quite complex and will require many many more hours to be polished and functional. In a perfect scenario, we would have enough people to complete it, but at this point in time, we don't have that luxury and need to stick to a more simple solution.

The current Keycloak solution is sufficient both on the backend and frontend as it provides ways to check for the permissions based on JWT token without custom implementations.

@atanasster Thanks for the energy and brainpower you've put into the PR.

![image](https://user-images.githubusercontent.com/893608/149507568-7822a210-8020-4727-99a4-5edc3b143c62.png)

PR #765, opened 2022-05-13
Replace keycloak with next auth

kachar:
## Motivation and context

 
- Fixes https://github.com/podkrepi-bg/frontend/issues/555
- Fixes https://github.com/podkrepi-bg/frontend/issues/371
- Fixes https://github.com/podkrepi-bg/frontend/issues/450
- Fixes https://github.com/podkrepi-bg/frontend/issues/759
- Fixes https://github.com/podkrepi-bg/frontend/issues/768
- Fixes https://github.com/podkrepi-bg/frontend/issues/766

Related to 
- #408 

### Migrate from Keycloak to NextAuth

When we use Keycloak sessions we need to redirect the user to the auth server on each refresh of the page. By replacing it with NextAuth we're able to keep the JWT in session cookie locally and have a customized auth experience.

- https://next-auth.js.org/
- https://github.com/nextauthjs/next-auth-example

### Migration from JSS to styled / sx

Replacing the Keycloak implementation on the frontend revealed that the migration to MaterialUI v5 had flows and the styles were broken during SSR of the pages. Previously the whole application have been client-rendered so this was not a problem.

With this PR we also migrate JSS to `styled` and `sx` 

- https://mui.com/material-ui/guides/migration-v4/#migrate-from-jss
- MUI v5 integration https://github.com/mui/material-ui/tree/master/examples/nextjs-with-typescript

Automatic codemod has been the base for the migration:

```
npx @mui/codemod v5.0.0/jss-to-styled .
```

## Screenshots:

Before|After
---|---
image|image
image|image

## Testing

Update `.env.local` according to `.env.local.example`

```
NEXTAUTH_URL=http://localhost:3040
NEXTAUTH_SECRET=super-dev-secret
```

**New dependencies**:

- `next-auth`

**Removed dependencies**:

- `keycloak-js`
- `@react-keycloak/ssr`


ani-kalpachka (approved):

src/components/auth/login/LoginPage.tsx
```diff
@@ -1,23 +1,40 @@
...
+                  <Grid item key={provider.name}>
+                    <Button
+                      fullWidth
+                      variant="outlined"
+                      onClick={() => signIn(provider.id, { callbackUrl: baseUrl })}>
+                      {t('nav.login-with')} {provider.name}
+                    </Button>
+                  </Grid>
+                ))}
+          </Grid>
```
dimitur2204, line 35: This should probably be a part of the login form, also the providers can have an `icon` property

src/components/auth/profile/DonationTab.tsx
```diff
@@ -28,11 +28,18 @@ import { ProfileTabs } from './tabs'
...
-  donationsBox: {
+const PREFIX = 'DonationTab'
+
+const classes = {
+  donationsBox: `${PREFIX}-donationsBox`,
+  donationsBoxRow: `${PREFIX}-donationsBoxRow`,
+}
+
+const StyledProfileTab = styled(ProfileTab)({
+  [`& .${classes.donationsBox}`]: {
```
dimitur2204, line 39: Is this the way we should do class styling now?

src/components/auth/register/RegisterForm.tsx
```diff
@@ -36,15 +39,41 @@ const defaults: RegisterFormData = {
...
+  const { mutateAsync: register } = useRegister()
 
   const onSubmit = async (values: RegisterFormData) => {
-    const result = await keycloak?.register({
-      loginHint: values.email, // Doesn't work with registration, see https://stackoverflow.com/q/63117669/668245
-      redirectUri: `${baseUrl}${routes.profile.index}`,
-    })
-    console.log(values)
-    console.log({ result })
+    AlertStore.show(t('auth:alerts.welcome'), 'success')
```
dimitur2204, line 48: This should probably not be here since down in the `resp?.ok` we have an alert showing welcome. I am not sure about that though

src/components/auth/login/LoginForm.tsx
```diff
@@ -25,85 +24,65 @@ const validationSchema: yup.SchemaOf<LoginFormData> = yup
...
 const defaults: LoginFormData = {
   email: '',
   password: '',
-  csrfToken: '',
 }
 
 export type LoginFormProps = {
   initialValues?: LoginFormData
-  csrfToken: string
 }
```
dimitur2204, line 36: Maybe we can have a `returnUrl` on the login form that you pass and when the user logs in it returns him there. By default it can be the `baseUrl` or `undefined`

src/components/auth/profile/UpdateBirthdayModal.tsx
```diff
@@ -16,6 +16,29 @@ import { format, parse, isDate } from 'date-fns'
...
+}
+
+const StyledModal = styled(Modal)({
+  [`& .${classes.modal}`]: {
+    position: 'absolute',
+    top: '50%',
+    left: '50%',
+    transform: 'translate(-50%, -50%)',
+    width: 650,
+    backgroundColor: '#EEEEEE',
```
dimitur2204, line 33: Should probably add this color somewhere in the theme

src/components/benefactor/AddBenefactorForm.tsx
```diff
@@ -85,7 +76,14 @@ export default function AddBenefactorForm({ initialValues = defaults }: Benefact
...
   return (
-    <Grid container direction="column" component="section" className={classes.container}>
+    <Grid
+      container
+      direction="column"
+      component="section"
+      sx={{
+        maxWidth: '700px',
+        margin: '0 auto',
+      }}>
```
dimitur2204, line 86: You can give the grid a `maxWidth` prop which can be a predefined from the theme either `sm`,`md`,`lg` etc. We should generally try to use it instead of random pixel values

dimitur2204 (approved):
Incredible changes ❤️! We can finally login without keycloak!!! There are some minor things

src/components/auth/login/LoginForm.tsx
dimitur2204: That can be a seperate issue though for the LoginForm as a whole!

src/components/auth/profile/DonationTab.tsx
kachar: This is the automated JSS to styled migration

We can minimize the usage of styled as much as possible with `sx` property and reusable components instead of reusable styles.

I'm not a big fan of the way styles are re-defined, so I did take an initiative the change them after the migration to local `sx` props, but it's too much to be handled at once

src/components/auth/login/LoginForm.tsx
kachar: There's a query param `callbackUrl` which is handled after login

http://localhost:3040/login?callbackUrl=%2Fadmin

On `LoginForm` we can add a `onLogin` prop which will handle the callback with a default redirect

Let's do that in a separate issue/PR

src/components/benefactor/AddBenefactorForm.tsx
kachar: Definitely no px values

Maybe the Grid component should not have any width at all but be wrapped in `<Container maxWidth="lg">` or similar.

However I've just migrated the styles, didn't tweak around

src/components/auth/profile/UpdateBirthdayModal.tsx
kachar: We can also use the default MUI colors palette - `gray[200]` 

https://mui.com/material-ui/customization/color/#examples

PR #764, opened 2022-05-12
Feature/mp 763 homepage carousels

dimitur2204:
## Motivation and context

Should close #763

Made changes to the theme regarding the yellow color. Should close  #760!

Added Swiper as a dependency to create the carousels

Added carousels instead of just cards on the WhatUnitesUs and TeamMembersSections

### Desktop
| Before              | After              |
| ------------------- | ------------------ |
|image|image|

### Mobile
| Before              | After              |
| ------------------- | ------------------ |
|image|image|

image

image

image

## Testing

Tested the page on all resolutions!

PR #769, opened 2022-05-13
Fix borderRadius styles

ani-kalpachka:
- Inherited the borderRadius styles from the theme - theme.shape.borderRadius (25px).
- Removed the style where it is not used.
- Removed unused styles on Homepage.

- borderRadius is still inlined in Admin panel where it is not inherited from the theme, but has the default value of 25px.
- borderRadius: '50%' is saved, it makes the icons round.

### DIFF
--- a/src/common/theme.ts
+++ b/src/common/theme.ts
@@ -61,7 +61,7 @@ export const themeOptions: ThemeOptions = {
     },
   },
   shape: {
-    borderRadius: 3,
+    borderRadius: '25px',
   },
   components: {
     MuiLink: {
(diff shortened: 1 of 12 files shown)


kachar:
Looks like there are other 2 PRs that are changing the same files so they will conflict a lot

- https://github.com/podkrepi-bg/frontend/pull/765
- https://github.com/podkrepi-bg/frontend/pull/764

Let's check if those two are merged we still have to do the same updates. I assume we can port just the updates instead of dealing with the merge conflict.

Let's check back once those two PRs are merged

PR #774, opened 2022-05-15
Add returnUrl to secured props and made the login to redirect to crea…

### LINKED ISSUE #363: Require login for campaign creation

dimitur2204:
## Motivation and context

Closes #363.

Added a returnUrl to the secured props with translation since when next return `ctx.req.url` as a callbackUrl it returns this
image

Is redirecting like that with a returnUrl parameter ok or is there something more to it?


kachar:
@dimitur2204 Maybe we can do some whitelisting for the redirects, indeed the `/development` part is not intended

kachar (commented):
@dimitur2204 Let's add whitelist for the `redirectUrl` param in `securedProps` so we exclude all requests starting with:

- `/_error`
- `/_next`

kachar (approved):
:clap:

PR #775, opened 2022-05-16
Added caling of pdf generation, updated data fetch in the handler

Bunny303:
## Motivation and context

In the list with user donations there is button that has to generate certificate for each donation.

### DIFF
--- a/src/pages/api/pdf/certificate/[donationId].tsx
+++ b/src/pages/api/pdf/certificate/[donationId].tsx
@@ -18,14 +18,14 @@ const Handler: NextApiHandler = async (req: NextApiRequest, res: NextApiResponse
   )
 
   if (!donation) {
-    res.send('What should happen?')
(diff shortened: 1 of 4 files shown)


src/pages/api/pdf/certificate/[donationId].tsx
```diff
@@ -1,28 +1,29 @@
...
   )
 
-  let person: PersonResponse | undefined = undefined
-  if (donation?.personId) {
-    const { data } = await apiClient.get<PersonResponse>(
-      endpoints.person.viewPerson(donation.personId).url,
-    )
-    person = data
+  if (!donation) {
+    res.send('What should happen?')
```
Comment (Bunny303) [inline src/pages/api/pdf/certificate/[donationId].tsx:21]: What should happen in this case? Donation could not be find if is not owned by logged user or if it's for draft donation or it's just invalid. 
Is there a page that should be returned?

kachar (approved):

PR #779, opened 2022-05-18
Add refresh token functionality

dimitur2204:
## Motivation and context

Depends on this PR on the backend https://github.com/podkrepi-bg/api/pull/246
Added a call to the refresh token endpoint with the `refreshToken` when the `accessToken` has expired.

Cleaned up the `session` object, now all that is passed from the server are tokens and the parsing is on the client side to take out the user information and store it in the session.

## Testing

Waited for the token to expire and it refreshes it correctly.

### DIFF
--- a/src/common/util/roles.ts
+++ b/src/common/util/roles.ts
@@ -51,7 +51,7 @@ export const canViewSupporters = (sessionRoles: SessionRoles): boolean => {
   )
 }
 
-export const isAdmin = (sessionRoles?: SessionRoles, session?: Session): boolean => {
+export const isAdmin = (sessionRoles?: SessionRoles, session?: Session | null): boolean => {
   if (sessionRoles) {
     return (
       canViewContactRequests(sessionRoles as SessionRoles) &&
--- a/src/common/util/roles.ts
+++ b/src/common/util/roles.ts
@@ -1,5 +1,7 @@
-import { Session } from 'next-auth'
-
+export type SessionRoles = {
+  realmRoles: RealmRole[] | undefined
+  resourceRoles: ResourceRole[] | undefined
+}
 export type RealmRole =
   | 'view-supporters'
   | 'view-contact-requests'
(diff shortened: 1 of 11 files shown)


src/pages/admin/_middleware.ts
```diff
@@ -1,13 +1,13 @@
...
 import withAuth from 'next-auth/middleware'
+import { getSession } from 'next-auth/react'
 
 export default withAuth({
   callbacks: {
-    authorized: async ({ req }) => {
-      const jwt = await getToken({ req })
-      return isAdmin(jwt)
+    authorized: async () => {
+      const session = await getSession()
```
kachar, line 9: Unfortunately we can't get the session in the middleware

I think this is the main cause for the failed next build

dimitur2204:
There are problems still that I haven't resolved. So I am putting the do not mergre flag

src/pages/admin/_middleware.ts
kachar: I think we can still use the `await getToken` instead of using the `{ token }` from the `authorize` fn

This did work in the previous revision

dimitur2204:
It should now be working :) @kachar

src/common/util/roles.ts
```diff
@@ -27,28 +31,37 @@ export const roles = {
...
+  if (session) {
+    const sessionRoles: SessionRoles = {
+      realmRoles: session.user?.realm_access.roles,
+      resourceRoles: session.user?.resource_access.account.roles,
+    }
+    return (
+      canViewContactRequests(sessionRoles as SessionRoles) &&
+      canViewSupporters(sessionRoles as SessionRoles)
+    )
+  }
```
kachar, line 65: I feel like these `as` will cover some undefined cases

I suggest we remove them and apply defaults at the level where we define the roles construct

```suggestion
  if (session) {
    const sessionRoles: SessionRoles = {
      realmRoles: session.user?.realm_access.roles ?? [],
      resourceRoles: session.user?.resource_access.account.roles ?? [],
    }
    return (
      canViewContactRequests(sessionRoles) &&
      canViewSupporters(sessionRoles)
    )
  }
```

src/pages/api/auth/[...nextauth].ts
```diff
@@ -82,25 +85,25 @@ declare module 'next-auth' {
    * Session object available everywhere
    */
   export interface Session {
-    user: ServerUser
     accessToken: string
-    refreshToken: string
+    user: ServerUser | null
+    error?: string
```
Comment (kachar) [inline src/pages/api/auth/[...nextauth].ts:90]: I can't find the usecase for this `error` prop as we assign only the other two

src/pages/admin/_middleware.ts
kachar: If it works - it doesn't matter :)

It was a suggestion for the failed build, which seems to have failed for a different reason.

Resolving the convo

src/pages/api/auth/[...nextauth].ts
Comment (dimitur2204) [reply src/pages/api/auth/[...nextauth].ts]: Done in the last commit [34dde0f](https://github.com/podkrepi-bg/frontend/pull/779/commits/34dde0fd19678adbda16716ca6f8ee45a440fda4)

kachar (approved):

kachar (approved):
wowoowow. so fresh :lemon:

PR #785, opened 2022-05-21
Feature/770 google provider login

dimitur2204:
## Motivation and context

Add functionality to login with Google which sends a call to the API with the Google Access token from the NextAuth server layer. Then it takes the `keycloakAccess` token and passes it on to the frontend as a normal user with credentials would

## Testing

Tested all pages and the token refresh. Everything seems to work!

You would now need to provide  and GOOGLE_SECRET in order to be able to
- `GOOGLE_ID` : Google ID from the Google console
- `GOOGLE_SECRET` : Google secret from the Google console

**New dependencies**:

- `jwtDecode` : An auth0 maintaned library to decode JWT's since the old way as blowing unexpectedly sometimes.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -27,5 +27,5 @@ FEATURE_CAMPAIGN=true
 ##############
 NEXTAUTH_URL=http://localhost:3040
 NEXTAUTH_SECRET=super-dev-secret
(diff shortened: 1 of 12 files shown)


kachar (approved):
Great updates @dimitur2204 

Let's roll it to staging to confirm the functionality

PR #791, opened 2022-05-25
Fix borderRadius styles

ani-kalpachka:
- Inherited the borderRadius styles from the theme - theme.shape.borderRadius (25px).
- Removed the style where it is not used.
- borderRadius is still inlined in Admin panel where it is not inherited from the theme, but has the default value of 25px.
- borderRadius: '50%' is saved, it makes the icons round.


kachar (approved):
Awesome work @ani-kalpachka! Thanks for re-applying the updates on top of master

PR #801, opened 2022-05-31
799 one time donation a custom amount with an input field

borislavstoychev:
## Screenshots:

| image | imageimageimage |

## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:


kachar:
@podkrepi-bg/design Can you please double confirm the UI elements of this update

src/components/one-time-donation/Steps.tsx
```diff
@@ -65,9 +67,13 @@ export default function DonationStepper() {
     try {
       const data = {
         currency: 'BGN',
-        priceId: values.amount,
+        priceId: values.amount !== 'other' ? values.amount : undefined,
+        amount:
+          values.amount === 'other'
+            ? Math.round((values.otherAmount + Number.EPSILON) * 100)
```
kachar, line 73: @borislavstoychev What's the goal here using `Number.EPSILON` ?

src/components/one-time-donation/steps/FirstStep.tsx
```diff
@@ -124,6 +136,32 @@ export default function FirstStep() {
...
             }
           />
+          <Collapse in={amount.value === 'other'} timeout="auto">
+            <Grid
+              style={
+                !mobile
+                  ? {
+                      float: 'right',
+                      marginTop: theme.spacing(-10),
+                      width: 325,
```
kachar, line 146: Using hardcoded `width: 325` might be too restrictive for the responsive styles.

Maybe using `100%` will deliver the same results, can you confirm

src/components/one-time-donation/Steps.tsx
borislavstoychev: We take the number and add a very small number, ```Number.EPSILON``` , to ensure the number’s accurate rounding.

kachar (approved):
Thanks for the initiative @borislavstoychev 

I'm okay that we merge this PR now in order to add the custom amount functionality to the app, but we need at least one more iteration to enhance the UI and UX

At the moment the labels leave the provided input space and this is sub-optimal in my view. However, this is not a problem of this PR and should be addressed by the general app theme for the page together with the @podkrepi-bg/design team.

Kudos for the energy spent on this task :zap:

PR #803, opened 2022-05-31
login form and register form with password visibility

borislavstoychev:
## Screenshots:

| image|image |

## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/components/auth/login/LoginForm.tsx
+++ b/src/components/auth/login/LoginForm.tsx
@@ -1,5 +1,5 @@
 import * as yup from 'yup'
-import { Button, Grid, IconButton, InputAdornment } from '@mui/material'
+import { Button, Grid } from '@mui/material'
 import React, { useState } from 'react'
 import { useRouter } from 'next/router'
 import { signIn } from 'next-auth/react'
@@ -13,7 +13,7 @@ import GenericForm from 'components/common/form/GenericForm'
 import SubmitButton from 'components/common/form/SubmitButton'
 import FormTextField from 'components/common/form/FormTextField'
 import Google from 'common/icons/Google'
-import { Visibility, VisibilityOff } from '@mui/icons-material'
+import PasswordVisibility from 'components/common/PasswordVisibility'
 
 export type LoginFormData = {
   email: string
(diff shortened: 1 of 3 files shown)


kachar (commented):
@borislavstoychev This is an iteration in a good direction - reusability!

If we extend the scope to the `ChangePasswordForm` and `ResetPasswordForm` then we're gonna copy the `endAdornment` two more times.

We can also extract the whole ` <FormTextField type={passwordType} name="password" ... />` in a dedicated `PasswordField` component which we can safely use in all forms.

In this case the form does not need to know for the state of the password visibility - it can be encapsulated in the new component `PasswordField`.

It can be something similar to:


```tsx
export default function PasswordField({ name = 'password', ...props }: TextFieldProps) {
  const [showPassword, setShowPassword] = useState(false)
  const handleShowPassword = () => setShowPassword((show) => !show)
  return (
    <FormTextField
      fullWidth
      name={name}
      {...props}
      type="password"
      autoComplete="current-password"
      label="Password"
      InputProps={{
        endAdornment: (
          <InputAdornment position="end">
            <IconButton onClick={handleShowPassword} edge="end">
              <Iconify icon={showPassword ? 'eva:eye-fill' : 'eva:eye-off-fill'} />
            </IconButton>
          </InputAdornment>
        ),
      }}
    />
  )
}
```

What are your thoughts on the topic :thought_balloon:

borislavstoychev:
> @borislavstoychev This is an iteration in a good direction - reusability!
> 
> If we extend the scope to the `ChangePasswordForm` and `ResetPasswordForm` then we're gonna copy the `endAdornment` two more times.
> 
> We can also extract the whole ` <FormTextField type={passwordType} name="password" ... />` in a dedicated `PasswordField` component which we can safely use in all forms.
> 
> In this case the form does not need to know for the state of the password visibility - it can be encapsulated in the new component `PasswordField`.
> 
> It can be something similar to:
> 
> ```tsx
> export default function PasswordField({ name = 'password', ...props }: TextFieldProps) {
>   const [showPassword, setShowPassword] = useState(false)
>   const handleShowPassword = () => setShowPassword((show) => !show)
>   return (
>     <FormTextField
>       fullWidth
>       name={name}
>       {...props}
>       type="password"
>       autoComplete="current-password"
>       label="auth:fields.password"
>       InputProps={{
>         endAdornment: (
>           <InputAdornment position="end">
>             <IconButton onClick={handleShowPassword} edge="end">
>               {showPassword ? <Visibility /> : <VisibilityOff />}
>             </IconButton>
>           </InputAdornment>
>         ),
>       }}
>     />
>   )
> }
> ```
> 
> What are your thoughts on the topic thought_balloon

I fully support. Thanks for the example :ok_hand:

kachar (approved):
Awesome upgrade @borislavstoychev :top:

PR #809, opened 2022-06-02
805 bug big loading circle on campaign list page

borislavstoychev:
## Screenshots:

| image |image |

| Chrome-empty              | Mozilla-empty              |
|image|image|
## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/components/campaigns/CampaignFilter.tsx
+++ b/src/components/campaigns/CampaignFilter.tsx
@@ -1,6 +1,6 @@
 import React, { useMemo, useState } from 'react'
 import { styled } from '@mui/material/styles'
-import { IconButton, ImageList, Typography } from '@mui/material'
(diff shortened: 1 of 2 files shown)


kachar (changes requested):
The problem here seems to be caused by emotion style engine and the version of Next.js

You might find more info in the related ticket https://github.com/vercel/next.js/issues/36675

I don't think we need to update the markup especially providing a hardcoded value of 400+ px for the loader.

What we need to do is to update to the latest version of `next` - it's being fixed in `12.1.7-canary.18`

kachar (approved):
@borislavstoychev I've updated the PR to move the loading component to the parent component and reduced the loading circle size to `3rem`.

Indeed this particular problem was not caused by Emotion engine

image

Issue #818, opened 2022-06-10
[Bug][Admin] Create Beneficiery Form is not working


igoychev:
**Describe the bug**
The create beneficiary form is not working - the button is disabled although there is data in all fields.

**To Reproduce**

1. Go to admin/beneficiary/create
2. Fill the form fields
3. Press the Submit button
4. No error - nothing happens

**Expected behavior**
A new beneficiary is created and form returns to refreshed beneficiary grid

PR #821, opened 2022-06-12
fixing admin create beneficiery issue 818

quantum-grit:
## Motivation and context
Fixes the Admin Create Beneficiary form as reported in issue https://github.com/podkrepi-bg/frontend/issues/818

- fixed: create beneficiery form. reworked PersonSelect and added info text for better UX in Campaign Creation page
- fixed: profile page breaks if there are 0 campaigns for the user profile to see

## Testing
manually and running the api tests


src/components/beneficiary/BeneficieryForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+export default function BeneficieryForm() {
+  const queryClient = useQueryClient()
+  const router = useRouter()
+  const { t } = useTranslation()
+  let id = router.query.id
+
+  const [beneficiaryType, setBeneficiaryType] = useState<string>('')
+  const [personRelation, setPersonRelation] = useState<string>('')
+  const [companyId, setCompanyId] = useState<string>('')
```
kachar, line 66: We can use formik fields instead of state to set those

src/components/beneficiary/CreatePage.tsx
```diff
@@ -13,7 +13,8 @@ export default function CreatePage() {
     <AdminLayout>
       <AdminContainer title={t('all')}>
         <Container maxWidth="md" sx={{ py: 5 }}>
-          <Form />
+          {t}
```
kachar, line 16:
```suggestion
```

kachar (approved):
Right on point :dart:

PR #847, opened 2022-06-29
842 feature currency selection when creatingediting campaign

### LINKED ISSUE #842: [Feature] Currency selection when creating/editing campaign
**Is your feature request related to a problem? Please describe.**
When creating or editing a non-active campaign, there is a need to select the currency type.

**Describe the solution you'd like**
In Admin Create/Edit form, add a dropdown for selecting currency. For MVP it can be hardcoded to BGN and EUR only.
In Admin Edit form, the currency selector should be enabled only for non-active campaigns that do not have donations.

nradkova:
## Motivation and context
closes #842 

Also added some fixes to the campaign admin:
- formatting date in details modal
- deleting duplicated text in the description field in details modal

- providing the second argument to the money function
-- previous
Screenshot (128)

--now
Screenshot (129)


nradkova (commented):
@igoychev, please check the condition for disabling the currency select button. Not sure if there should be other statuses of the campaign and if the income condition is correct.  Should I set for valid currency types BGN and EUR?

src/components/campaigns/grid/EditForm.tsx
```diff
@@ -64,15 +66,42 @@ const validationSchema: yup.SchemaOf<Omit<CampaignEditFormData, 'campaignFiles'>
...
+
+  const donationsAmount = useMemo(() => {
+    return campaign.vaults.reduce((acc, vault) => {
+      return acc + vault.amount
+    }, 0)
+  }, [campaign])
+
+  const IsCurrencySelectDisabled =
+    statesForDisableCurrencySelect.includes(campaign.state) ||
+    (campaign.state === CampaignState.active &&
```
igoychev, line 102: please move the CampaignState.active to the statesForDisableCurrencySelect too, so this condition to become simpler.

igoychev (changes requested):
Awesome work!

src/components/campaigns/grid/EditForm.tsx
nradkova: Income value > 0 is a second and compulsory condition only for  `active` state:
```
const IsCurrencySelectDisabled =
    statesForDisableCurrencySelect.includes(campaign.state) && (incomingTransfersAmount > 0 || donationsAmount > 0)
```
`active` && income = 0  => false /correct/
`active` && income = 100 => true /correct/
`rejected`/`error` && income = 0 (it is supposed their income to be 0) => false /incorrect/

I am not sure how to optimize the function.

igoychev (approved):

PR #856, opened 2022-07-05
844 feature button for sharing a campaign needs to be enabled

borislavstoychev:
## Screenshots:

| NONE | image |

## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:

## V1-when button share is clicked the dialog screen with socials media  is displayed
image
## V2-when button share is clicked the social media icons is shown under the  button
image

### DIFF
--- a/src/components/campaigns/InlineDonation.tsx
+++ b/src/components/campaigns/InlineDonation.tsx
@@ -2,7 +2,7 @@ import { useMemo, useState } from 'react'
 import { styled } from '@mui/material/styles'
 import { useTranslation } from 'next-i18next'
 import { CampaignResponse } from 'gql/campaigns'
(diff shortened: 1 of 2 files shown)


igoychev:
Great with suggesting the features that react-share library supports! 
The library looks quite popular, but also strange that it hasn't been updated for ~1 year.  Do we have other options?

borislavstoychev:
> Great with suggesting the features that react-share library supports! The library looks quite popular, but also strange that it hasn't been updated for ~1 year. Do we have other options?

Checkout the last commit. I try something without any library. If you locally test the code keep in mind that some media detect localhost like not valid, but i tested with other urls and all was good.
image
image
image

PR #897, opened 2022-07-18
Show vault available balances and validate balances on vault transactions

borislavstoychev:
## Screenshots:

## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:


src/components/campaigns/grid/CampaignGrid.tsx
```diff
@@ -50,19 +50,21 @@ const DisplayExpandableDescription = (params: GridRenderCellParams<string>) => {
 }
 
 const DisplayReachedAmount = ({ params }: CampaignCellProps) => {
-  const summary = params.row.summary.find(() => true)
```
stann1, line 53: The old implementation was correct, reached amount should be coming from the summary prop

borislavstoychev: https://github.com/podkrepi-bg/api/blob/129ca8fbe42483a242f9463aff67c69b843682e4/apps/api/src/campaign/campaign.service.ts#L468-L505
if i use this function to pin on the summary prop. it will be not possible to display blocked amount, because returns empty vaults list.

stann1:
cannot verify if everything is ok, because when I open campaign details I get an error:
```(console log:) donations.slice is not a function```

borislavstoychev:
> cannot verify if everything is ok, because when I open campaign details I get an error: `(console log:) donations.slice is not a function (src/components/campaigns/DonorsAndDonations.tsx (60:22))`

Yep the response for donations are different that we expect. Please change  227 row ``` <DonorsAndDonations donations={donations?.items} />``` with this. It will help you to test. Don`t worry about ```type script ``` it will work for tests.

imilchev (dismissed):

quantum-grit (approved):

PR #921, opened 2022-07-19
[Feature] beneficiary admin entity

### LINKED ISSUE #896: [Feature] Beneficiary - type company

nradkova:
## Motivation and context
closes #896 

- To be reviewed if there should be an option to create a person in the admin (this will change `create-person.dto` in the api)  or leave it as it was - to select a person only from existing:

Screenshot (204)

Screenshot (203)

#### A possible conflict - currently there is an organizer and its relation assigned to the beneficiary and there is an option to assign an organizer of the campaign created on the behalf of the same beneficiary.
- The organizer has been removed from the beneficiary entity, just the `organizerRelation` property has been left. Both models, as well as the coordinator, are part of the campaign object.


igoychev (changes requested):
Great job with these complex relations for person, company and beneficiary!!

src/components/beneficiary/CreateForm.tsx
```diff
@@ -0,0 +1,180 @@
...
+          )}
+          <Grid item xs={12}>
+            <FormTextField
+              type="text"
+              name="description"
+              autoComplete="target-amount"
+              label={t('beneficiary:grid:description')}
+              multiline
+              rows={1.5}
+            />
```
igoychev, line 166: the autoComplete here seems not correct, please remove as it doesn't make sense for description field
also lets have rows=2

src/components/beneficiary/EditForm.tsx
```diff
@@ -0,0 +1,211 @@
...
+          )}
+          <Grid item xs={12}>
+            <FormTextField
+              type="text"
+              name="description"
+              autoComplete="target-amount"
+              label={t('beneficiary:grid:description')}
+              multiline
+              rows={1.5}
+            />
```
igoychev, line 197: the autoComplete here seems not correct, please remove as it doesn't make sense for description field
also lets have rows=2

igoychev (approved):

PR #940, opened 2022-07-27
Changed thousand separators in BG money format

nradkova:
## Motivation and context
closes #939 

## Screenshots:

- in BG
Screenshot (209)

- in EN
Screenshot (211)

PR #943, opened 2022-07-27
Change number formatting for money

imilchev:
It was requested to change the number formatting for money. Currently, we use `.` for splitting the thousands which becomes confusing. This PR changes the format to use ` ` instead.

Old: `18.000`
New: `18 000`


imilchev:
already handled in #940

PR #945, opened 2022-07-27
935 feature add pagination of all donation requests

borislavstoychev:
## Screenshots:

|image|image |

[Screencast from 27.07.2022 14:53:48.webm](https://user-images.githubusercontent.com/67734870/181264539-a6b5b45b-983c-4da4-8b1d-306b06a7d2a7.webm)
## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:

# ToDo DonationTable pagination 
## Screenshot:
image

- currently we don't have server pagination on  ```donation/user-donations```  endpoint


src/components/campaigns/InlineDonation.tsx
```diff
@@ -120,18 +130,19 @@ export default function InlineDonation({ campaign }: Props) {
...
-    data: { items: donations } = { items: [] },
+    data: { items: donations, total: all_rows } = { items: [] },
     error: donationHistoryError,
     isLoading: isDonationHistoryLoading,
-  } = useCampaignDonationHistory(campaign.id)
+  } = useCampaignDonationHistory(campaign.id, page, 5)
   const { mobile } = useMobile()
   const [isOpen, setIsOpen] = useState(false)
-
+  const rowCount = page * 5 + donations.length
```
imilchev, line 145: it would be good if that `5` is a const because if you change this to a different number here or on line 142, the code will no longer work. It should be obvious that these two need to have the same value

borislavstoychev:
## Design closer to DataGrid pagination
[Screencast from 29.07.2022 10:35:30.webm](https://user-images.githubusercontent.com/67734870/181715019-9724d453-45f1-4094-87a3-5f4643be5379.webm)

PR #950, opened 2022-07-29
Changed success message on bank payment

borislavstoychev:
## Screenshots:
# BG
| image | image |

## Screenshots:
# EN
| image |image |

## Testing

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:

## Please, take a look on EN version too!


imilchev:
Doesn't this change actually change the message for both card payments and bank transfers? The issue says that the message should be different only for bank transfers

imilchev:
nvm I saw it now. All good!

src/components/one-time-donation/steps/Success.tsx
```diff
@@ -19,11 +21,13 @@ export default function Success({ donationId }: { donationId?: string }) {
       <Grid container rowSpacing={2} justifyContent={'center'} textAlign="center">
         <Grid item xs={12}>
           <Typography variant="h4" fontSize={14}>
-            {t('success.title')}
+            {(payment && t('success.title-bank')) || t('success.title')}
```
igoychev, line 24: If you need to distinguish if payment was by bank, please check if you still have access to form data and use the payment type from there paymentField.value === 'bank'

borislavstoychev: > If you need to distinguish if payment was by bank, please check if you still have access to form data and use the payment type from there paymentField.value === 'bank'

Yep, we can access the field 
image
so i will do it on this way....

Issue #971, opened 2022-08-02
[Bug] Bank transfer completion messages are missing


Margi13:
**Describe the bug**
A donor chooses a campaign, clicks on Support, chooses to use bank transfer. When he/she goes to the last step of donation workflow messages are missing, but they are missing only for BG version

I suppose there was a mistake and messages are overwritten in one-time-donation.json file

**To Reproduce**
Steps to reproduce the behavior in bg version:

1. Go to any campaign
2. Click on 'Подкрепи'
3. Choose 'Банков превод'
4. Go all the way in the workflow to the last step
5. See that messages are missing

**Expected behavior**
New messages for bank transfer completion should be shown.

**Screenshots**
![no-text](https://user-images.githubusercontent.com/14825765/182379777-585467bd-4af1-4413-83ff-6ec4516eb28c.PNG)

PR #974, opened 2022-08-02
Add bank transfer completion messages

Margi13:
## Motivation and context
#971 

Currently when a donor chooses a campaign, clicks on Support, chooses to use bank transfer and goes all the way to the last step of donation workflow, messages are missing, but they are missing only for BG version.

Messages are added back to success page in donation workflow for bank transfer.
## Screenshots:

| before| after |


public/locales/bg/one-time-donation.json
```diff
@@ -36,10 +36,13 @@
   },
   "success": {
     "title": "Благодарим за доверието и подкрепата!",
+    "title-bank": "Ще очакваме Вашето дарение!",
     "subtitle": "Вашето дарение ще помогне на кампанията по-бързо да постигне своята цел!",
-    "share-to": "Споделете съобщението на кампанията, за да ѝ помогнете чрез достигане до по-широк кръг дарители",
+    "subtitle-bank": "Благодарим за доверието и запоменте да впишете кода на дарението в основанието на превода си!",
```
imilchev, line 41: There is a typo "запоменте"

kachar (approved):
Well done @Margi13 
Thanks for your contribution!

PR #994, opened 2022-08-08
Add automate UI tests

imilchev:
- [x] Run API db
- [x] Migrate db
- [x] Seed db
- [ ] Run API
- [ ] Mock stripe
- [x] Run frontend
- [x] Run tests

I will also need to check if we can somehow save time by re-using the already built front-end container instead of building it again


imilchev:
Closing in favour of #998

Issue #977, opened 2022-08-03
[Bug] Login with Google during card donation not working


Juicebgs:
**Describe the bug**
Submitted to feedback@podkrepi.bg 

The user is not logged in. Choose a campaign. Click on donate and choose card. 
On step 2 the system prompts user to login. Login with Google button is not working

Both on Chrome Desktop and Android.

igoychev:
Seems like the button does not have onClick defined

PR #995, opened 2022-08-09
fixing google login and campaign counts

quantum-grit:
- fixed: campaign count in profile donation to be from unique slugs, because id is not returned from backend
- fixed: Google login on Second step of donation was missing onClick function - closes https://github.com/podkrepi-bg/frontend/issues/977

### DIFF
--- a/src/components/one-time-donation/LoginForm.tsx
+++ b/src/components/one-time-donation/LoginForm.tsx
@@ -13,7 +13,6 @@ import PasswordField from 'components/common/form/PasswordField'
 
 const onGoogleLogin = () => {
   const resp = signIn('google')
-  console.log(resp)
 }
 
 function LoginForm() {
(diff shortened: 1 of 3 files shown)


src/components/one-time-donation/LoginForm.tsx
```diff
@@ -11,6 +11,11 @@ import { StepsContext } from './helpers/stepperContext'
 import { AlertStore } from 'stores/AlertStore'
 import PasswordField from 'components/common/form/PasswordField'
 
+const onGoogleLogin = () => {
+  const resp = signIn('google')
+  console.log(resp)
```
imilchev, line 16: I guess the `console.log` can go?

quantum-grit: aaah, the copy paste - removed from the original place too... and pls don't ask why we have two login forms... this needs to be fixed separately

igoychev (approved):

PR #996, opened 2022-08-09
Integrate playwright UI tests

kachar:
## Motivation and context

Integrate Playwright end-2-end tests for Podkrepi.bg platform

- https://playwright.dev 
- [VSCode Extension](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright)

## Screenshots:

image

## Testing

```shell
yarn test:e2e
```


e2e/staging/anon-donation-custom.spec.ts
```diff
@@ -0,0 +1,106 @@
+import { test, expect } from '@playwright/test'
```
imilchev, line 1: it would be nice if local and staging are actually exactly the same tests with the only difference being the base URL.

kachar: I've tested against staging until we setup the local environment. Still not convinced we actually need to test against staging in real conditions, but if we do the tests should be the same with only difference the URL as you suggested.

kachar:
Closed in favor of https://github.com/podkrepi-bg/frontend/pull/998

PR #1003, opened 2022-08-19
Disable Beta Popup Dialog

ani-kalpachka:
Disabled Beta Popup Dialog:

image

The component is not deleted in case we need it back. So far its import is commented.
Removed the unused imports in the component.


imilchev:
The Playwright tests are actually testing the pop up as well. Now that it is gone, they fail. Try to fix them. It should be as easy as removing the lines that actually interact with the popup. 👀

e2e/staging/homepage.spec.ts
```diff
@@ -10,19 +10,9 @@ test('test homepage on staging', async ({ page }) => {
...
 
-  // Click text=Това е бета версията на платформата на Подкрепи.бг преди предстоящия наесен офиц
-  await page
-    .locator(
-      'text=Това е бета версията на платформата на Подкрепи.бг преди предстоящия наесен офиц',
-    )
-    .click()
-
   // Check input[type="checkbox"]
   await page.locator('input[type="checkbox"]').check()
```
kachar, line 15: Let's remove this checkbox click as it's part of the beta popup

src/components/index/IndexPage.tsx
```diff
@@ -20,7 +18,7 @@ export default function IndexPage() {
       disableGutters
       title={t('index:title2')}
       metaDescription={t('index:metaDescription')}>
-      <BetaPopUpDialog />
+      {/* <BetaPopUpDialog /> */}
```
kachar, line 21: Let's remove the component, we can always bring it back from git history

src/components/index/sections/BetaPopUpDialog.tsx
```diff
@@ -11,12 +11,8 @@ import {
   FormControlLabel,
   Typography,
 } from '@mui/material'
```
kachar, line 13: Let's remove the component, we can always bring it back from git history

imilchev (approved):

Issue #878, opened 2022-07-13
[Bug] User Session gets broken when any error happens on backend


igoychev:
**Describe the bug**
When the backend responds with error, the user session on frontend is lost as if the user tried to access a page without security token. This results in error 401 unauthorized in the browser console and then the user is redirected to User Profile page where the session is silently restored.

**To Reproduce**

1. Open any frontend page that requires data from backend
2. On the backend modify the code to throw an intentional error
3. Reload the frontend page, so that it makes request to backend
4. See error 401 in the browser console and then redirect to Profile page happens

**Expected behavior**
Backend error should be nicely displayed in error message on frontend without loosing the session

ilievZlatko:
This seems to be an issue with the size of the cookie created by keycloak, or something we additionally add to the cookie.
When debugging I got this error:

<img width="982" alt="Screenshot 2022-07-19 at 13 06 25" src="https://user-images.githubusercontent.com/18635490/179725683-2630b123-e160-4bfd-a9a9-0f053c20c97b.png">

**This should be a back-end task!**

quantum-grit:
fixed by https://github.com/podkrepi-bg/frontend/pull/1007

Issue #914, opened 2022-07-19
[Bug] Personal profile not working after session error 


Juicebgs:
<img width="892" alt="image" src="https://user-images.githubusercontent.com/74968982/179678220-4f23e4d7-601a-4a79-be79-e4f35bb16f23.png">

If there is a session error, the user is unable to look at his personal information (name, donations etc.) although it looks like he/she is logged in. The ideal solution would be to not let a session error unlog the user. 

Second best option is to check when a donor clicks on his/her profile if the user is logged and the session is intact.

Juicebgs:
The Issue #878 is connected to this one.

quantum-grit:
Fixed with PR: https://github.com/podkrepi-bg/frontend/pull/1007

PR #1007, opened 2022-08-30
Fixing broken user session

quantum-grit:
## Motivation and context
User session was breaking during active usage of the site and the user was redirected to profile page.
The problem was in that auth token was expiring before the refresh to run, and the error 401 was thrown and the handler was redirecting the user to login and subsequently to profile page.
Fixed by reducing the expiration interval, removing the 401 error handler and using the getServerSession in getServerProps to reduce the 401 errors. 

Fixes: 
https://github.com/podkrepi-bg/frontend/issues/878
https://github.com/podkrepi-bg/frontend/issues/914


kachar:
@quantum-grit Nice lookin' PR

We might update the `next-auth` dependency to the latest as we use the `unstable_` fn now.  
NextAuth docs describe their latest version and I think we're still using their previous suggested best practices.

quantum-grit:
> We might update the `next-auth` dependency to the latest as we use the `unstable_` fn now. NextAuth docs describe their latest version and I think we're still using their previous suggested best practices.

I just checked and it appears we're already using the latest version of next-auth at  v4.10.3 -> https://www.npmjs.com/package/next-auth

kachar (approved):

PR #1073, opened 2022-10-24
Removed the delete functionality from my campaigns tab

marinovl7:
## Motivation and context

The change is required because the normal user shouldn't be able to delete their campaigns in my campaigns tab in the user's profile

I fixed the GridActions.tsx component in components/campaigns to accept one additional property via props - allowDelete, because this component is also used in *podkrepi.bg/admin/campaigns* and if the property is set to false the delete icon is not displayed and the opposite if the property is set to true. I also cleaned up the MyCampaignsTable component to not have additional unused code. I saw that there was Root property with styles, which was unused and i just deleted it, if this has some reason to be there just comment and i will fix it .

### DIFF
--- a/src/components/campaigns/grid/GridActions.tsx
+++ b/src/components/campaigns/grid/GridActions.tsx
@@ -9,11 +9,12 @@ import { routes } from 'common/routes'
 
 type Props = {
   id: string
+  allowDelete: boolean
   onView: () => void
   onDelete: () => void
 }
 
-export default function GridActions({ id, onView, onDelete }: Props) {
+export default function GridActions({ id, allowDelete, onView, onDelete }: Props) {
   return (
     <Box
       style={{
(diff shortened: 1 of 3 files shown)


igoychev (approved):

ani-kalpachka (approved):

src/components/campaigns/grid/GridActions.tsx
```diff
@@ -9,11 +9,12 @@ import { routes } from 'common/routes'
 
 type Props = {
   id: string
+  allowDelete: boolean
   onView: () => void
   onDelete: () => void
 }
 
-export default function GridActions({ id, onView, onDelete }: Props) {
+export default function GridActions({ id, allowDelete, onView, onDelete }: Props) {
```
kachar, line 17: @marinovl7 You can add a default value for `allowDelete` to avoid passing it to grids explicitly.

```tsx
type Props = {
  id: string
  allowDelete?: boolean
  onView: () => void
  onDelete: () => void
}

export default function GridActions({ id, allowDelete = false, onView, onDelete }: Props)
```

PR #1075, opened 2022-10-26
Hotjar tracking

atomov93:
I added Google optimize tracker and a tracker for Hotjar. Hotjar added a dependency that was needed for easier access as per instructions. Providing screenshot because the page is account specific:
https://snipboard.io/yjLa6G.jpg

As for the google snippet i parsed it as VSCode was having issues with the <script> tag so now it should work. I named wrongfully the second commit hotjar aswell but its for google optimize

**New dependencies**:

package.json
    "@hotjar/browser": "^1.0.6",

yarn.lock
"@hotjar/browser@^1.0.6":
  version "1.0.6"
  resolved "https://registry.yarnpkg.com/@hotjar/browser/-/browser-1.0.6.tgz#5e1c9a2b0c7c21d65acb741b76c39ed81eaa4e5f"
  integrity sha512-yq8pFrCk/pGe1LeI9kgd4NZNI3C8eQlck+ISmWnL3T8RtN/YYEmzDFEZ1AFZ9GGAbxyxG6xUdaL3/S7sOSqeCQ==

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "@emotion/react": "^11.8.1",
     "@emotion/server": "^11.4.0",
     "@emotion/styled": "^11.8.1",
+    "@hotjar/browser": "^1.0.6",
     "@mui/icons-material": "^5.4.2",
     "@mui/lab": "^5.0.0-alpha.69",
     "@mui/material": "^5.4.2",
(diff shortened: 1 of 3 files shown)


kachar:
@atomov93 Installation of HotJar can and should better be done via Google Tag Manager (GTM)

If we use GTM it will be easy and convenient to turn it on or off whenever we want.

atomov93:
I will do that

PR #1076, opened 2022-10-28
Added default value for allowDelete in GridActions

marinovl7:
I added a default value for allowDelete in the GridActions, so the allowDelete propery is per default set to false and in the admin CampaignGrid it set to true, while in the /profile/my-campaigns it's set to false per default (without setting it explicitly). The change was made as per recommendation of @kachar in #1073


kachar (approved):
Awesome update @marinovl7 
Thanks for the contribution!

ani-kalpachka (approved):

Issue #958, opened 2022-07-30
[Feature] Admin should be able to edit donations details imported from bank transfer


Juicebgs:
**Is your feature request related to a problem? Please describe.**
When someone decides to support a cause / campaign via bank transfer the system asks the user to enter a certain code to bank transfer description. 
This code identifies only the campaign and does not contain information about the donor. 
Hence when information is imported into the system every bank transfer will be identified as anonymous donation. 

**Describe the solution you'd like**
Quickest  workaround - system admins with specific rights should be able to edit bank transfer and change ONLY the user from anonymous to existing customer. In an ideal world this change should be logged for the sake of transperancy. 

**Describe alternatives you've considered**
As a long term solution - the information given by the user through the bank transfer should contain user information. It can be done via adding an additional string #[user identificatior] or email address registered in the system in the bank transfer description. 
This should lead to a change in  bank transfer import script and donation message for bank transfer. 

**Additional context**

stann1:
I am adding the **Acceptance criteria** for this story:

1. As an admin, I can open the list of donations.
2. I can click on a donation entry/line in the box for "Donor", which would allow me to edit the donor. 2 possibilities: either inline editing in the grid, or a small prompt box. 
3. The edit box should have a search input, where I can enter a text and it performs a partial matching in the following props: **username, firstname, lastname**. 
4. From the search result, I select the desired person and click save.
5. On success - the grid entry should be updated automatically,  or the whole grid should reload the show the latest changes

igoychev:
fixed

PR #1097, opened 2022-11-10
Admin can edit donations details imported from bank transfer

Margi13:
## Motivation and context
#958 

Currently when information for donation is imported into the system every bank transfer will be identified as anonymous donation.
System admins should be able to edit bank transfer and change the user from anonymous to existing customer. 

## Screenshots:
podkrepi1

**In edit mode**
podkrepi2

**Admin can edit only bank transfer donations:**
podkrepi3


ani-kalpachka (approved):

igoychev (approved):
great for adding this feature!

PR #1098, opened 2022-11-13
Show all campaigns on Campaigns listing page

ani-kalpachka:


imilchev (commented):
What is the goal here? Are we completely getting rid of paging? Isn’t that going to be a problem when we get 20 or more campaigns?

PR #1105, opened 2022-11-16
[Feat] Added See More/Less about page

ivoka9:
The Management Board had too long of a description, So a button was added to see more/less

Screenshot 2022-11-15 221811
Screenshot 2022-11-15 221650

### DIFF
--- a/src/components/about/sections/ManagementBoardSection/ManagementBoardDescription.tsx
+++ b/src/components/about/sections/ManagementBoardSection/ManagementBoardDescription.tsx
@@ -1,0 +1,23 @@
+import { Button } from '@mui/material'
+import { Description } from 'components/about/AboutPage.styled'
+import { useTranslation } from 'next-i18next'
+import { useState } from 'react'
+
+export function ManagementBoardDescritpion({ description }) {
+  const { t } = useTranslation()
+
+  const [show, setShow] = useState(false)
+
+  return (
+    <>
+      <Description variant="body2" height={show ? 'auto' : '200px'}>
+        {description}
+      </Description>
+      <div margin="0 auto" width="100%">
+        <Button variant="outlined" onClick={() => setShow(!show)}>
+          {show ? t('campaigns:cta.see-less') : t('campaigns:cta.see-more')}
+        </Button>
+      </div>
+    </>
+  )
+}
--- a/src/components/about/AboutPage.styled.tsx
+++ b/src/components/about/AboutPage.styled.tsx
@@ -48,6 +48,8 @@ export const LinkedInText = styled(Typography)(() => ({
 
 export const Description = styled(Typography)(() => ({
   textAlign: 'initial',
+  marginBottom: theme.spacing(3),
+  overflow: 'hidden',
 }))
 
 export const AboutWrapper = styled(Grid)(() => ({
--- a/src/components/about/sections/ManagementBoardSection/ManagementBoardSection.tsx
+++ b/src/components/about/sections/ManagementBoardSection/ManagementBoardSection.tsx
@@ -17,6 +17,7 @@ import {
   LinkedInIcon,
   LinkedInText,
 } from 'components/about/AboutPage.styled'
+import { ManagementBoardDescritpion } from './ManagementBoardDescription'
 
 export default function ManagementBoardSection() {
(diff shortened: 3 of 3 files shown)


src/components/about/sections/ManagementBoardSection/ManagementBoardDescription.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <>
+      <Description variant="body2" height={show ? 'auto' : '200px'}>
+        {description}
+      </Description>
+      <div margin="0 auto" width="100%">
+        <Button variant="outlined" onClick={() => setShow(!show)}>
+          {show ? t('campaigns:cta.see-less') : t('campaigns:cta.see-more')}
+        </Button>
+      </div>
```
dimitur2204, line 20: You probably shouldn't be using a `<div>` and styling it directly. Also I don't exactly get why you need the `margin: 0 auto;` when you are have a `width: 100%`.

If the whole idea of that was to make the button full width there is a property on the MUI Button called `fullWidth` that you can toggle https://mui.com/material-ui/api/button/

src/components/about/sections/ManagementBoardSection/ManagementBoardDescription.tsx
```diff
@@ -0,0 +1,23 @@
...
+export function ManagementBoardDescritpion({ description }) {
+  const { t } = useTranslation()
+
+  const [show, setShow] = useState(false)
+
+  return (
+    <>
+      <Description variant="body2" height={show ? 'auto' : '200px'}>
+        {description}
+      </Description>
```
dimitur2204, line 15: Cutting off elements like that is not particularly good. You might cut off text in the middle of it's y axis and it will look bad. 
You should consider a different approach truncating the text to ex. 200 symbols and then adding .... at the end!

dimitur2204 (changes requested):

src/components/about/sections/ManagementBoardSection/ManagementBoardDescription.tsx
ivoka9: Thanks I think you are right, I was thinking that there must be a better way and this seems way better

kabaivanov: > Cutting off elements like that is not particularly good. You might cut off text in the middle of it's y axis and it will look bad.
> You should consider a different approach truncating the text to ex. 200 symbols and then adding .... at the end!
> Here is a lodash function you can use https://lodash.com/docs/4.17.15#truncate

Maybe try this (CSS only fix)
/sorry can't really implement this myself yet as I still have to learn the tech behind, but I can write the CSS hack for someone that can do it accordingly/

Screenshot 2022-11-17 at 4 08 07

*_There is a second issue of Member names not being of equal height (and vertically centered), thus shifting the paragraphs below when the Member names consist of more than one lines, which is also being fixed here_.

/add proposed styles to existing/

// fix titles
`h6.MuiTypography-subtitle1 {
  height: 3em;
  display: flex;
  flex-direction: column;
  justify-content: center;
}`

// truncate to 14 lines of text
`p.MuiTypography-body2 {
  max-height: calc(1em*14*1.43); 
  overflow: hidden;
  position: relative;
}`

// fade out the last 4 lines instead of adding an ellipsis (...)
`p.MuiTypography-body2:after {
  content: "";
  position: absolute;
  bottom: 0px;
  left: 0px;
  width: 100%;
  height: calc(0.875rem*1.43*4);
  background: linear-gradient(0deg, rgb(255, 255, 255) 0%, rgba(255,255,255,.33) 65%, rgba(255,255,255,0) 100%);
  z-index: 1;
}`

** Add a **_Learn more_** button below.

src/components/about/sections/ManagementBoardSection/ManagementBoardDescription.tsx
kachar: Keep in mind that `margin` is not really a standard HTML attribute, and here it's used as CSS style.

Probably the whole div wrapper should not be there.

ani-kalpachka:
The fix is provided in https://github.com/podkrepi-bg/frontend/pull/1139
Closing as duplicate.

PR #1114, opened 2022-11-17
Made my profile tabs scrollable in mobile view

marinovl7:
## Motivation and context

There was a problem, that I saw from a message in discrod, that when the user in in https://podkrepi.bg/profile/[slug] in mobile view, not all tabs are visible. So basically the user can see only 3 of the 4 tabs, and there is no way that the user can see the fourth tab - the tabs are neither scrollable, nor they provide a button that can act like horizontal scrollbar.

I also changed the icon of my-campaigns section because they were kinda the same with the edit personal information icon( see the image below) 
I don't know if this is the perfect icon and I'm open to suggestions.

Screenshot (80)

I followed the MUI documentation and implemented the Forced scroll buttons from there, because if you add just scroll buttons, they are available only in desktop mode, but not in mobile and that's not what we want. 
There was another way to fix this, so I basically had to make the tabs scrollable, but I think that this way the user won't even understand that there is fourth tab ( if he doesn't see the chevron icon, that indicates that the user can scroll ). So now the user can see the fourth tab with click on the chevron and scrolling can also be initiated through user agent scrolling mechanisms (e.g. left/right swipe, shift mouse wheel, etc.)

##Screenshots

THE PROBLEM ( 4th tab not visible and not reachable ) 

Screenshot_20221117-083239_Chrome

THE SOLUTION

Screenshot (82)
Screenshot (83)

If you wonder why is there to much place on the left and right, when the chevrons are not visible, the reason is that they don't dissapear from the dom and stay always there.

### DIFF
--- a/src/components/auth/profile/ProfilePage.tsx
+++ b/src/components/auth/profile/ProfilePage.tsx
@@ -9,7 +9,7 @@ import {
   VolunteerActivism as DonationIcon,
   AccountBox as AccountBoxIcon,
   Assignment as CertificateIcon,
-  AssignmentInd as CampaignIcon,
+  AccountBalance as CampaignIcon,
 } from '@mui/icons-material'
 import { useSession } from 'next-auth/react'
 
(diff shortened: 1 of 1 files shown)


src/components/auth/profile/ProfilePage.tsx
```diff
@@ -77,12 +77,14 @@ export default function ProfilePage() {
           sx={{
             backgroundColor: 'white',
             borderRadius: '25px 25px 0px 0px',
-            padding: '10px 30px',
+            padding: '10px 10px',
```
ani-kalpachka, line 80: We could use only padding: '10px'. It will apply 10px for each side of the box.

ani-kalpachka (approved):

ani-kalpachka:
Well done! Thanks for implementing this!

Issue #1095, opened 2022-11-09
Update the date format of 'Change birthdate' dialog


ani-kalpachka:
Steps to reproduce:
1. Login
2. Open 'Profile'
3. Open 'Personal Information' tab.
4. Scroll down to 'Birthdate'. The date format is 'dd/MM/yyyy':
![image](https://user-images.githubusercontent.com/14351733/200903741-eb90000c-7bfd-4d08-89e8-b473c6ad45a9.png)
5. Click on 'Edit'. On the 'Change birthdate' dialog the date format is 'dd/MM/yyyy':
![image](https://user-images.githubusercontent.com/14351733/200904233-5313e9a9-8ebe-472f-bd40-5873206b306d.png)

Expected behavior: 
Unify the date format to be 'dd/MM/yyyy'.

batebobo:
Hi @ani-kalpachka! The date format on the first screenshot is `dd.MM.yyyy` and on the second is `MM/dd/yyyy`. Can you please confirm that we prefer the latter?

Anyways I'd like to take this task. Can you please assign me, since I can't see a way to assign myself :)

ani-kalpachka:
@batebobo I agree. We need dd.MM.yyyy format. 
The task is assigned to you. Thank you!

batebobo:
Hi @ani-kalpachka! Can you please send some details about your browser (locale, language settings, etc.) :pray: I'm not able to reproduce the issue - I see the correct format in the dialog - `dd.MM.yyyy`.

The dialog uses a native `input type="date"` element that respects users' locale settings and its format can't be (easily) changed. On the other hand, the date in the _Personal Info_ tab is not localized - it's always `dd.MM.yyyy` and that's not okay for most locales.

ovcharski:
I am reproducing the issue. I am attaching a .gif 
I am using Firefox on a Windows 11 laptop English US https://whatismybrowser.com/w/DGYC4LC

![edit-birthday](https://user-images.githubusercontent.com/48056945/202130942-c09d08b6-8d11-48b3-a7ed-218deaa78b30.gif)

PR #1115, opened 2022-11-17
Consistent date format in update birthday modal

batebobo:
## Motivation and context
Implements #1095 

The order of day/month between locales can be handled with localization, but the presentation differs by browser
e.g. in Chrome we have a `.` separator, while in Firefox we have `/` for the same locale. This has been worked around with a locale dictionary that maps the user lanugage to a specific date presentation format. Current formats include only `bg` and `en-US`, but others can be added as well.

The _update birthday_ modal now uses a MUI datepicker, as the presentation of the native `input type="date"` is not configurable. The mentioned datepicker lives in a separate convenience component in the `common/form` folder.

## Screenshots:

Before: (https://user-images.githubusercontent.com/48056945/202130942-c09d08b6-8d11-48b3-a7ed-218deaa78b30.gif)
After: 
Screenshot 2022-11-17 at 14 06 12

## Testing

For testing, users should change the browser language and locale (for best chances of the browser respecting their want of a different locale).

In chrome, this happens via `Settings -> Language` and then moving your desired language to the top of the list of preferred languages.

In firefox this happens via `Preferences` under `Language`. The whole browser language needs to be changed in order for the localization to take place.

For Bulgarian, we should expect a `day.month.year.` format, whileas for US we should expect `month/day/year/` format.

### DIFF
--- a/src/common/util/date.ts
+++ b/src/common/util/date.ts
@@ -10,18 +10,32 @@ export const dateFormatter = (value: Date | string | number, locale?: Locale) =>
   return `${exact} (${relative})`
 }
 
-export const formatDateString = (dateString: string | Date) => {
-  const date = new Date(dateString)
-  const day = date.getDate().toString().padStart(2, '0')
-  const month = (date.getMonth() + 1).toString().padStart(2, '0')
-  const year = date.getFullYear()
-
-  return `${day}.${month}.${year}`
+export const formatDateString = (dateString: string | Date, locale?: string | undefined) => {
+  return format(new Date(dateString), getDateFormat(locale ?? ''))
 }
 
 export const getRelativeDate = (value: Date | string, locale?: Locale) => {
   const date = new Date(value)
   return formatRelative(date, new Date(), { locale })
+}
+
+/**
+ * Value format for storing and passing date strings. Should be used whenever we pass date strings between components.
+ */
+export const DATE_VALUE_FORMAT = 'yyyy-MM-dd'
+
+/**
+ * Utility to return localized date format string. Should be used whenever dates are displayed to end users.
+ * @param language Language to get the corresponding date format
+ * @returns Format string
+ */
+export const getDateFormat = (language: string) => {
+  const defaultFormat = 'dd.MM.yyyy'
+  const languageDateFormats: { [language: string]: string } = {
+    bg: defaultFormat,
+    'en-US': 'MM/dd/yyyy',
+  }
+  return languageDateFormats[language] ?? defaultFormat
 }
 
 /**
--- a/src/common/util/date.ts
+++ b/src/common/util/date.ts
@@ -11,7 +11,10 @@ export const dateFormatter = (value: Date | string | number, locale?: Locale) =>
 }
 
 export const formatDateString = (dateString: string | Date, locale?: string | undefined) => {
-  return format(new Date(dateString), getDateFormat(locale ?? ''))
+  if (locale) {
+    return Intl.DateTimeFormat(locale.split('-')).format(new Date(dateString))
+  }
+  return new Date(dateString).toLocaleDateString()
 }
 
 export const getRelativeDate = (value: Date | string, locale?: Locale) => {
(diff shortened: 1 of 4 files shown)


src/components/common/form/FormDatePicker.tsx
```diff
@@ -0,0 +1,46 @@
+import { TextField } from '@mui/material'
+import { LocalizationProvider, DatePicker } from '@mui/x-date-pickers'
```
dimitur2204, line 2: I think they had an idea to monetize `@mui/x` components and this might become paid in the future
https://mui.com/x/introduction/
From what I am seeing from the documentation though it seems to be free!

src/components/common/form/FormDatePicker.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <LocalizationProvider dateAdapter={AdapterDateFns}>
+      <DatePicker
+        mask={mask}
+        inputFormat={dateViewFormat}
+        label={label}
+        value={field.value}
+        onChange={(newValue) => updateValue(newValue)}
+        renderInput={(params) => <TextField size="small" {...params} />}
```
dimitur2204, line 42: Maybe add a `prop` to change the size of the input rendered, so that it's not always `size=small`

dimitur2204 (approved):
Seems good to me generally! Good job 😃!
Wait on more reviewers though.

src/components/common/form/FormDatePicker.tsx
kachar: @dimitur2204 I guess if it becomes paid, then we'll stick to the free npm version or we will replace it with another fish in the sea of date picker libraries. :fishing_pole_and_fish:

src/common/util/date.ts
```diff
@@ -10,20 +10,34 @@ export const dateFormatter = (value: Date | string | number, locale?: Locale) =>
...
+ * @returns Format string
+ */
+export const getDateFormat = (language: string) => {
+  const defaultFormat = 'dd.MM.yyyy'
+  const languageDateFormats: { [language: string]: string } = {
+    bg: defaultFormat,
+    'en-US': 'MM/dd/yyyy',
+  }
+  return languageDateFormats[language] ?? defaultFormat
+}
```
kachar, line 39: @batebobo We might also use the standard js libs to accomplish this functionality:

```
new Intl.DateTimeFormat(['bg', 'bg']).format(date)
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

src/components/common/form/FormDatePicker.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * MUI date picker to be connected with Formik. Propagates updates to the passed Formik field name
+ * @param name - name of the Formik field to bind
+ * @param label - prompt text
+ * @returns
+ */
+export default function FormDatePicker({ name, label }: { name: string; label: string }) {
+  const [field] = useField(name)
+  const { setFieldValue } = useFormikContext()
+
+  const dateViewFormat = getDateFormat(i18n?.language ?? '')
```
kachar, line 20: It's best to use the `i18n` object from the `useTranslation()` hook so the component is re-rendered once the language changes.

```tsx
const { i18n } = useTranslation()
```

src/components/common/form/FormDatePicker.tsx
batebobo: Thanks @dimitur2204 for the insight, I had no idea 😄 I agree with @kachar to leave it as it is for now and act accordingly if it becomes paid

src/components/common/form/FormDatePicker.tsx
batebobo: I'd leave such customization for whenever needed. I've only parametrized the needed parts as part of this task. What do you think?

src/common/util/date.ts
batebobo: Thank you! I spent a while looking for generic solutions for proper formatting, but somehow managed to miss that :smile: I'll change it and let you know :)

batebobo: Hi again! I removed the hardcoded date formats from the function in favor of a more generic solution using `formatToParts` and extracting the relevant fields. The reason I need the format and not the formatted date string is because the MUI datepicker expects a format string. The reason I need a format string is that the `date-fns` localization adapter produces `dd/MM/yyyy` for `bg` locale 😕  Let me know how this looks 🙏

kachar (approved):
@batebobo Nice contribution! Thanks!

PR #1120, opened 2022-11-18
Fix missing autocompletion when entering a new password

antonmihaylov:
Hello and admiration for the project!
This is a fix for a tiny issue that I noticed while signing up for the platform.

## Motivation and context

Password managers or some browsers' built-in password features can generate and save passwords for you when they detect you're entering a new password - e.g. when you're registering, changing your password, etc.

They use the attribute `autocomplete` to do that. When it's set to `current-password`, they suggest existing stored passwords for you. When it's set to `new-password` they suggest generating and saving a new password.

[Mozilla docs on the subject](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)

In this project currently, all authentication forms use a default `PasswordField` component, which sets autocomplete to `current-password`.
This PR allows to override this default prop and overrides it to `new-password` in the appropriate places.

## Screenshots:
Here's a preview using 1Password

| [Podkrepi BG signup form before.webm](https://user-images.githubusercontent.com/57956282/202743017-1de3b2a7-936b-467b-84e1-6df364b1fcf3.webm) | [Podkrepi BG signup form after.webm](https://user-images.githubusercontent.com/57956282/202743070-4d4ca0d1-8691-40e4-b43d-15cfcf30c5e3.webm) |

## Testing

I tested the changed forms manually.

### DIFF
--- a/src/components/auth/profile/UpdatePasswordModal.tsx
+++ b/src/components/auth/profile/UpdatePasswordModal.tsx
@@ -132,10 +132,18 @@ function UpdatePasswordModal({
               <PasswordField name={'previous-password'} label={'auth:account.previous-password'} />
             </Grid>
             <Grid item xs={12} sm={8}>
-              <PasswordField name={'password'} label={'auth:account.new-password'} />
+              <PasswordField
+                name={'password'}
+                label={'auth:account.new-password'}
+                autoComplete="new-password"
+              />
             </Grid>
             <Grid item xs={12} sm={8}>
-              <PasswordField name={'confirm-password'} label={'auth:account.confirm-password'} />
+              <PasswordField
+                name={'confirm-password'}
+                label={'auth:account.confirm-password'}
+                autoComplete="new-password"
+              />
             </Grid>
             <Grid item xs={6}>
               <SubmitButton fullWidth label="auth:cta.send" loading={loading} />
--- a/src/components/auth/profile/UpdatePasswordModal.tsx
+++ b/src/components/auth/profile/UpdatePasswordModal.tsx
@@ -140,8 +140,8 @@ function UpdatePasswordModal({
             </Grid>
             <Grid item xs={12} sm={8}>
               <PasswordField
-                name={'confirm-password'}
-                label={'auth:account.confirm-password'}
+                name="confirm-password"
+                label="auth:account.confirm-password"
                 autoComplete="new-password"
               />
             </Grid>
--- a/src/components/auth/profile/UpdatePasswordModal.tsx
+++ b/src/components/auth/profile/UpdatePasswordModal.tsx
@@ -133,8 +133,8 @@ function UpdatePasswordModal({
             </Grid>
             <Grid item xs={12} sm={8}>
               <PasswordField
-                name={'password'}
-                label={'auth:account.new-password'}
+                name="password"
+                label="auth:account.new-password"
                 autoComplete="new-password"
               />
             </Grid>
--- a/src/components/auth/register/RegisterForm.tsx
+++ b/src/components/auth/register/RegisterForm.tsx
@@ -101,10 +101,14 @@ export default function RegisterForm({ initialValues = defaults }: RegisterFormP
           <FormTextField type="text" label="auth:fields.email" name="email" autoComplete="email" />
         </Grid>
         <Grid item xs={12}>
-          <PasswordField />
+          <PasswordField autoComplete="new-password" />
         </Grid>
         <Grid item xs={12}>
-          <PasswordField name={'confirm-password'} label={'auth:account.confirm-password'} />
+          <PasswordField
+            name={'confirm-password'}
+            label={'auth:account.confirm-password'}
+            autoComplete="new-password"
+          />
         </Grid>
         <Grid item xs={12}>
           <AcceptTermsField name="terms" />
--- a/src/components/auth/register/RegisterForm.tsx
+++ b/src/components/auth/register/RegisterForm.tsx
@@ -105,8 +105,8 @@ export default function RegisterForm({ initialValues = defaults }: RegisterFormP
         </Grid>
         <Grid item xs={12}>
           <PasswordField
-            name={'confirm-password'}
-            label={'auth:account.confirm-password'}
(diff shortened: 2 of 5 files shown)


src/components/auth/profile/UpdatePasswordModal.tsx
```diff
@@ -132,10 +132,18 @@ function UpdatePasswordModal({
               <PasswordField name={'previous-password'} label={'auth:account.previous-password'} />
             </Grid>
             <Grid item xs={12} sm={8}>
-              <PasswordField name={'password'} label={'auth:account.new-password'} />
+              <PasswordField
+                name={'password'}
+                label={'auth:account.new-password'}
+                autoComplete="new-password"
+              />
```
kachar, line 139:
```suggestion
              <PasswordField
                name="password"
                label="auth:account.new-password"
                autoComplete="new-password"
              />
```

src/components/auth/profile/UpdatePasswordModal.tsx
```diff
@@ -132,10 +132,18 @@ function UpdatePasswordModal({
...
             </Grid>
             <Grid item xs={12} sm={8}>
-              <PasswordField name={'confirm-password'} label={'auth:account.confirm-password'} />
+              <PasswordField
+                name={'confirm-password'}
+                label={'auth:account.confirm-password'}
+                autoComplete="new-password"
+              />
```
kachar, line 146:
```suggestion
              <PasswordField
                name="confirm-password"
                label="auth:account.confirm-password"
                autoComplete="new-password"
              />
```

src/components/auth/register/RegisterForm.tsx
```diff
@@ -101,10 +101,14 @@ export default function RegisterForm({ initialValues = defaults }: RegisterFormP
...
-          <PasswordField />
+          <PasswordField autoComplete="new-password" />
         </Grid>
         <Grid item xs={12}>
-          <PasswordField name={'confirm-password'} label={'auth:account.confirm-password'} />
+          <PasswordField
+            name={'confirm-password'}
+            label={'auth:account.confirm-password'}
+            autoComplete="new-password"
+          />
```
kachar, line 111:
```suggestion
          <PasswordField
            name="confirm-password"
            label="auth:account.confirm-password"
            autoComplete="new-password"
          />
```

kachar (commented):
@antonmihaylov Nice update! Thanks for the idea & execution.

I would suggest that we add the new `autoComplete` updates also to the two `<LoginForm />` components:

- https://github.com/podkrepi-bg/frontend/blob/31fbc63968462a30bd50455c71ed57ec453d2c58/src/components/auth/login/LoginForm.tsx
- https://github.com/podkrepi-bg/frontend/blob/2a87ab8fe4d57a9fa6d5438136a65c73967d2a25/src/components/one-time-donation/LoginForm.tsx

valkirilov:
I've noticed something similar, but related with the Login page. My Password Manager (LastPass) can't auto-fill my email/password as it usually do in the other websites. For some reason it doesn't identify these fields at all. 

Is there any referenced issue already logged or you went directly to fix it within this PR?

antonmihaylov:
@kachar 
It doesn't make much sense to update the login forms, since they appropriately have their autocomplete attributes set to `current-password`, which is set by the `PasswordField` component by default if you don't override it. So there's no benefit except being more explicit.

antonmihaylov:
@valkirilov 
1Password fills in my details in the login form as usual. Tested on Firefox's built-in password manager - it also works. Are you sure it's not a LastPass configuration issue?

kachar:
@antonmihaylov I was referring to the `email` field on the `LoginForm`s, it lacks the `autoComplete` attribute in both places. I guess the browsers are smart enough to pick it up from attributes like `name="email"` and `label="Email"`.

We may also extract it as a separate component `EmailField`, just like `PasswordField`, and reuse it.

I'm okay we don't do it, just sharing ideas here.

antonmihaylov:
> @antonmihaylov I was referring to the `email` field on the `LoginForm`s, it lacks the `autoComplete` attribute in both places. I guess the browsers are smart enough to pick it up from attributes like `name="email"` and `label="Email"`.
> 
> 
> 
> We may also extract it as a separate component `EmailField`, just like `PasswordField`, and reuse it.
> 
> 
> 
> I'm okay we don't do it, just sharing ideas here.

Ah didn't think about the email fields at all, since they indeed seem to get picked up implicitly.
But yeah, makes sense to add them too. 

I now notice that they have type=text. Is this on purpose to avoid the default validation? 

While we're at it, I suggest adding type="email" and, if it's required to avoid the browser's default validation - to add novalidate. This would improve mobile experience a bit since you'd get a [specific keyboard layout for entering emails.](https://css-tricks.com/better-form-inputs-for-better-mobile-user-experiences/)

kachar:
@antonmihaylov Let's await the update and then merge this PR.
If you prefer otherwise please let me know so we can split the login issues as a separate task.

dimitur2204:
> > @antonmihaylov I was referring to the `email` field on the `LoginForm`s, it lacks the `autoComplete` attribute in both places. I guess the browsers are smart enough to pick it up from attributes like `name="email"` and `label="Email"`.
> > We may also extract it as a separate component `EmailField`, just like `PasswordField`, and reuse it.
> > I'm okay we don't do it, just sharing ideas here.
> 
> Ah didn't think about the email fields at all, since they indeed seem to get picked up implicitly. But yeah, makes sense to add them too.
> 
> I now notice that they have type=text. Is this on purpose to avoid the default validation?
> 
> While we're at it, I suggest adding type="email" and, if it's required to avoid the browser's default validation - to add novalidate. This would improve mobile experience a bit since you'd get a [specific keyboard layout for entering emails.](https://css-tricks.com/better-form-inputs-for-better-mobile-user-experiences/)

I'm personally using NordPass and the email field works for me. But generally it would be an amazing addition!

kachar (approved):
@antonmihaylov Thanks for the nice contribution! Nice work on the auth components :bowing_man: 

Let's postpone the login form update in a separate issue so this one is not blocked.

Issue #1016, opened 2022-09-04
[Bug] Dropdown menu is covering the "Donate" button


atomov93:
**Describe the bug**
We noticed that opening the Campaigns Menu it covers slightly the Donate button. See screenshots below. The issue is observed for Bulgarian language(BG), but is also valid for English if words were longer.

The solution could be to just move the dropdown 2 pixels downwards.

**To Reproduce**

1. Go to 'dev.podkrepi.bg'
2. Click on 'Кампании'

**Expected behavior**
Dropdowns to not cover Menus

**Screenshots**
<img width="315" alt="image" src="https://user-images.githubusercontent.com/62711773/188319143-9f875bba-8162-49cf-b8c0-e92248314838.png">

**Desktop (please complete the following information):**

- OS: MacOs Monterey 15.1.2
- Browser Opera, Safari, Chrome

PR #1121, opened 2022-11-18
Fix visual issue with Menu items in the MainNav

valkirilov:
## Motivation and context
It was noticed that when you open the "**Campaigns**" menu in the main navigation, the dropdown covers slightly the "**Donate**" button (see screenshots below). The issue is observed for all languages and also applies for the other menu items (like the "**About us**") if they consist of options with unusually long wording.

References: #1016

### How did you solve the problem?
Increase the spacing between the Button ("**Campaigns**") and the opened dropdown with the menu items so they don't interfere with the border of the "**Donate**" button (in the `MainNav`). The changes are applied globally to the `GenericMenu` component which serves as a wrapper to all menu items in the top navigation, because the bug is valid not only for the "**Campaigns**" item, but for all of them (although we don't have so long words in the "**About us**" menu and we can't reproduce it naturally).

## Screenshots:

| image| image |

## Testing

To test this change simply go to any page and click on the "**Campaign**" item in the main navigation. It will open a dropdown with options and you can verify whether it interferes with the border of the "**Donate**" button.

The changes introduced in this PR impact all items in the main navigation (i.e. all elements which use the `GenericMenu` component, like "**Campaigns**", "**About us**", etc.)


kachar (changes requested):
@valkirilov I think we can achieve the same result with update to the menu instance instead of the generic component.

My main concern with adjusting the `<GenericMenu />` component is that we'll add margin-top for all instances of the menu, and not only the one in the nav bar.

My suggestion is to add the new styles to the instance `<ProjectMenu />` at 

https://github.com/podkrepi-bg/frontend/blob/db9ae5d46b74b4bce310a2d44d0a5efa7408b26a/src/components/layout/nav/ProjectMenu.tsx#L19-L33

valkirilov:
@kachar 

Yes, initially I wet for putting a margin only on the `DonationMenu` ("**Campaigns**" item in the navbar) as you suggested, but then I realized that the issue is not only for one this item, but for all of menu items in the navbar (although we can visually observe it only for the `DonationMenu` which is the closest one to the "Donation" button). 

image

So I put the change in `GenericMenu` intentionally, to cover all possible cases. If you think that it's too much, I'll revert my changes and apply it only to the menu item which is causing the visual issue ("Campaigns" a.k.a `DonationMenu`).

kachar:
@valkirilov Alright, if that's the  case we can proceed with an update of `GenericMenu` 

Few suggestions though:

- Rename `GenericMenu` to `GenericNavMenu` to clearly distinguish its purpose
- Use `PaperProps` instead of `styled(Menu)` approach to reduce the style overhead

```tsx
<Menu
  keepMounted
  disableScrollLock // <-- no need of `={true}`
  id="menu-${id}" // <-- now all menus are `menu-donation`, not generic
  anchorEl={anchorEl}
  elevation={6}
  onClose={handleClose}
  open={Boolean(anchorEl)}
  PaperProps={{ sx: { mt: 1 }}} // <-- use SX instead of styled
  anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
  transformOrigin={{ vertical: 'top', horizontal: 'right' }}>
  {children}
</Menu>
```

How do you feel about these updates?

valkirilov:
@kachar yes, totally agree with you. I was afraid to do such refactoring with my first task, but since you suggested it, I'll do it :)

valkirilov:
I made some refactoring of the `GenericMenu` component, as requested:
- changed the way we set the spacing between the `Button` and the `Menu` dropdown list to use `PaperProps` instead of wrapping it behind a `styled` component. It's simpler now.
- renamed the `GenericMenu` component to `GenericNavMenu` so it will distinguish its purpose way more clearly
- fixed the `id` attribute to be properly set via an input prop instead of using always the default wrong `menu-donation` for all items of this type

kachar (approved):
@valkirilov Valuable contribution to the project!
Thank you for your cooperative approach to the code improvements. :rocket:

Issue #1123, opened 2022-11-18
[Feature] Upgrade to Next@13 and React@18 


dimitur2204:
**Is your feature request related to a problem? Please describe.**
Upgrade to [next@13](https://beta.nextjs.org/docs/upgrade-guide) and [React 18 ](https://reactjs.org/blog/2022/03/29/react-v18.html)

**Things to keep in mind**
There might be some breaking changes considering these are both major versions on our core dependencies!

PR #1125, opened 2022-11-19
[Improvement] Upgrade dependencies including Next 13 and React 18

dimitur2204:
## Motivation and context

Updating to the new next@13 and React@18 version as described in #1123.

Upgraded following the nextjs upgrade guide: https://beta.nextjs.org/docs/upgrade-guide#upgrading-new-features

There are still improvments to be made and errors might come up.
I haven't done anything more than what was breaking the app. There are still improvements to be made like:
* Font optimization
* Change to the new `app` folder instead of pages 

## Notes

**Auth**:
Didn't manage to get the old next auth callbacks working with the new `middleware.ts`, so now you can access `/admin` as a normal user if you type it in, everything else however is protected
image

**Types**:
After upgrading to React 18 there seemed to be a problem with `@react-pdf/renderer` types that I couldn' t fix so I added 
`declare module '@react-pdf/renderer'` I have opened a PR to them with the fix if that is actually a problem. But it seems like it to me https://github.com/diegomura/react-pdf/pull/2087
## Testing
I have went through the pages  and donation flow. Everything seems to work.
Also checked the Admin platform.

After upgrading to React 18 there seemed to be a problem with `@react-pdf/renderer` types that I couldn't fix so I added  `declare module '@react-pdf/renderer'`
**However I think we should delay releasing this on production until we had some development time on the new versions!**

**New dependencies**:

- `next: 13` 
- `react: 18.2.0`
- `next-auth: 4.16.4`
- `next-i18next: 13.0.0`
- `react-i18next: 12.0.0`
- `i18next: 22.0.6`


dimitur2204:
Tests are passing locally. I am not exactly sure what the problem is

src/middleware.ts
```diff
@@ -0,0 +1,5 @@
+export { default } from 'next-auth/middleware'
+
+export const config = {
+  matcher: ['/admin', '/profile'],
+}
```
kachar, line 5: @dimitur2204 Here's a suggestion for the auth middleware to support the previous functionality - the `isAdmin` check

```suggestion
import { getToken } from 'next-auth/jwt'
import withAuth from 'next-auth/middleware'

export default withAuth({
  pages,
  callbacks: {
    authorized: async ({ req }) => {
        const session = await getToken({ req })
        if (!session) return false
        if (session.accessTokenExpires <= Date.now()) return false

        if (req.nextUrl.pathname.startsWith('/profile')) {
          return true
        }

        if (req.nextUrl.pathname.startsWith('/admin')) {
          return isAdmin(session)
        }
    },
  },
})

export const config = {
  matcher: ['/admin', '/admin/:path*', '/profile', '/profile/:path*'],
}
```

kachar (commented):
@dimitur2204 Incredible upgrade of the app foundations! Can't wait to see it in action! Thanks for your invested time and energy :zap: 


I've added a suggestion on the middleware so the user needs to pass the `isAdmin()` check before they enter the `/admin*` routes. I haven't been able to properly test it, but the approach is clear.

:raised_hands:

src/middleware.ts
dimitur2204: I will test that!

dimitur2204: Good catch it does work! Changed in [1123 - update auth to now protect admin](https://github.com/podkrepi-bg/frontend/pull/1125/commits/7a564396c4f3246008424e766d3a695c526b74ac)

kachar (approved):

PR #1126, opened 2022-11-19
Enable Terms and Conditions and GDPR validation on Register form

ani-kalpachka:
Enable Terms and Conditions and GDPR validation on Register form so the user cannot continue without accepting both of the fields:


kachar (approved):
Swift upgrade :rocket:

PR #1134, opened 2022-11-20
[Bug] You can click Next if you are both not logged and not anonymous

dimitur2204:
## Motivation and context

The anonymous donation felt like it is not working properly. It was unpredictable.
**Now in order enable the button click next on the login step you have to either**:
* Be logged in in which case if you click next you will donate with your name
* Click the anonymous tab in which case it doesn't matter if you are logged in or not you will in both cases donate as *Anonymous*

## Screenshots:

| donation-anon-before | donation-anon-after |

## Testing

Went through the donation flow in all 3 cases:
- Logged in normal
- Logged in anonymous
- Just anonymous

### DIFF
--- a/src/components/one-time-donation/RegisterDialog.tsx
+++ b/src/components/one-time-donation/RegisterDialog.tsx
@@ -46,7 +46,7 @@ export default function RegisterForm() {
       if (resp?.ok) {
         setLoading(false)
         AlertStore.show(t('auth:alerts.welcome'), 'success')
-        formik.values.isAnonymous = false
+        formik.setFieldValue('isAnonymous', false)
         setStep(2)
       }
     } catch (error) {
--- a/src/components/one-time-donation/Steps.tsx
+++ b/src/components/one-time-donation/Steps.tsx
@@ -25,7 +25,7 @@ import { useCurrentPerson } from 'common/util/useCurrentPerson'
 
 const initialValues: OneTimeDonation = {
   message: '',
-  isAnonymous: true,
+  isAnonymous: false,
   amount: '',
   amountWithFees: 0,
   cardIncludeFees: false,
@@ -64,7 +64,6 @@ export default function DonationStepper({ onStepChange }: DonationStepperProps) 
     return session && session.accessToken ? true : false
   }
 
-  initialValues.isAnonymous = !isLogged()
   const userEmail = session?.user?.email
 
   const donate = React.useCallback(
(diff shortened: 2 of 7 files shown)


src/components/one-time-donation/RegisterDialog.tsx
```diff
@@ -46,7 +46,7 @@ export default function RegisterForm() {
       if (resp?.ok) {
         setLoading(false)
         AlertStore.show(t('auth:alerts.welcome'), 'success')
-        formik.values.isAnonymous = false
+        formik.setFieldValue('isAnonymous', false)
```
dimitur2204, line 49: I suspect the strange behaviour before was coming from either here!

src/components/one-time-donation/Steps.tsx
```diff
@@ -76,7 +75,7 @@ export default function DonationStepper({ onStepChange }: DonationStepperProps)
         firstName: values?.personsFirstName ? values.personsFirstName : 'Anonymous',
         lastName: values?.personsLastName ? values.personsLastName : 'Donor',
         personEmail: values?.personsEmail ? values.personsEmail : userEmail,
-        isAnonymous: values?.isAnonymous ?? !isLogged(),
+        isAnonymous: values?.isAnonymous !== undefined ? values.isAnonymous : true,
```
dimitur2204, line 78: Or here

src/components/one-time-donation/steps/SecondStep.tsx
```diff
@@ -15,8 +17,13 @@ export default function SecondStep() {
   const { data: session } = useSession()
 
   const [value, setValue] = useState('1')
-
+  const formik = useFormikContext<OneTimeDonation>()
   const handleChange = (event: React.SyntheticEvent, newValue: string) => {
+    if (Number(newValue) === 3) {
```
kachar, line 22: @dimitur2204 It's not really clear to me what the constant `3` stands for. 

If we're using that as a number of step let's name it in some kind of enum or similar so it brings more context.

Similar to:

```tsx
if (Number(step) === Steps.Anonymous) {
  // ...
}
```

kachar: As I see we can also use names instead of numbers for the tabs.

Ex. `logged`, `newProfile`, `anonymous` :thought_balloon:

kachar (approved):
:raised_hands: @dimitur2204 The new flow really simplifies the donation success path

PR #1135, opened 2022-11-20
Modularize MUI imports and bump up MUI version

dimitur2204:
## Motivation and context

The local development server was very slow.
The problem seemed to be with `@mui` see this: https://github.com/vercel/next.js/discussions/37614

I didn't find a way to improve it other than this experimental flag also took the opportunity to bump up the MUI version
https://nextjs.org/docs/advanced-features/compiler#modularize-imports

Locally for me it is way faster now. Instead of compiling 13000 modules on page change they dropped to 3000 max and only on certain pages.


kachar (approved):
Nice idea to speed up the loading. Thanks for the R&D @dimitur2204

Issue #1111, opened 2022-11-17
[Bug] Campaigns progress bars mismatch.


kabaivanov:
Upon loading the homepage _sometimes_ there appears to be a progress bar mismatch.
(Second time see this happen, but this time was able to make screenshots.)

**To Reproduce[hard]**
1. Go to 'Homepage'
2. See campaign progress bars reporting wrong values.
3. Resize browser window.
4. Bars shift, but still some are wrong.
5. Refresh and it's gone. _Hard to reproduce. Does not always happen._

**Expected behavior**
Progress bars that always reflect the actual progress of the campaigns in real time.

**Screenshots**
<img width="1992" alt="campaign-progress-bar-bug" src="https://user-images.githubusercontent.com/118026471/202327194-79022885-c878-44f9-98d0-dee161478341.png">
<img width="1325" alt="campaign-progress-bar-bug-resized" src="https://user-images.githubusercontent.com/118026471/202327219-ce136c52-717a-4304-b099-95ed565c3fad.png">

**Desktop:**

- OS: Mac OS X 12.4
- Browser: Safari 
- Version: 15.5

ani-kalpachka:
I tested this issue for desktop and mobile on Chrome and Firefox and could not reproduce it.
Could you confirm it is fixed?

igoychev:
The issue was caused by the randomization of the order of campaigns. The logic was refactored so not the progress bar mismatch is not happening anymore.

PR #1152, opened 2022-11-21
1150 - add array shuffling on the server prefetch

### LINKED ISSUE #1150: [Bug] Hydration error on the Campaigns Page
**Describe the bug**
image

**To Reproduce**

1. Go to /campaigns on the local environment
2. See error

dimitur2204:
## Motivation and context
Fixes #1150.

Upgraded `react-query` to the newest version, which includes a fix to properly hydrate after React 18.

Changes:
* We now **have to** provide an array of queryKeys, instead of a single one as was possible before. 
* Import has changed from `react-query` to `@tanstack/react-query` 

## Other fixes and notes
Please note the change from `queryFn` to `queryFnFactory` I did it, because of Typing issues, but the `queryFnFactory` has some additonal logic, I hope that this does not slow down the queries or introduce problems.

Fixed LocalizationProvider to now work properly and the translations issues on the profile page are resolved [1150 - fix localization provider locale to new api](https://github.com/podkrepi-bg/frontend/pull/1152/commits/cd81ae5215694c72aa441d0306b0fde0dde54f94)
## Testing
Tried the list and it now seems to be both shuffled and without an error.


kachar:
@dimitur2204 This probably also should fix #1111

src/common/util/shuffle.ts
```diff
@@ -1,6 +1,8 @@
...
-  for (let i = array.length - 1; i > 0; i--) {
+export function shuffleArray<T>(array: T[]) {
+  const arrCopy = [...array]
+  for (let i = arrCopy.length - 1; i > 0; i--) {
     const j = Math.floor(Math.random() * (i + 1))
-    ;[array[i], array[j]] = [array[j], array[i]]
+    ;[arrCopy[i], arrCopy[j]] = [arrCopy[j], arrCopy[i]]
   }
+  return arrCopy
 }
```
kachar, line 8: @dimitur2204 We use `lodash` in the project so I find it easy to replace this custom random fn with lodash util for example `shuffle`. 

```tsx
_.shuffle([1, 2, 3, 4]);
```

https://lodash.com/docs/4.17.15#shuffle

igoychev (changes requested):
Was quite happy that you found way to fix, however just tested and unfortunately it looks like the shuffling does not work anymore. 

What I see is that campaigns get shuffled, but a quick flicker returns them to always the same order. Tried with 3 tabs and 2 browsers (chrome/firefox) and all have the same order for all campaigns after refreshing.

src/common/util/shuffle.ts
dimitur2204: Will do that!

dimitur2204:
The issue was fixed, but something else occurred now after the update. I noticed we have a couple of pages that do this check. This way for a second between the DOM is loaded we can see a `<NotFoundPage/>`. @kachar I feel like we shouldn't be returning a not found page like that. What do you say?
image

kachar:
> The issue is fixed, but something else occurred now after the update. I noticed we have a couple of pages that do this check. This way for a second until the DOM is loaded we can see a `<NotFoundPage/>`. @kachar I feel like we shouldn't be returning a not found page like that. What do you say? image

I think we need to use `getServerSideProps` which returns `{ notFound: true }` if the campaign is not available in the API.

valkirilov (approved):
Good job! It seems that the hydration issue is fixed after some tests on my local, I confirm it. Although I still don't feel confident enough and I'm still exploring the website and "learning" the different features which are available.

PS: I can confirm a performance improvement as well, the site is working way more faster than before. Kudos :clap:

igoychev (approved):

igoychev:
Amazing job Nizamov!

kachar (approved):
Kudos for the enhancement @dimitur2204 

Works great now!

PR #1156, opened 2022-11-22
Improve the UX on the FAQ Page when searching

valkirilov:
## Motivation and context
Improve the UX on the FAQ Page when searching

- auto-select the first category in the left Tabs pane if we don't have matching results in the current category when searching (see attached videos)
- fix the size/proportion of the tab pane with the categories, so it won't flicker when searching and reducing the amount of visible categories (see attached videos)
- minor refactoring of the names in the `FaqPage` component, so it will be more clear that `value` state stands for `selectedCategory`

References: #1017

## Videos:

Before: 
https://user-images.githubusercontent.com/1233496/203375670-91e59b96-814d-412b-85e6-f5255f7a4e11.mp4

After: 
https://user-images.githubusercontent.com/1233496/203375769-d7850fb6-f9e4-4aec-b96c-44f2b081d416.mp4

## Testing

I tested the new changes manually during the development process by typing different keywords into the new "Search" field on the FAQ page which results in reducing the amount of the categories with available matching results (like "процент", "общество", etc.).

References pages:
- http://localhost:3040/faq
- http://localhost:3040/faq/common-questions

### DIFF
--- a/src/components/faq/FaqPage.tsx
+++ b/src/components/faq/FaqPage.tsx
@@ -1,7 +1,7 @@
 import { TabContext, TabPanel } from '@mui/lab'
 import { Stack } from '@mui/material'
 import { useTranslation } from 'next-i18next'
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 
 import Layout from 'components/layout/Layout'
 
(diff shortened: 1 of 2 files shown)


src/components/faq/FaqPage.tsx
```diff
@@ -32,22 +32,31 @@ const FAQ_PAGE_QUESTIONS: Record<string, ContentType[]> = {
...
   const [searchKeyword, setSearchKeyword] = useState('')
 
   const faqQuestionsData = filterFaqQuestions(FAQ_PAGE_QUESTIONS, searchKeyword)
   const faqCategories = Object.keys(faqQuestionsData) as FaqCategory[]
 
+  useEffect(() => {
+    if (faqCategories.length > 0 && !faqCategories.includes(selectedFaqCategory)) {
+      setSelectedFaqCategory(faqCategories[0])
+    }
+  }, [searchKeyword])
```
kachar, line 45: @valkirilov This can be changed to `useMemo` for brevity.
When `useEffect` only uses `setState` we can normally convert it.

See https://beta.reactjs.org/learn/you-might-not-need-an-effect




```suggestion
  const [searchKeyword, setSearchKeyword] = useState('')

  const faqQuestionsData = useMemo(() => filterFaqQuestions(FAQ_PAGE_QUESTIONS, searchKeyword), [searchKeyword])
  const faqCategories = useMemo(() => Object.keys(faqQuestionsData) as FaqCategory[], [faqQuestionsData)

  const selectedFaqCategory = useMemo(() => {
    if (faqCategories.length > 0 && !faqCategories.includes(selectedFaqCategory)) {
      return faqCategories[0]
    }
    return section
  }, [faqCategories, section])
```

src/components/faq/VerticalTabs.tsx
```diff
@@ -28,12 +28,12 @@ const VerticalTabs = ({ faqCategories, setSelectedFaqCategory }: Props) => {
...
 
   return (
-    <Box sx={{ bgcolor: 'background.paper', display: 'flex', justifyContent: 'center' }}>
+    <Box sx={{ bgcolor: 'background.paper', display: 'flex', justifyContent: 'center', flex: 1 }}>
       <TabList
         orientation="vertical"
         variant="scrollable"
         onChange={handleChange}
-        sx={{ borderRight: 1, borderColor: 'divider', mb: 4 }}>
+        sx={{ borderRight: 1, borderColor: 'divider', mb: 4, width: '100%' }}>
```
kachar, line 36: Good call to add a fixed width of the sections container :+1:

kachar (changes requested):
Well done @valkirilov!

I've added a suggestion for removal of one `useEffect` and optimization to reduce the re-render of search filtering so it happens only when the keyword changes.

Please review it and let's discuss if needed

src/components/faq/FaqPage.tsx
valkirilov: Great suggestion! The changes are applied, but I had to remove the option to select categories from the Tabs via the state setter (because we remove it from the parent component). It's a bit slower on local, because we rely only on redircet of the page to change the selected category/section, but I guess that it won't be an issue on DEV and PROD:
- rework the `FaqPage` component so we memoize the selected category and the logic which we apply to it when using the search filter (to keep always a selected category when we run out of matches in the current one)
- remove the `setSelectedCategory` from the `VerticalTabs` component and make it so we rely only on the redirect of the page to change the selected category

kachar (approved):
Nicely done @valkirilov Thanks!

Issue #1153, opened 2022-11-21
[Bug] Change the language from English to Bulgarian automatically in Campaigns


YanitsaTantarova:
Change the language from English to Bulgarian automatically Campaigns

**Steps to reproduce the behavior: **
1. Go to podkrepi.bg
2. Click on EN button to change the language.
4. Scroll down to Campaigns or select from Drop-down menu a Campaign. The buttons are in English "Support now"
5.  Click on random Campaign 
6. See the language change from English to Bulgarian automatically. The buttons are "Сподели", "Подкрепи".

**Expected behavior**
The language must stay English if you don`t change it.

**Desktop: ** 
- Browser: Chrome
- Version 107.0.5304.107

PR #1159, opened 2022-11-23
1153 Added component type Link to CardActionArea

AnzheloD:
Added component type Link to CardActionArea, because by default it uses anchor tag which refreshes the page and removes the locale from the route.

Issue link https://github.com/podkrepi-bg/frontend/issues/1153


kachar (approved):
Nice update @AnzheloD 

This enhances the UX as well as the code

Issue #1013, opened 2022-09-02
[BankAccounts] Improvements on create/edit BankAccount


igoychev:
**Describe the bug**
When creating or editing new Bank Account:
 - Status field should be a dropdown with available status values ( translated)
 - Account Type should be a dropdown with available status values ( translated)
 - fingerprint to be hidden - not needed for now

**To Reproduce**

1. Login as Admin user
2. Navigate to /admin/bankaccounts
3. Click on create/edit bank account
4. See the described improvements needed above

AndreyGoranov:
Is this issue available to work on? If so can you please assign me to it.

igoychev:
Yes, free and happy to see you want to take it!

kachar:
Closed by https://github.com/podkrepi-bg/frontend/pull/1160

PR #1160, opened 2022-11-23
1013 - Update Bank Account Forms

AndreyGoranov:
## Motivation and context
Bank Account Forms needed some small tweaks
Issue: https://github.com/podkrepi-bg/frontend/issues/1013

Problem solved by: 
1.  Changing the Status and Account Type form fields from TextField to SelectFields
2.  Hiding the fingerprint field by commenting it out

## Screenshots:

| Before    
image
          | After     
image
image
         |

## Testing

1. Tested if fields are prepopulated when editing
2. Switching between different options
4. Making create request
5. Making edit request
6. Translation

- `env var` : env var details

**New dependencies**:

**New dev dependencies**:


public/locales/en/bankaccounts.json
```diff
@@ -3,13 +3,26 @@
...
+    "AccountHolderType": "Account holder type",
+    "bankName": "Bank name",
+    "bankIdCode": "BIC/SWIFT",
+    "fingerprint": "E-Sign"
+  },
+  "status": {
+    "new": "New",
+    "validated": "Validated",
+    "verified": "Verified",
+    "verification_failed": "Verified Failed",
```
kachar, line 19:
```suggestion
    "verification_failed": "Verification Failed",
```

public/locales/bg/bankaccounts.json
```diff
@@ -3,13 +3,26 @@
...
+    "accountHolderType": "Вид акаунт",
+    "bankName": "Име на банка",
+    "bankIdCode": "BIC/SWIFT",
+    "fingerprint": "Е-Подпис"
+  },
+  "status": {
+    "new": "Нов",
+    "validated": "Валидиран",
+    "verified": "Верифициран",
+    "verification_failed": "Неуспешно верифициран",
```
kachar, line 19:
```suggestion
    "verification_failed": "Неуспешна верификация",
```

src/components/bankaccounts/BankAccountsEditForm.tsx
```diff
@@ -33,7 +35,7 @@ export default function BankAccountsEditForm({ id }: Props) {
     accountHolderType: data?.accountHolderType,
     bankName: data?.bankName,
     bankIdCode: data?.bankIdCode,
-    fingerprint: data?.fingerprint,
+    // fingerprint: data?.fingerprint,
```
kachar, line 38: @AndreyGoranov Why do we remove the `fingerprint` here? Looks like this is a temporary local solution to some kind of problem.

src/components/bankaccounts/BankAccountsEditForm.tsx
```diff
@@ -73,34 +75,50 @@ export default function BankAccountsEditForm({ id }: Props) {
...
+              label="bankaccounts:fields.status"
+              name="status"
+              options={Object.keys(BankAccountStatus).map((key: string) => {
+                return {
+                  key,
+                  value: BankAccountStatus[key as BankAccountStatus],
+                  name: t(`bankaccounts:status.${key}`),
+                }
+              })}
+            />
```
kachar, line 88: Let's extract this field as a separate reusable `BankAccountStatusSelect` component.

If we extract it will be quite simple to reuse it in both `BankAccountsEditFrom` and `BankAccountsForm` forms


```tsx
type Props = {
  name?: string
}
export default function BankAccountStatusSelect({ name = 'status' }: Props) {
  return (
    <FormSelectField
      name={name}
      label="bankaccounts:fields.status"
      options={Object.keys(BankAccountStatus).map((key: string) => {
        return {
          key,
          value: BankAccountStatus[key as BankAccountStatus],
          name: t(`bankaccounts:status.${key}`),
        }
      })}
    />
  )
}
```

I suggest that we add the `name` prop explicitly so it's more reusable in different forms. Sometimes we'll need it as `status` but other times we might need the field name to be `bank-account-status` for example.

src/components/bankaccounts/BankAccountsEditForm.tsx
```diff
@@ -73,34 +75,50 @@ export default function BankAccountsEditForm({ id }: Props) {
...
+            <FormSelectField
+              name="accountHolderType"
+              label="bankaccounts:fields.accountHolderType"
+              options={Object.keys(AccountHolderType).map((key: string) => {
+                return {
+                  key,
+                  value: AccountHolderType[key as AccountHolderType],
+                  name: t(`bankaccounts:accountHolderType.${key}`),
+                }
+              })}
```
kachar, line 110: Let's extract this as separate reusable `AccountHolderSelect` component.

src/components/bankaccounts/BankAccountsEditForm.tsx
```diff
@@ -73,34 +75,50 @@ export default function BankAccountsEditForm({ id }: Props) {
...
-            <FormTextField type="text" name="bankIdCode" label="bankaccounts:bankIdCode" />
+            <FormTextField type="text" name="bankName" label="bankaccounts:fields.bankName" />
           </Grid>
           <Grid item xs={12}>
-            <FormTextField type="text" name="fingerprint" label="bankaccounts:fingerprint" />
+            <FormTextField type="text" name="bankIdCode" label="bankaccounts:fields.bankIdCode" />
           </Grid>
+          {/* <Grid item xs={12}>
+            <FormTextField type="text" name="fingerprint" label="bankaccounts:fields.fingerprint" />
+          </Grid> */}
```
kachar, line 121: Maybe we need to uncomment this

src/components/bankaccounts/BankAccountsForm.tsx
```diff
@@ -26,10 +27,10 @@ export const validationSchemaBankAccForm: yup.SchemaOf<BankAccountsData> = yup
     status: yup.string().trim().min(1).max(100).required(),
     ibanNumber: yup.string().trim().min(5).max(34).required(),
     accountHolderName: yup.string().trim().min(10).max(100).required(),
-    AccountHolderType: yup.string().trim().min(1).max(100).required(),
+    accountHolderType: yup.string().trim().min(1).max(100).required(),
     bankName: yup.string().trim().min(10).max(100).required(),
     bankIdCode: yup.string().trim().min(8).max(11).required(),
-    fingerprint: yup.string().trim().min(10).max(100).required(),
+    // fingerprint: yup.string().trim().min(10).max(100).required(),
```
kachar, line 33: Maybe we need to uncomment this

kachar (changes requested):
@AndreyGoranov Pretty good contribution, thank you! :bow: 

I've shared some ideas for the reusability of the code, please check them out.

I've noticed that the `fingerprint` field is commented out, but can't find the reason for that. If it's just temporary please undo that.

src/components/bankaccounts/BankAccountsEditForm.tsx
AndreyGoranov: The ticket says: "fingerprint to be hidden - not needed for now", so should I delete it?

src/components/bankaccounts/BankAccountsEditForm.tsx
AndreyGoranov: Yes, I wasn't sure what to do so I left it commented out.

kachar (approved):
Outstanding contribution @AndreyGoranov 

Thank you!

PR #1162, opened 2022-11-23
Update Mobile Nav styles

ani-kalpachka:
- Removed createStyles/makeStyles from MobieNav menu,

- Fixed the menu for authenticated user:

| authenticated0| authenticated1 |

- Centered the Login / Register buttons for not authenticated users:

| na0 | na1 |


src/components/layout/nav/AuthLinks/AuthLinks.styled.tsx
```diff
@@ -0,0 +1,32 @@
+import { Box } from '@mui/material'
+import { styled } from '@mui/material/styles'
+
+import theme from 'common/theme'
+import LinkButton from 'components/common/LinkButton'
+
+export const AuthLinksWrapper = styled(Box)(() => ({
+  display: 'flex',
+  alignItems: 'center',
+  borderTop: '2px solid lightgrey',
```
dimitur2204, line 10: Maybe we should define these somewhere on the theme if we use them a lot.

dimitur2204 (approved):
It's looking amazing!

src/components/layout/nav/MobileNav/MobileNav.styled.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const StyledPodkrepiIcon = styled(PodkrepiIcon)(() => ({
+  margin: theme.spacing(0.7, 0, 1.5, 0),
+}))
+
+export const NavMenuWrapper = styled(Box)(() => ({
+  display: 'flex',
+  justifyContent: 'center',
+  padding: theme.spacing(2),
+  flexDirection: 'column',
+}))
```
kachar, line 23: We can replace this whole styled Box with Stack as it contains all these css attributes as props which makes it really easy to use, also easy to get responsive.

src/components/layout/nav/MobileNav/MobileNav.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <Grid item>
+            <DonationMenuMobile />
+          </Grid>
+          <Grid item>
+            <ProjectMenuMobile />
+          </Grid>
+          <AuthLinks />
+          <LocaleButtonWrapper>
+            <LocaleButton />
+          </LocaleButtonWrapper>
```
kachar, line 66: Here we mix Grid items with Box elements.

I'm not sure it will propagate the grid styles and spacing accordingly without another Grid item.

kachar (approved):
Thanks @ani-kalpachka for the invested time and energy! 

Looks like we're few steps closer to taking out the old inherited styles from mui v4. 

We can integrate the suggestions on the next iterations. I really liked the idea of @dimitur2204 to move the border style definition to the theme.

Issue #1011, opened 2022-09-01
[Feature] When choosing "Трезор" for "Добавяне на разход" to have an option to choose a campaign name or to be able paste the ID.


anikolova-bg:
**The challenge**
When choosing "трезор" while adding a cost it is difficult to find the right ID (when many campaigns it will be even more challenging). 

**Describe the solution you'd like**
Either to have the option to choose by name or to paste the ID.

**Describe alternatives you've considered**
To see only the "трезор" of the campaings that the organiser is managing, respectively the coordinator is monitoring. 
![Screenshot (895)](https://user-images.githubusercontent.com/74831444/187944473-42b2aa17-64bc-494f-aeb5-fa403704dcee.png)

PR #1165, opened 2022-11-23
1011 Fixed broken translation

AnzheloD:
I can see that the ID's have been changed with names in previous task.
This PR fixes broken label translation

Issue link: [https://github.com/podkrepi-bg/frontend/issues/1011](url)

## Screenshots:

Before|After
---|---
|

### DIFF
--- a/src/components/vaults/VaultSelect.tsx
+++ b/src/components/vaults/VaultSelect.tsx
@@ -21,7 +21,7 @@ export default function VaultSelect({
   handleVaultSelected,
   ...textFieldProps
 }: Props) {
-  const { t } = useTranslation()
+  const { t } = useTranslation('vaults')
 
   const [field, meta] = useField(name)
   const { setFieldValue } = useFormikContext()
@@ -48,7 +48,7 @@ export default function VaultSelect({
         {...textFieldProps}
         onChange={handleChange}>
         <MenuItem value="" disabled>
-          {t(label)}
+          {t(`fields.${name}`)}
         </MenuItem>
         {vaults?.map((value, index) => (
           <MenuItem key={index} value={value.id}>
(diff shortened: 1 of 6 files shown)


src/components/vaults/VaultSelect.tsx
```diff
@@ -48,7 +48,7 @@ export default function VaultSelect({
         {...textFieldProps}
         onChange={handleChange}>
         <MenuItem value="" disabled>
-          {t('fields.' + name)}
+          {t(label)}
```
kachar, line 51: Looks like this solution contradicts a pattern that we use all over the system.

In many places we add the key `fields` to the language json so we can define the names of fields in a single location.

Looks like in this specific place we don't have the correct key in the language file so it shows the actual key.

I'd like to see a solution that complies with the widely used approach in the code. Indeed it's not written anywhere so we can correct that and add a hint so new devs are not confused.

We've got this code, but it's not wrapped in `fields` object:

https://github.com/podkrepi-bg/frontend/blob/7411045c290e6f70d3b8254ae3f3e38623db690f/public/locales/bg/recurring-donation.json#L12

We need to define the following `fields` in order to have `<VaultSelect />` component working correctly:

```json
"fields": {
  "vaultId": "ID на трезор",
  "sourceVaultId": "От трезор",
  "targetVaultId": "Към трезор"
}
```

kachar (changes requested):

src/components/vaults/VaultSelect.tsx
kachar: We also need to add the translation domain to the hook

```tsx
const { t } = useTranslation('vaults')
```

and load it in the concerned pages

```
export const getServerSideProps = securedAdminProps(
  ['common', 'auth', 'vaults', 'admin', 'validation'],
  () => endpoints.vaults.vaultsList.url,
)```

kachar (approved):
Awesome work @AnzheloD 
Thanks for the contribution! I feel this way of handling the translations will be much more future-proof.

Issue #1166, opened 2022-11-23
[Bug] Flashing 404 page on page change


dimitur2204:
**Describe the bug**
Flashing of 404 page between page changes.
Happens on all pages that have this check:
<img width="626" alt="image" src="https://user-images.githubusercontent.com/61479393/203654014-391a21e1-f97a-4bad-94db-26514f68d8e2.png">

**To Reproduce**

1. Go to the homepage on https://dev.podkrepi.bg
2. Click on any campaign
3. While loading you can see a 404 page flash between the loads.

PR #1169, opened 2022-11-23
1166 - Fix flashing 404 page between page changes

dimitur2204:
## Motivation and context

Addressed the issue specified here https://github.com/podkrepi-bg/frontend/pull/1152#issuecomment-1323969262 and #1166.

This returning of a `<NotFoundPage />` in the components that were changed seemed to cause the issue. Added a loader in it's place. This way though the case where `data` is empty is not handled, but we shouldn't get to this case since if the `campaigns/<slug>` does not exist the server should give us back a 404 and that will get handled by Next.

|404-flash-bug|404-flash-fix|

* Also added a `staleTime` to the campaign list query, which keeps them in cache for 3 hours. I felt like the campaigns getting reshuffled on every page change was confusing.

## Testing

Tried all the pages that had the if check for the `<NotFoundPage>`.

### DIFF
--- a/src/common/hooks/campaigns.ts
+++ b/src/common/hooks/campaigns.ts
@@ -21,8 +21,8 @@ const shuffleQueryFn: QueryFunction<CampaignResponse[]> = async ({ queryKey }) =
 
 export function useCampaignList() {
   return useQuery<CampaignResponse[]>([endpoints.campaign.listCampaigns.url], shuffleQueryFn, {
-    // Add 3 hours of cache time
-    staleTime: 1000 * 60 * 60 * 3,
+    // Add 15 minutes of cache time
+    staleTime: 1000 * 60 * 15,
   })
 }
 
--- a/src/common/hooks/campaigns.ts
+++ b/src/common/hooks/campaigns.ts
@@ -20,7 +20,10 @@ const shuffleQueryFn: QueryFunction<CampaignResponse[]> = async ({ queryKey }) =
 }
 
 export function useCampaignList() {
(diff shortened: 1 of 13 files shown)


ani-kalpachka (approved):

src/common/hooks/campaigns.ts
```diff
@@ -20,7 +20,10 @@ const shuffleQueryFn: QueryFunction<CampaignResponse[]> = async ({ queryKey }) =
 }
 
 export function useCampaignList() {
-  return useQuery<CampaignResponse[]>([endpoints.campaign.listCampaigns.url], shuffleQueryFn)
+  return useQuery<CampaignResponse[]>([endpoints.campaign.listCampaigns.url], shuffleQueryFn, {
+    // Add 15 minutes of cache time
+    staleTime: 1000 * 60 * 15,
```
kachar, line 25: Nice approach to caching the result! Awesome :raised_hands:

src/service/restRequests.ts
```diff
@@ -23,7 +28,13 @@ export const queryFn: QueryFunction = async function ({ queryKey }) {
...
 }
 
-export const queryFnFactory = <T>(token?: string): QueryFunction<T> =>
+export const queryFnFactory = <T>(): QueryFunction<T> =>
+  async function ({ queryKey }) {
+    const response = await apiClient.get(queryKey.join('/'))
+    return await response.data
+  }
+
+const attachTokenToQueryFn = <T>(token?: string): QueryFunction<T> =>
```
kachar, line 37: That's really better - more modularity and reusability

kachar (approved):

PR #1185, opened 2022-11-24
1172 - Prevent campaign banner images from being cut on mobile devices

### LINKED ISSUE #1172: [Bug] Campaigns' banner images are cut on tablet and mobile devices
Steps to reproduce:
1. Go to campaigns listing page.
2. Choose a campaign.
3. Apply responsive view smaller than 900px screen width. 
   - Note: This is a default MUI breakpoint - md, medium: 900px
5. Observe that the image is cut from the top because the Donation block overlaps it:

image

Expected behavior:
Campaigns' images are fully displayed.

kabaivanov:
Closes #1172 

## Motivation and context
A CSS change is required in order to prevent the banner images from being cut while viewing the single campaign pages on mobile devices.
Solved the issue by applying CSS media query to the .CampaignDetails-banner class

## Screenshots:

Before|After
---|---
Screenshot 2022-11-24 at 23 19 07|Screenshot 2022-11-24 at 23 19 25

**Affects:** All single campaign pages


ani-kalpachka:
This solves the problem from 600px to 900px screen width but  for less than 600px it adds white space:

Screenshot 2022-11-25 074638

Maybe we are going to need one more breakpoint.

Also could we please use either top or marginTop property and avoid using both?

kabaivanov:
> Thanks for your effort, @kabaivanov!
> 
> This solves the problem from 600px to 900px screen width but for less than 600px it adds white space:
> 
> Screenshot 2022-11-25 074638
> 
> Maybe we are going to need one more breakpoint.
> 
> Also could we please use either top or marginTop property and avoid using both?

@ani-kalpachka you're right about the margin, I'll adjust.

And I see another issue that causes the shift below 600px. -  It has to do the min-height value of .MuiBox-root shifting from  64px to 56px below the 600px breakpoint. I'll take care of that in an update later today.

kachar:
@kabaivanov @ani-kalpachka It's gonna be awesome if we get rid of the hardcoded `px` values as well as the `!important` rule overrides.

I expect that the users will upload images with different height/width ratio so we may use `objectFit: 'cover'`

kabaivanov:
@ani-kalpachka , switched to utilising the margin-top property and added a sm breakpoint spacing fix to  ViewCampaignPage.tsx  that targets the sub 300px issue as well.

ani-kalpachka (approved):

PR #1197, opened 2022-11-28
[Fix] - Translations on the profile page

### LINKED ISSUE #1187: [Bug] Problem with translations on My Profile page
**Describe the bug**
After login, MyProfile page is shown, but the translation text are missing and only text.ids are shown. See screenshot.

**To Reproduce**

1. Go to https://podkrepi.bg/login
2. Login with any user
3. My Profile page is loading and no translations are seen 

Side notes:
 - changing EN <-> BG doesn't help, however the font style is changed
 - Translations appear after clicking F5 for reload, but after navigating to another page and returning back even the Menu translations become lost.

**Expected behavior**
All text on My Profile to be translated.

**Screenshots**
Screenshot 2022-11-25 at 12 07 16

**Desktop (please complete the following information):**

- OS: MacOS
- Browser Chrome

dimitur2204:
Closes #1187.
Removed a redirect in the `next.config.js`

## Motivation and context

Seems like the redirect bugged out the server side translations somehow. The redirect was supposed to redirect to `/profile/donations`, which is the first Tab on the page, but it does seem to work fine without the redirect.

## Screenshots:

Before|After
---|---
image|image

## Testing
Tried logging out and in, clearing the cache, refreshing and it always seemed to translate.

### Affected urls
Affected: `/profile` when you are logged in.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -46,15 +46,6 @@ const moduleExports = {
       process.env.GHOST_API_URL?.replace('https://', '') || 'blog.podkrepi.bg',
     ],
   },
-  async redirects() {
-    return [
-      {
-        source: '/profile',
-        destination: '/profile/donations',
-        permanent: true,
-      },
-    ]
-  },
   async rewrites() {
     return [
       {


kachar (approved):
Nice catch @dimitur2204 
Thanks for following this rabbit hole till the end :hole: :rabbit2:

PR #1200, opened 2022-11-28
Fix visual issue with the Donations block on the Campaign page

### LINKED ISSUE #1108: [Bug] Campaign's donation block is not fully visible
Steps to reproduce:
1. Go to https://podkrepi.bg/campaigns
2. Open a campaign
3. Observe the campaign's donation block - it is not fully visible at the bottom until the user scrolls down to the bottom of the page and the last donor and the pagination are cut:

image

Expected behavior: It should be fully visible.

valkirilov:
Fix visual issue with the "Donations" block on the "Campaign" page

- make the list with the donations scrollable and always fit the viewport on the Campaign details page

Closes #1108

## Motivation and context

Observe the campaign's donation block - it is not fully visible at the bottom until the user scrolls down to the bottom of the page which makes the UX really bad, especially for campaign with a lot of comments (it takes some time until you reach the bottom of the page in order to see the full list with the donations).

Examples:
- https://podkrepi.bg/campaigns/tanya-kosmicheskoto-momiche-na-bulgariya-po-putya-kum-zvezdite

### How did you solve the problem?
I made some minor changes in the way we make the donations block scrollable:
- changed the size of the parent element by putting a maximum height so it won't exceed the viewport anymore
- changed the overflow of the parent element, so it will trigger properly the auto-scroll of the inline block with the donations

### Any links to external sources of documentation:
I made an example js fiddle to showcase the changes with an isolated and a bit more simpler markup:
- https://jsfiddle.net/e2g6nthy/

## Screenshots:

Before|After
---|---
image|image

And here it is how it looks like on a bigger screen when we have enough viewport height to show all the donations without an inline scroll:
image

### Steps to test

1. Go to https://podkrepi.bg/campaigns
2. Open a campaign
3. Observe the campaign's donation block

_Note: For some reason there are no any campaigns on DEV right now, so you can test it only on your LOCAL env._

### DIFF
--- a/src/components/campaigns/DonorsAndDonations.tsx
+++ b/src/components/campaigns/DonorsAndDonations.tsx
@@ -13,18 +13,14 @@ import { moneyPublic } from 'common/util/money'
 const PREFIX = 'DonorsAndDonations'
 
 const classes = {
-  donationsWrapper: `${PREFIX}-donationsWrapper`,
   donationItemWrapper: `${PREFIX}-donationItemWrapper`,
   donationQuantityAndTimeWrapper: `${PREFIX}-donationQuantityAndTimeWrapper`,
   separatorIcon: `${PREFIX}-separatorIcon`,
 }
 
 const Root = styled('div')(({ theme }) => ({
-  [`& .${classes.donationsWrapper}`]: {
-    marginTop: theme.spacing(5),
-    maxHeight: 400,
-    overflowY: 'scroll',
-  },
+  marginTop: theme.spacing(5),
+  overflowY: 'auto',
 
   [`& .${classes.donationItemWrapper}`]: {
     display: 'flex',
@@ -62,7 +58,7 @@ export default function DonorsAndDonations({
 
   return (
     <Root>
-      <Grid item className={classes.donationsWrapper}>
+      <Grid item>
         {donationsToShow && donationsToShow.length !== 0 ? (
           donationsToShow.map(({ person, amount, createdAt, currency }, key) => (
             <Grid key={key} className={classes.donationItemWrapper}>
--- a/src/components/campaigns/InlineDonation.tsx
+++ b/src/components/campaigns/InlineDonation.tsx
@@ -28,6 +28,7 @@ const classes = {
   inlineDonationWrapper: `${PREFIX}-inlineDonationWrapper`,
   reachedMoney: `${PREFIX}-reachedMoney`,
   targetMoney: `${PREFIX}-targetMoney`,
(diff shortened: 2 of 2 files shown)


valkirilov:
> By design we do not have a scroll on the donators' list, but a pagination instead.  

I was not sure whether I took the correct approach to resolve this one, but also I saw that we already have a scrollbar (both present in the code and visible in the UI).
image

> I think we should rather try to display three donators per page and not have the scroll.

I agree on that, it won't be an issue to change it, but also I'm afraid that we're going to reduce way too much the amount of the records per page. And respectively, it will increase the amount of the pages, so I have some doubts regarding the UX if we decide to go in this direction :thinking: 

> Also I spotted one small UI improvement that could be done: the last donator would look better if it is displayed fully along with its name, donated money and time of donation.

Totally agree on that, but I'm a bit afraid, that it may be not that easy to make it work on all screen sizes (especially smaller laptops with not full hd displays)

Example: Here it is how it looks on 1366x768, we don't have space to show 2 donators :disappointed 
_(and if we have a bookmarks list in the browser, the space will be even less, so it will be only one donator)_:

image

Maybe we need to re-think the design of this block and where we actually want to put the focus? 
- for example, we can move the "Share" and "Donate" button next to each other, on a same line
- also, do we need to keep the counters for the donations/shares to be on multiple lines, this takes too much of the available height

1366x768|1920x1080
---|---
image|image

ani-kalpachka:
1. It is an UI discrepancy that we currently have an inactive scroll.

2. For this point I am referencing from the Figma design:
image
The new design of the donation block is not finalized yet, but with the current ticket I suggest we fix the cutting of the pagination until the component is redesigned because the user cannot go to the next page until he scrolls to the bottom of the page. 
However, I think your concern is valid. We could sync with the Design team about it.

3. The styles should only change from smaller devices than 900px (for such devices the donation component appears from the top and has different styles). For the given resolution I think it should behave the same as for dektop:

image

Here is a valid resolution scenario (Surface Pro 7):
image
In this case we could reduce the top spacing of the component. It would also look better for desktop.

4. The shares and the donators counters appear to take less space on the new design but I suggest we handle this fix in a separate ticket when it is finalized.

kabaivanov:
@valkirilov 
Not sure if you saw my comment on using a JS library developed to tackle exactly this issue.

FloatSidebar.js or sticky-sidebar.js are good examples – demos below.

https://github.com/vursen/FloatSidebar.js - [demo here](https://js-2sy9en.stackblitz.io/)
https://github.com/abouolia/sticky-sidebar - [demo here](https://abouolia.github.io/sticky-sidebar/examples/scrollable-element.html)

Issue #1077, opened 2022-10-31
New card delete account/profile


marianacat:
**Delete account/profile**
Please change the card when the delete account is clicked.
Login/personal information/delete account/profile

**Additional context**
The new design and text can be found in the Figma file here:
https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=14682%3A45152

![изтриване](https://user-images.githubusercontent.com/92619043/198989267-dd56ce86-559c-41cc-9f10-b81b155c0da6.png)

Katzarov:
Starting to work on this, can I be assigned to it please.

PR #1210, opened 2022-11-29
1077 - Update UI of delete account modal.

Katzarov:
Closes https://github.com/podkrepi-bg/frontend/issues/1077

## Motivation and context
Resolves issue [1077](https://github.com/podkrepi-bg/frontend/issues/1077). I used the proposed design from the ticket  as more of a guideline, so there are some differences.
- I put more  emphasis on the consequences of the action of deleting an account.
- I tried to incorporate the quote, but couldn't find a good fit for it.
- The design suggests that the delete button is disabled until a valid password is filled in. But I think I don't see this being done elsewhere ? -  (that being said, I think it is a good idea.)
- Apart from that, I roughly translated the rest of the design.

Noticed that for small screen sizes modals are still 70% wide, (and no scrollbars I think ?). So in this case, I had to make it a bit wider.
 

## Screenshots:

Before|After
---|---
image|imageimage

## Testing

- None of the deleted translations keys are referenced anywhere else.
- Tested with social login account (google) and with non social login account. The difference in UI is just that the social login case doesn't have the password field, and doesn't have a loading indicator on the delete button.
- Tested different screen sizes all the way down to iPhone SE.
- Tested with different system/browser font sizes, all the way up to Chrome's "Very Large". (that being said on mobile devices, with larger system fonts selected, I think it would still break )

### Affected urls
https://podkrepi.bg/profile/personal-information


src/components/auth/profile/DisableAccountModal.tsx
```diff
@@ -37,24 +38,40 @@ const StyledModal = styled(Modal)(({ theme }) => ({
...
     },
   },
+  [`& .${classes.subtitle}`]: {
+    fontFamily: 'Lato, sans-serif',
+    fontStyle: 'normal',
+    fontWeight: '400',
+    fontSize: theme.typography.pxToRem(14),
+    lineHeight: theme.spacing(3),
+    color: theme.palette.primary.main,
+    margin: `${theme.spacing(1)} 0`,
```
ani-kalpachka, line 53: We don't need to interpolate the theme values.
I suggest that here you could use margin: theme.spacing(1, 0),

Katzarov: Addressed, thank you.

ani-kalpachka (approved):

PR #1213, opened 2022-11-29
1187 - made all the profile links to redirect to /profile/donations

dimitur2204:
Again regarding #1187

After @AnzheloD pointed out he had the error again and the `redirect` in the `next.config.js` is only a server side redirect and didn't do anything about client side routing.
In this PR I changed it so now there is not index page for the `profile`,
All the routes that were leading to the index now lead to `/profile/donations` which to me makes more sense considering we have the tabbing functionality which also changes routes.

## Testing
Trying refreshing, logging in, out again! Seems to be consistent!

### Affected urls
- /profile does not exist anymore if you go to it you will get redirected to /profile/donations


kachar (approved):
Thanks, @AnzheloD for bringing that up, and @dimitur2204 for making it happen. :rocket:

PR #1214, opened 2022-11-30
1119 - Confirm password in registration field is now required

### LINKED ISSUE #1119: [Bug] confirm password field in registration is not required
**Describe the bug**
You can register without filing anything in confirm password field. It does its job only if you fill something.

**Expected behavior**
If the confirm password field is empty, there should be some error message and you should not be able to register without filling it.

jozef-drabik:
Now when someone want to register and they dont fill up confirmation password, it will throw error message for users.

Closes #1119 

## Screenshots:
Before:
image
After:
image
After:
image

http://localhost:3040/en/register

### Steps to test
Get on registration page: http://localhost:3040/en/register.
Try to fill up form without confirm password field.
Click on register button.

### Affected urls
http://localhost:3040/en/register

### DIFF
--- a/src/components/auth/register/RegisterForm.tsx
+++ b/src/components/auth/register/RegisterForm.tsx
@@ -21,6 +21,7 @@ export type RegisterFormData = {
   lastName: string
   email: string
   password: string
+  'confirm-password': string
   terms: boolean
   gdpr: boolean
 }
@@ -33,7 +34,10 @@ const validationSchema: yup.SchemaOf<RegisterFormData> = yup
     lastName: name.required(),
     email: email.required(),
     password: password.required(),
-    'confirm-password': yup.string().oneOf([yup.ref('password')], 'validation:password-match'),
+    'confirm-password': yup
+      .string()
+      .required('validation:password-match')
+      .oneOf([yup.ref('password')], 'validation:password-match'),
     terms: yup.bool().required().oneOf([true], 'validation:terms-of-use'),
     gdpr: yup.bool().required().oneOf([true], 'validation:terms-of-service'),
   })
@@ -43,6 +47,7 @@ const defaults: RegisterFormData = {
   lastName: '',
   email: '',
   password: '',
+  'confirm-password': '',
   terms: false,
   gdpr: false,
 }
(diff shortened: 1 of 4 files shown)


ani-kalpachka (approved):

ani-kalpachka (changes requested):
Do you think it would be better to use confirmPassword in order to avoid passing as string?

src/components/auth/register/RegisterForm.tsx
```diff
@@ -33,7 +34,10 @@ const validationSchema: yup.SchemaOf<RegisterFormData> = yup
     lastName: name.required(),
     email: email.required(),
     password: password.required(),
-    'confirm-password': yup.string().oneOf([yup.ref('password')], 'validation:password-match'),
+    'confirm-password': yup
```
kachar, line 37: Let's use `confirmPassword` instead of kebab cased name. Just like `lastName`

kachar: See 

https://github.com/podkrepi-bg/frontend/blob/df9850f26818e8a6b7c6125285920ace898e0d04/src/components/auth/changePassword/ChangePasswordForm.tsx#L29-L33

kachar: @jozef-drabik Since we use this functionality in more than one place it's gonna be appropriate to extract it to the common validators at: https://github.com/podkrepi-bg/frontend/blob/8dc9da1b0ca7994dfef6f07ce2402f972fa8df83/src/common/form/validation.ts

jan-fbc: i used camelCase now for confirmPassword as well and functionality is extracted in form/validation.ts

jozef-drabik: i used camelCase now for confirmPassword as well and functionality is extracted in form/validation.ts

kachar:
@jozef-drabik The build seems to be failling in the CI since there's one more register form that needs to be adjusted:

https://github.com/podkrepi-bg/frontend/blob/a19a6c6db21a25bb4c47242d0224fe3db2818f89/src/components/one-time-donation/RegisterDialog.tsx#L22-L29

Looks like we need to add the `confirmPassword` validation there as well.

We also need to add one additional field:

```tsx
<PasswordField name="registerPassword" autoComplete="new-password" />
<PasswordField name="confirmPassword" autoComplete="confirm-password" />
```

jozef-drabik:
@kachar I am sorry, it should work now. I ran build to see if there are more errors.

kachar (approved):

Issue #1212, opened 2022-11-29
[Feature] Improve fonts in the frontend


kachar:
**Is your feature request related to a problem? Please describe.**

We use different types of fonts all over the app. We need to switch to a single standard - `Montserrat` 

**Describe the solution you'd like**

- [ ] Remove custom `fontFamily` properties like `Lato` and others if any.
- [ ] Use native [fonts approach available with Next 13 ](https://nextjs.org/blog/next-13#nextfont)
- [ ] Enable [font optimization](https://nextjs.org/docs/basic-features/font-optimization)

PR #1223, opened 2022-12-02
1212 Added Montserrat as a global font family

AnzheloD:
Removed all other fonts
Added next/font from nextjs 13, which has automatic optimizations

Closes #{1212}
Issue: [https://github.com/podkrepi-bg/frontend/issues/1212](url)


kachar (approved):
@AnzheloD The contribution looks awesome! Kudos for the energy spent on this :zap:

kachar:
@AnzheloD  There's one warning when running the dev server

```
warn  - The @next/font/google font Montserrat has no selected subsets. Please specify subsets in the function call or in your next.config.js, otherwise no fonts will be preloaded. Read more: https://nextjs.org/docs/messages/google-fonts-missing-subsets
```

https://nextjs.org/docs/messages/google-fonts-missing-subsets

Let's fix that prior the merge

src/common/theme.ts
```diff
@@ -7,7 +7,12 @@ import {
   ThemeOptions,
 } from '@mui/material/styles'
 
-const fontFamily = 'Montserrat'
+import { Montserrat } from '@next/font/google'
+
+export const montserrat = Montserrat({
+  display: 'swap',
+  subsets: ['latin'],
+})
```
kachar, line 15: @AnzheloD I suggest that we also add `cyrillic` to the subsets and switch to `auto` display instead of `swap` as it was jumping between the fonts on local.

```suggestion
export const montserrat = Montserrat({
  display: 'auto',
  subsets: ['latin', 'cyrillic'],
})
```

https://user-images.githubusercontent.com/893608/205339057-ab1a3ece-e3a6-4694-9a22-31c8ddb18159.mp4

AnzheloD: I had a problem yesterday with display:auto, it was loading only the system fonts, that was the reason to choose swap, but it seems to work fine now

kachar: I think that happens if the server is not restarted after the font is applied. Thanks for the confirmation.

PR #1228, opened 2022-12-02
[Testing] E2e tests donation by bank and campaigns

dimitur2204:
Does not have an issue, but might be a good idea to create some general epic for the testing?

## Motivation and context

Added more tests to the donation flow and the campaigns page.
There isn't much to see, except one bad workaround of clipboard testing.

Would love to get feedback from the QA team and if they have suggestions for improvements it would be nice.
Can't add @ovcharski (https://github.com/ovcharski) as a reviewer so I hope he sees that


kachar:
Cc @ovcharski

kachar (approved):
@dimitur2204 the app changes look appropriate. When the tests are green we may proceed. 

Kudos for the energy

Issue #1124, opened 2022-11-18
[Bug] Email field on the Login page is not trimmed automatically


valkirilov:
**Describe the bug**
When you enter your email on the "**Login**" page and you accidentally put a leading/trailing space, it results into failed login because of invalid credentials. On most of the websites it's a common practice to ignore those "false" spaces when verifying the credentials, so we may apply it as well.

_Note: It can be considered as a change request, not a bug._
_Note: The same behavior is observed for the password field as well, but the safe approach is to apply this only for the "E-mail" field._

**To Reproduce**

1. Go to 'https://dev.podkrepi.bg/login'
2. Enter valid email for an existing account in the "**Email**" field, but put one additional space in the beginning/end of it
3. Enter valid password in the "**Password**" field _(valid password for the account matching the above email)_
4. Click on the "**Login**" button and see the following error - "**Wrong username / password.**"

**Expected behavior**
Instead of showing a misleading error message, we should trim the spaces from the "**E-mail**" field automatically, which will result into a successful login action.

**Screenshots**
![image](https://user-images.githubusercontent.com/1233496/202757569-a5958a53-2b52-46b2-aae2-ad35aa6a19e6.png)
![image](https://user-images.githubusercontent.com/1233496/202758911-f91e9b83-3145-493c-a8f1-27adf3d6174c.png)

**Additional context**
This bug is not dependent on any desktop/mobile/OS.

kachar:
I can confirm the problem. We need to add `trim()` to the yup validation chain for the email and other fields. We also need to apply it to all authentication forms, not only login forms.

Thanks for bringing it up @valkirilov

Nnachevvv:
I would like to take this bug :)

PR #1232, opened 2022-12-06
Fix fields in auth package not being trimmed automatically

Nnachevvv:
When you enter email, firstName or lastName in one of the form in authorisation package and you accidentally put a leading/trailing space, it results into failed login because of invalid credentials. On most of the websites it's a common practice to ignore those "false" spaces when verifying the credentials, so we may apply it as well.

Closes https://github.com/podkrepi-bg/frontend/issues/1124

## Motivation and context

On most of the websites it's a common practice to ignore those "false" spaces when verifying the credentials, so we may apply it as well.

In order to fix that I had added trim to the end of the strings.

### Note
Unfortunately, in the issue it was suggested that we should add trim to the yup validation chain to fix the problem. This didn't work and it seems that there is a incompatibility between Formik and Yup when there is a transformation. More information can be found here :
https://github.com/jaredpalmer/formik/issues/473
https://stackoverflow.com/questions/60423402/automatically-trim-white-spaces-with-yup-and-formik

## Screenshots:

Screenshot 2022-12-06 at 20 45 07

### Affected Forms

* LoginForm
* RegisterForm
* ForgottenPasswordForm
* UpdateNameModal
* UpdateEmailModal

### DIFF
--- a/src/components/auth/forgottenPassword/ForgottenPasswordForm.tsx
+++ b/src/components/auth/forgottenPassword/ForgottenPasswordForm.tsx
@@ -43,6 +43,7 @@ export default function ForgottenPasswordForm({
   const onSubmit = async (values: ForgottenPasswordForm) => {
     try {
       setLoading(true)
+      values.email = values.email.trim()
       await mutation.mutateAsync(values)
     } catch (error) {
       console.error(error)
--- a/src/components/auth/login/LoginForm.tsx
+++ b/src/components/auth/login/LoginForm.tsx
@@ -44,7 +44,7 @@ export default function LoginForm({ initialValues = defaults }: LoginFormProps) 
       setLoading(true)
 
       const resp = await signIn<'credentials'>('credentials', {
-        email: values.email,
+        email: values.email.trim(),
         password: values.password,
         redirect: false,
       })
--- a/src/components/auth/profile/UpdateEmailModal.tsx
+++ b/src/components/auth/profile/UpdateEmailModal.tsx
@@ -87,6 +87,7 @@ function UpdateEmailModal({
         throw new Error('Invalid login')
       }
 
+      values.email = values.email.trim()
       const updateUser = await mutation.mutateAsync({
         ...person,
         email: values.email,
--- a/src/components/auth/profile/UpdateNameModal.tsx
+++ b/src/components/auth/profile/UpdateNameModal.tsx
@@ -91,8 +91,8 @@ function UpdateNameModal({
 
       const updateUser = await mutation.mutateAsync({
         ...person,
-        firstName: values.firstName,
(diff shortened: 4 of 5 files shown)


kachar (changes requested):
@Nnachevvv Usually we handle the trimming functionality as part of the input validation, so it can be reused and clear in terms of responsibility. 

I'd suggest that we undo the trim for individual fields and add `yup.string().trim()` method to the forms if not present.

If this trimming method doesn't work correctly let's investigate why.

https://github.com/podkrepi-bg/frontend/blob/196cab3b87931a1fc6e59f7733f48fc88522fcc3/src/common/form/validation.ts#L68-L71

For the locations that do not use the generic validators let's use them.

Check the docs at https://github.com/jquense/yup#stringtrimmessage-string--function-schema

kachar:
@Nnachevvv Oh that's so sad for Formik. Thanks for sharing this investigation.

One idea that comes to me is to use a custom transform method globally. 

```tsx
yup.string().transform((value, originalvalue) => value.trim())
// or even
yup.string().transform(trim)
```

Do you mind checking if this can help us solve the problem?

If it does then we can add a new helper for the transformation and use that instead of the standard `trim()`

kachar:
We can also add a custom method or override the `trim` method


```tsx
yup.addMethod(yup.string, 'trim', (value) => value.trim()) // override?
yup.addMethod(yup.string, 'realTrim', (value) => value.trim()) // custom method
```

Nnachevvv:
@kachar I have tested both ways to do this and they didn't work. 
Also I took a look again into the github issue reported for this problem. It seems that formik supports only validations, all transformation such as trim(), lowercase(), uppercase() will not work. I ran this code to test your ideas:
```js
function realTrim() {
  return this.transform(function (value) {
      return value.trim();
  });
}

Yup.addMethod(Yup.string, 'realTrim', realTrim);

const SignupForm = () => {
  return (
    <>
      <h1>Subscribe!</h1>
      <Formik
        initialValues={{
          firstName: "",
          lastName: "",
        }}
        validationSchema={Yup.object({
          firstName: Yup.string().transform((value, originalvalue) => value.trim()),
          lastName: Yup.string().realTrim()
        })}
        onSubmit={async (values, { setSubmitting }) => {
          console.log(values)
          await new Promise(r => setTimeout(r, 500));
          setSubmitting(false);
        }}
```
Full demo: https://codesandbox.io/s/formik-v2-tutorial-final-forked-gzbx48?file=/src/index.js

kachar (approved):
Thanks for providing a fix @Nnachevvv 

Indeed it's a bit sad that the transformation functions and auto trimming didn't work. We might revise that at some point, but now it's important to proceed forward.

PR #1245, opened 2022-12-09
Admin panel - Export Donations to excel

### LINKED ISSUE #700: [Feature] Admin panel - Export Donations to excel
Export donation records from Donations Grid in Admin UI

Ideally, the solution should be reusable for the other grids in the UI too.

AndreyGoranov:
Closes #700
Goes together with a PR made on the API repo https://github.com/podkrepi-bg/api/pull/391

## Motivation and context

- Add a new button for downloading a file to the  Donations AppGrid with a translated title
- Add the new endpoint
- Update donation service to make requests to the new endpoint

## Screenshots:

The new UI change
image
image
Excel file with specific columns
image

## Testing
Tested by clicking on the icon and expecting to download a file in excel format
Tested with both API methods one with specified columns and one default with all available columns
Tested with no data ( It shows that there is a problem) and won't export an empty file

The new dependency is on the server side -> https://github.com/exceljs/exceljs

---|---|---|---


src/components/donations/grid/GridAppbar.tsx
```diff
@@ -16,6 +19,19 @@ const addIconStyles = {
...
+  const exportToExcel = useMutation({
+    mutationFn: useExportToExcel(),
+    onError: () => AlertStore.show(t('common:alerts.error'), 'error'),
+    onSuccess: (response) => {
+      const url = window.URL.createObjectURL(response.data)
+      const anchor = document.createElement('a')
+      anchor.href = url
+      anchor.download = 'Donations.xls'
+      anchor.dispatchEvent(new MouseEvent('click'))
+      window.URL.revokeObjectURL(url)
```
kachar, line 31: @AndreyGoranov We may extract this so it becomes a utility to download files. It might be useful if we do other exports besides donations.

```tsx
const downloadFile = (filename: string, data: string) => {
  const url = window.URL.createObjectURL(data)
  const anchor = document.createElement('a')
  anchor.href = url
  anchor.download = filename
  anchor.dispatchEvent(new MouseEvent('click'))
  window.URL.revokeObjectURL(url)
}

downloadFile('Donations.xls', response.data)
```

AndreyGoranov: Good point. Will extract it

kachar:
Waits for backend PR to be merged

src/components/donations/grid/GridAppbar.tsx
```diff
@@ -16,6 +20,14 @@ const addIconStyles = {
 export default function GridAppbar() {
   const router = useRouter()
   const { t } = useTranslation()
+  const exportToExcel = useMutation({
+    mutationFn: useExportToExcel(),
+    onError: () => AlertStore.show(t('common:alerts.error'), 'error'),
+    onSuccess: ({ data }) => {
+      downloadFile('Donations.xls', data)
```
igoychev, line 27: the extension should be xlsx, because the file contains xlsx and gives error when opened as xls

kachar: @igoychev Changed directly on master branch

Thanks for pointing out

PR #1250, opened 2022-12-12
1248 - add locale to stripe return urls

### LINKED ISSUE #1248: [Bug] English version is not preserved on redirection from Stripe after successful donation and error is shown when try to switch back to EN

Donations

### Describe the Bug

This bug has two sides which are probably connected. I'm not sure, so I'm logging it as one issue. 

1. When we start donating from EN version, fill in all the data in Stripe checkout page and make a successful payment, then we are redirected to the Donation page, but the language version is not preserved, we're redirected to the BG version. 
2. At that moment, if we try to change the version to EN in the Donation page -> Step `Payment`, then we get an error: `Unfortunately there was a problem!` and the following warnings into the browser console: 
- Fetching session returned null.
- XHR without token!!!
(Screenshot attached externally)
NOTE: This is reproduced during testing. It is not tried on production with a real donation. 

### To Reproduce

1. Make a donation while using the EN version
2. Fill all fields in Stripe checkout page and pay successfully 
3. Check the redirection after the donation

**Actual result**: The user is redirected to the BG version instead of the EN version. The user gets an error `Unfortunately there was a problem!` when trying to switch back to EN version. 
- Screenshots here: https://drive.google.com/drive/folders/1Qo2mV7BdfD5C8eB7Wrgi_c-x2wCTxRtW?usp=share_link

### Expected Behavior

The language version is preserved after the payment. The user should not see any errors. 

Chrome 108.0.5359.99

dimitur2204:
Closes #1248 

## Motivation and context
Added the language to the `returnUrls` of the stripe requests.

## Testing

Tried both failed and success states and they are now translated to the language you had set when you come back.

### Steps to test
Go through the donation flow to either succeed or fail.

### Affected urls
http://localhost:3040/en/campaigns/donation/:campaignSlug

### DIFF
--- a/src/components/one-time-donation/Steps.tsx
+++ b/src/components/one-time-donation/Steps.tsx
@@ -50,7 +50,7 @@ interface DonationStepperProps {
 }
 
 export default function DonationStepper({ onStepChange }: DonationStepperProps) {
-  const { t } = useTranslation('one-time-donation')
+  const { t, i18n } = useTranslation('one-time-donation')
   const router = useRouter()
   const success = router.query.success === 'true' ? true : false
   initialValues.amount = (router.query.price as string) || ''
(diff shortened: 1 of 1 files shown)


kachar (approved):
Thanks for the contribution @dimitur2204 


I was wondering if the default language is recognized when we provide it explicitly. I've tested that and it's all good. :gear:

Issue #1129, opened 2022-11-19
[Feature] Pagination on the messages on a Campaign


dimitur2204:
**Is your feature request related to a problem? Please describe.**
We need some way to limit the messages. We are fetching too many of them, there should be pagination

**Describe the solution you'd like**
Infnite scroll would be the best solution I think, but a simple pagination would also do! Might require some work on the api

Katzarov:
Is anyone working on this? If not, can I be assigned to it? I think it would be a good first full-stack issue for a newcomer like me.

PR #1252, opened 2022-12-13
1129 - Add pagination to donation wishes.

Katzarov:
Closes #[1129](https://github.com/podkrepi-bg/frontend/issues/1129)
Back-end PR https://github.com/podkrepi-bg/api/pull/397
## Motivation and context
 
Added simple pagination to donation wishes by campaign.
- Displays 12 items per page. (12 was picked almost arbitrarily)
- When there is only 1 page, the page picker/nav is not shown.
- When there a lot of pages, at most 6 page indices are displayed on the page picker/nav. (can be seen in the screenshots)
- When user clicks on a different page, the current data is shown while the one for the new page is being loaded.
- As can be seen, only success state is handled. If we have an error, we just don't display anything for the donation wishes.
- **This change is not compatible with currently deployed API**.  (I assumed making things backwards compatible is not a concern, and such changes are "just" deployed together)

Also, considered alternatives such as infinite scroll and load more.
- I believe we don't have an infinite scroll component we can already use. And I thought implementing one or adding a lib was not worth it.
- Was on the fence for load more.

## Screenshots:

image
image

## Testing
- Tested campaigns with no wishes and with 1 wish
- Tested cases where we go from no pages to 1 pages / from 2 pages to 3 pages. By adding a wish from a different instance of the app and letting react query refetch.
- Tested on iPhone SE (see screenshot).

### Affected urls

http://localhost:3040/campaigns/[slug]

none

none


Katzarov:
I should actually elaborate on the not compatible part. Unless both FE and API changes are deployed together, it results in a broken campaign page, and not just broken wish feature, which is not okay.
I am not familiar with the deployment process and how much control you have over such things. My question is, how would you generally handle such cases ?

src/components/campaigns/CampaignMessages.tsx
```diff
@@ -33,40 +35,62 @@ const StyledGrid = styled(Grid)(({ theme }) => ({
...
-                ? wish.person.firstName + ' ' + wish.person.lastName
-                : t('campaigns:donations.anonymous')}
-            </span>
-          </Grid>
-          <Grid item xs={12} pl={'33px'}>
-            <Typography>
-              <q>{wish.message}</q>
-            </Typography>
-            <Typography>{getExactDate(wish.createdAt, locale)}</Typography>
+      {isSuccess && numOfPages! > 1 && (
```
kachar, line 64: The intention here would be more explicit if we make it as:

```suggestion
      {isSuccess && numOfPages && numOfPages > 1 && (
```

kachar (approved):
@Katzarov Awesome progress! Kudos for the contribution!

I've allowed myself to add a few polishing touches to the visuals of the component.

- moved the pagination to the bottom since the user was unable to go to page 2 when he scrolls to the bottom of the list
- added smooth scroll to the top of the list when the page is changed
- used a more subtle color for the pagination instead of `primary`
- added emphasis on the author's name and date of post
- removed styled-components and used `<Stack />` and `<Grid2 />` for the layout
- renamed `CampaignMessages` component to `DonationWishes` to better fit the context


Desktop|Mobile
---|---
image|localhost_3040_campaigns_est-ad-et

Issue #1194, opened 2022-11-27
[Feature] Create Playwright Page Object Model for the E2E tests


DanielTakev:
**Currently we don't have any Page Object Model (POM) for our E2E tests with Playwright**
Creating POM would help to create and maintain the E2E tests for the application.

**What should be the structure?**
The structure should represent the base structure of our web pages. The idea behind POM is to define test classes with methods for each web page and in this way, we reuse selectors and methods, which help to avoid duplicated code. 
* More details here: https://playwright.dev/docs/pom

Issue #1188, opened 2022-11-25
[Dev] Update Playwright to latest


ovcharski:
**Is your feature request related to a problem? Please describe.**
 We are using playwright v1.24.2 while latest is v1.28.1

**Describe the solution you'd like**
Since Playwright is an npm package, you update the Playwright to the latest or specific version using the npm update or install command. There are several variations of updating the npm package: [How to Update Playwright Version](https://www.programsbuzz.com/article/how-update-playwright-version).

[Playwright Changelog](https://github.com/microsoft/playwright/releases)

PR #1265, opened 2022-12-22
#1188, #1194 - Add POM for the E2E tests and update all tests

dimitur2204:
> 
> 
> 
> Closes #{#1188, #1194 }
> 
> ## Motivation and context
> This change is important because we need stable E2E tests and reusable code. 
> 
> 1. What we have:
> - Page Object Model
> - Separated dependencies into the E2E folder
> - Reusable methods which work both with EN and BG language versions
> - Test reports + screenshots on failure and other details
> - Additional: removed Cyrillic symbol (`e`) from a few files (ex.: public/locales/bg/common.json)
> 
> 2. If we want to run local E2E tests:
> - Just change one method in the BeforeAll section of the desired spec and the test will run against `localhost` 
> 
> ### New or updated dependencies:
> - All Playwright dependencies were updated and moved into folder `e2e`
> 
> ### TODO:
> - Check with the developers all CI/CD configurations 
> - Fix all warnings from the pipeline
> - Update what's left from the previous specs

This is a PR based on https://github.com/podkrepi-bg/frontend/pull/1259

### TODO:
- [x] Check with the developers all CI/CD configurations 
- [x] Add a way to get around the different content on staging and local
- [x] Fix all warnings from the pipeline
- [x] Update what's left from the previous specs

## Environment

- Created a new yarn project inside of the e2e folder which needs to install its dependencies seperately.
- Added a `tsconfig.build.json` which ignores the e2e project. It is needed because when `next build` is called typescript checked the e2e folder with the missing dependencies and was throwing an error.
- Added a `STAGING` environment variable, which can be passed to the `yarn run test:e2e STAGING=true` to run the tests on `dev.podkrepi.bg` instead of locally.

- [x] `STAGING`: Env variable to toggle the tests on staging, only present in the `e2e` project folder

Dependency name|Version|Details
---|---|---|---
`@playwright/test`|`v1.29.1`|This dependency is for the `e2e` project|


dimitur2204:
Currently some of the tests are dependent on the content on staging. Will have to change that first

DanielTakev:
We should decide who we want to test against dev and localhost. 
As far as I understood, the test data for dev env. is different, so we could just open the first available campaign and make validations against it. 
But in that case, is possible to miss some bugs (for example - we can not expect any particular campaign title or URL, so this data validation will be skipped).

kachar:
We can also add some seed data that is not random, so we can check it with the e2e tests. It will lack variety but it will help us ensure the smoke tests are running OK.

dimitur2204:
I am unsure that even testing with hardcoded data like that is the best idea, maybe somehow mimicking the user behaviour with clicking on the first campaign card on the homepage, instead of searching for it by a hardcoded `title` is a better idea.
For now it works though.

imilchev:
Why do we want to test against dev.podkrepi.bg? That sounds like something that would cause problems. We don't have control over that environment. Deployments are happening constantly which means that tests against that env can be flaky...

DanielTakev (approved):

DanielTakev:
I had some gaps, but now I think the latest fixes and improvements are great and everything works fine on localhost and staging environment! 
Thanks, @dimitur2204!  👏 🥇 

@igoychev, @kachar, I hope you will find time for reviews and comments. 

We still have a long way to go, but I think this first step is very good! 💯

DanielTakev:
> Why do we want to test against dev.podkrepi.bg? That sounds like something that would cause problems. We don't have control over that environment. Deployments are happening constantly which means that tests against that env can be flaky...

I don't think so! 
If we have a control over our `dev` environment, we won't have any flaky tests. Currently our tests are very stable. 

It is very important to have stable tests against `dev`, because in some cases the developers could need to check the execution against `dev` and to compare it with their local env. 

Also would be better to have a scheduled execution against `dev` to monitor the results.

dimitur2204:
> Why do we want to test against dev.podkrepi.bg? That sounds like something that would cause problems. We don't have control over that environment. Deployments are happening constantly which means that tests against that env can be flaky...

@imilchev 

That is how tests were set up when I started work on them. With both a staging and local suite of tests, I imagine that's why @DanielTakev continued with that also.

I also agree that it is a little strange. Maybe we can run staging tests only before a release or something of the sort?

igoychev:
> I had some gaps in my PR, but now the latest fixes and improvements are great and everything works fine on localhost and staging environment! Thanks, @dimitur2204! 👏 🥇
> 
> @igoychev, @kachar, I hope you will find time for reviews and comments.
> 
> We still have a long way to go, but I think this first step is very good! 💯

Hey @DanielTakev @dimitur2204 great improvement overall and setting the stage for more cool stuff to come on the automation front!!! I'll check and hopefully merge in the next few days.

e2e/playwright.config.ts
```diff
@@ -0,0 +1,109 @@
...
+      mode: 'only-on-failure',
+      fullPage: true,
+    },
+    video: {
+      mode: 'retain-on-failure',
+      size: {
+        width: 1500,
+        height: 900,
+      },
+    },
```
igoychev, line 44: let's collect video only on demand as it will be needed in very rare cases anyway

e2e/tests/regression/donation-flow/anon-donation-custom.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+  'Anonymous contributor is able to donate custom amount - BG language version',
+  async () => {
+    let page: Page
+    let homepage: HomePage
+    let headerPage: HeaderPage
+    let campaignsPage: CampaignsPage
+    let donationPage: DonationPage
+    let stripeCheckoutPage: StripeCheckoutPage
+    // For the URL is used RegExp so the tests will pass for each environment - localhost, dev, etc.
+    const firstCampaignPageUrl = `/campaigns/${campaignData[0].slug}`
```
igoychev, line 24: [change requested]
the first campaign should come by selecting from the list of existing campaigns, because the predefined hard-coded title or slug which may not exist in all environments

e2e/tests/regression/donation-flow/anon-donation-custom.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      await homepage.navigateToEnvHomepage()
+    })
+
+    test.afterAll(async () => {
+      await page.close()
+    })
+
+    test('Particular campaign can be opened through the Campaign page', async () => {
+      await headerPage.clickDonateHeaderNavButton()
+      await campaignsPage.clickCampaignSchoolChildrenVarna()
```
igoychev, line 49: this should be clickOnFirstCampaign() to make the test work on every environment

e2e/tests/regression/donation-flow/anon-donation-custom.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      // We move from the common Campaigns page to the particular campain page
+      await campaignsPage.checkPageUrlByRegExp(firstCampaignPageUrl)
+      expect(
+        await campaignsPage.isSchoolChildrenVarnaHeadingVisible(),
```
igoychev, line 53: this should be "isCampaignTitleVisible" instead of checking for particular text

e2e/tests/smoke/smoke-campaigns.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      .soft(
+        await campaignsPage.isCampaignsHeadingVisible(),
+        'The main Campaigns heading is not visible.',
+      )
+      .toBeTruthy()
+    expect(
+      await campaignsPage.isSupportCauseTodayHeadingVisible(),
+      'The Support Cause Today heading is not visible.',
+    ).toBeTruthy()
+  })
```
igoychev, line 45: the test should check if at least one campaign is visible, instead of checking for a particular campaign

next.config.js
```diff
@@ -15,6 +15,10 @@ const moduleExports = {
   sassOptions: {
     includePaths: [path.join(__dirname, 'src/styles')],
   },
+  typescript: {
+    tsconfigPath: 'tsconfig.build.json',
+  },
+  tsconfigPath: 'tsconfig.build.json',
```
igoychev, line 21: could you check why this line gives a warning:
```
warn  - Invalid next.config.js options detected: 
  - The root value has an unexpected property, tsconfigPath, which is not in the list of allowed properties (amp, analyticsId, assetPrefix, basePath, cleanDistDir, compiler, compress, crossOrigin, devIndicators, distDir, env, eslint, excludeDefaultMomentLocales, experimental, exportPathMap, generateBuildId, generateEtags, headers, httpAgentOptions, i18n, images, onDemandEntries, optimizeFonts, output, outputFileTracing, pageExtensions, poweredByHeader, productionBrowserSourceMaps, publicRuntimeConfig, reactStrictMode, redirects, rewrites, sassOptions, serverRuntimeConfig, staticPageGenerationTimeout, swcMinify, trailingSlash, typescript, useFileSystemPublicRoutes, webpack).
```

igoychev (changes requested):
Looks great overall! 

However please let's make the campaign tests to not depend on hardcoded campaign titles, so that they can run on any environment. For example the tests could select one of the existing campaigns and check the functionality using it - see specific comments inline.

e2e/data/campaigns-env.data.ts
```diff
@@ -0,0 +1,25 @@
...
+const stagingCampaignData: CampaignData = [
+  {
+    title: 'Училище за деца с нарушено зрение гр. Варна - стая за ерготерапия',
+    slug: 'uchilishe-za-deca-s-narusheno-zrenie-gr-varna-staya-za-ergoterapiya',
+  },
+  {
+    title: 'Кризисен център за пострадали от насилие - шанс за нов живот',
+    slug: 'krizisen-centur-za-postradali-ot-nasilie-shans-za-nov-zhivot',
+  },
+]
```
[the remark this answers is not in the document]
DanielTakev: Hello and thanks! 
My general assumption was that we have overall test data control, but in consequence I understood the situation. Actually I refactored all tests checking only the Dev environment (that slowed me down to realize how the project works in general). 

Regarding the tests dependencies - the best case is to expect particular data and to verify the exact values of each heading/paragraph etc., because this will be way more accurate assertion. 
Considering the situation, we the alternative is update the tests just to check the first campaign result as you suggested. 

Next couple of days I will be out of the country, so if Dimitar has the opportunity - he could make the change. 
Otherwise I will do it (maybe) next week.

e2e/playwright.config.ts
DanielTakev: If we delete all artifacts after each execution I don't think they would be a problem. Playwright can record videos using one of the following options: 

- 'off' - Do not record video.
- 'on' - Record video for each test.
- 'retain-on-failure' - Record video for each test, but remove all videos from successful test runs.
- 'on-first-retry' - Record video only when retrying a test for the first time. 

We could just turn the option OFF and use only screenshots.

igoychev: yep, my idea was by default to be "off" and only when investigating a tricky case to switch to "retain-on-failure"

e2e/tests/regression/donation-flow/anon-donation-custom.spec.ts
dimitur2204: Check https://github.com/podkrepi-bg/frontend/pull/1265/commits/ee7116a07564a6b0f11c134add3a765966acc3fa

e2e/tests/regression/donation-flow/anon-donation-custom.spec.ts
dimitur2204: Check https://github.com/podkrepi-bg/frontend/pull/1265/commits/ee7116a07564a6b0f11c134add3a765966acc3fa

e2e/tests/regression/donation-flow/anon-donation-custom.spec.ts
dimitur2204: Check https://github.com/podkrepi-bg/frontend/pull/1265/commits/ee7116a07564a6b0f11c134add3a765966acc3fa

e2e/playwright.config.ts
dimitur2204: https://github.com/podkrepi-bg/frontend/pull/1265/commits/18ec68e1f3646d3f63bc12e297caee77f44f5eef

e2e/tests/smoke/smoke-campaigns.spec.ts
dimitur2204: Check https://github.com/podkrepi-bg/frontend/commit/ee7116a07564a6b0f11c134add3a765966acc3fa

e2e/data/campaigns-env.data.ts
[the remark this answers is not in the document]
dimitur2204: Changed in https://github.com/podkrepi-bg/frontend/commit/ee7116a07564a6b0f11c134add3a765966acc3fa

next.config.js
dimitur2204: [1194 - remove uneccessasry tsconfigPath inside of the config](https://github.com/podkrepi-bg/frontend/pull/1265/commits/56023e4eeaf564c194670bca622ea67e4c0886fa)

igoychev (approved):
great for making the tests run on any environment!

kachar (approved):
Nicely done :clap:

PR #1301, opened 2023-01-24
Payment Card Responsiveness

### LINKED ISSUE #1278: Campaign card image responsiveness

_No response_

As as donor I would like to see images responsive so I can see it on different devices desktop, tablet, mobile equally good

Screenshot_2023-01-07-22-11-14-560_com opera browser~2

Responsive image containers 

_No response_

pshumareva:
Closes #1278 

## Motivation and context

Better representation
 The width expansion of campaign images
 With the ObjectFit property

## Screenshots:

Before|After
Screenshot_1
Screenshot_2

---|---|---|---


ani-kalpachka (changes requested):
If we apply objectFit: 'cover' to the campaign card images they will be full width but will be cut. Here are some examples:

Screenshot 2023-01-25 114609

Screenshot 2023-01-25 114504

Screenshot 2023-01-25 114701

A possible solution is to adapt the height of the images too and add minHeight to the CardMedia. @pshumareva Could you please check it? I will spend time to investigate too.

pshumareva:
@ani-kalpachka Is it possible to increase the height of the image container (to maybe 400px)? That way we can keep cover (for the width) and have close to full images.

kachar:
@ani-kalpachka Can you please recheck the PR and comments as you initially requested changes?

ani-kalpachka:
@pshumareva could you implement and test your idea and validate if it would be a solution for all devices? Have in mind that we need to avoid cutting the images.

pshumareva:
@ani-kalpachka sure thing!

sashko9807:
Technically speaking you don't need `object-fit` property, if your idea is to fit the image to the size of the parent container. The problem is that one could stretch a low res image just so far without it starting to look weird. 
Perhaps a better approach to this problem would be to use fixed width with breakpoints, rather than responsive values. This way the card will still remain responsive towards different resolutions, but due to the fixed width the card wont get wide enough to make the image look weird.

I leave the necessary links to the MUI documentation if you're interested to trying out this idea
[Default breakpoints](https://mui.com/material-ui/customization/breakpoints/)
[Applying breakpoints to component via sx property](https://mui.com/system/getting-started/usage/#responsive-values)

igoychev:
ping! 
@pepashumareva would you have time to complete this one?

pshumareva:
@igoychev yes, im working on it

README.md
```diff
@@ -99,8 +99,13 @@ Watch releases of this repository to be notified about future updates:
 
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
 
+<<<<<<< HEAD
+[![All Contributors](https://img.shields.io/badge/all_contributors-64-orange.svg?style=flat-square)](#contributors-)
+=======
 [![All Contributors](https://img.shields.io/badge/all_contributors-65-orange.svg?style=flat-square)](#contributors-)
 
+> > > > > > > master
+
```
dimitur2204, line 108: Merge conflict artifacts here

src/components/client/campaigns/CampaignCard.tsx
```diff
@@ -57,6 +57,30 @@ const StyledCard = styled(Card)(({ theme }) => ({
...
+    },
+    '@media(min-width: 600px)': {
+      height: 220,
+    },
+    '@media(min-width: 900px)': {
+      height: 262,
+    },
+    '@media(min-width: 1200px)': {
+      height: 175,
+    },
```
dimitur2204, line 83: I think that we should fix the `minWidth` of the cards and then adjust the grid to keep cards relatively close to that `minWidth` so that the images stay in similar aspect ratio on most resolutions and then agree on some a height for the images themselves.

We should also probably communicate that to the people uploading the images, that they should upload the featured images in some specific aspect ratio.

Could you try that?

dimitur2204 (changes requested):

src/components/client/campaigns/CampaignCard.tsx
sashko9807: > I think that we should fix the `minWidth` of the whole card and then adjust the grids that render the cards to keep them relatively close to that `minWidth` so that the images stay in similar aspect ratio on most resolutions. Then agree on some a height for the images themselves.

> We should also probably communicate that to the people uploading the images, that they should upload the featured images in some specific aspect ratio.

Overlooked some of MUI's capabilities when I posted the above suggestion, but perhaps the best approach here would be to just, add breakpoints for the number of columns to the grid layouts [link](https://mui.com/material-ui/react-grid/#responsive-values). This way we could maintain cards width easily.

As for the image's aspect ratio we could just, replace the `height:200` property  from `classes.media` with  `width: '100%`. Then we can pass aspectRatio property, to the [`<div> inside <CardMedia>` component](https://github.com/podkrepi-bg/frontend/blob/master/src/components/client/campaigns/CampaignCard.tsx#L132). This will automatically adjust the height of the img container, to the width of the card,  just so the set aspectRatio is maintained.

kachar: You can also check the native CSS attribute `aspect-ratio`

https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio

ani-kalpachka:
I will close this PR because we have a new design for the campaign cards where this issue is not present. The development is in progress.

Issue #1302, opened 2023-01-25
New home page design 


swolf86:
### Describe the feature you'd like to request

![home page](https://user-images.githubusercontent.com/75036745/214520586-039e324b-79fc-4121-a15f-3995a698edd9.png)
New home page design 

Facebook life feed and preview 

### Describe the solution you'd like

Interactions to be described here

marianacat:
You can find the home page and its descriptions in a Figma file here:

https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=16874%3A51853&t=2OlP42xobsSn5x1J-1

swolf86:
![image](https://user-images.githubusercontent.com/75036745/214974501-78039036-10ab-45d2-913f-6e1ebf1bd0dc.png)
Description on interactions and design notes in figma
https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=16857%3A51852&t=96Bw09P0HuENRQq7-1

swolf86:
in pink

swolf86:
Can we A/B split test before we launch to see differences using google optimize

Issue #1227, opened 2022-12-02
[Bug] Campaigns listing page is not full width


ani-kalpachka:
**Describe the bug**
Campaigns listing page is not full width.

**To Reproduce**
1. Go to https://podkrepi.bg/campaigns
2. Observe the spacing around the campaigns listing section.

**Expected behavior**
Campaigns listing section has to be full width identical to the campaigns listing section on Homepage.

**Screenshots**
Campaigns page:
![image](https://user-images.githubusercontent.com/14351733/205310862-5ab014d0-8a90-4517-8d39-705baddc22ad.png)

Homepage:
![image](https://user-images.githubusercontent.com/14351733/205311115-9816d348-c232-43fb-8f34-ecb1ee6edc18.png)

kachar:
@ani-kalpachka I think it's intentional to have a max width of the container. I would personally reduce the width of the homepage to a certain size. On a big screen like 24" it looks really annoying.

kabaivanov:
> @ani-kalpachka I think it's intentional to have a max width of the container. I would personally reduce the width of the homepage to a certain size. On a big screen like 24" it looks really annoying.

Double that, here's a screenie from my 27k iMac at it's default 5k resolution.
<img width="2560" alt="Screenshot 2022-12-03 at 22 02 00" src="https://user-images.githubusercontent.com/118026471/205459708-b940d386-13f0-4edf-bfd3-eca47a8e8439.png">

@ani-kalpachka if you see what we mean, let's rename the ticket name and just fix the homepage instead.

swolf86:
I think we can adapt the code that photos střech also title leaves a lot of space below
The benefit of having full width is that you use the whole real estate and see bigger more impactful cards  with photos 
Can we try to find optimal solution

Btw it should say urgent campaigns as some of them are competed

kachar:
@swolf86 We're talking about the space on the side - if we make the content full-width of the page it will look cluttered and users will actually see less relevant with a broader focus area. 

What @kabaivanov and I are suggesting is that we use a narrower focus for the user and use the native up/down scroll to see more campaigns.

See for example how github are handling that with space on the sides

![image](https://user-images.githubusercontent.com/893608/205478949-2ce84d89-4854-4e91-b132-b0a6fd3a2bc5.png)

kachar:
Adding some preview for brevity

:two: `maxWidth: lg` | :one: `maxWidth: 100vw`
---|---
[...]

PR #1303, opened 2023-01-25
Fix Campaigns section max width and alignment on Homepage and Campaigns page

ani-kalpachka:
Fix Campaigns section max width and alignment on Homepage and Campaigns page to be identical and responsive for bigger screens.

## Screenshots:
Homepage:

Screenshot 2023-01-25 112939

Screenshot 2023-01-25 113003

Campaigns page:

Screenshot 2023-01-25 112759

Screenshot 2023-01-25 112732

### DIFF
--- a/src/components/campaigns/CampaignsList.tsx
+++ b/src/components/campaigns/CampaignsList.tsx
@@ -37,6 +37,7 @@ export default function CampaignsList({ campaignToShow }: Props) {
     <Grid
       container
       justifyContent="center"
+      spacing={4}
       sx={(theme) => ({
         width: `calc(100% + ${theme.spacing(1.5)})`,
         marginLeft: `-${theme.spacing(2.75)}`,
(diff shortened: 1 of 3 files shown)


dimitur2204:
I am personally not a fan of these long cards in your second and fourth images.
I think that we should have 3 on a row so they are more square-like.
Can we consult the design team @swolf86 ?

ani-kalpachka:
@dimitur2204 These changes are rather focused on the max width of the campaign section which we have both on homepage and campaigns listing page. It is currently not identical on both pages and its width is not limited for larger screens. You can see the discussion here: https://github.com/podkrepi-bg/frontend/issues/1227.

The cards are 4 by design for all devices. I haven't done any changes to the cards in this PR. I think your idea might be valid and we could optimize this. Can we discuss it with the design team and create a separate ticket for this if needed?

dimitur2204 (approved):

dimitur2204:
Considering the new home page should be coming in the near future we can probably not bother design with that https://github.com/podkrepi-bg/frontend/issues/1302

PR #1306, opened 2023-01-26
Fix the header on Campaigns listing page

ani-kalpachka:
## Screenshots:

Before|After
---|---
image|image


dimitur2204 (approved):
Thank you for the fast response! It should be good to go!

PR #1314, opened 2023-02-01
New Donation Flow - Alerts and Summary Column

### LINKED ISSUE #1286: New Donation Flow - Alerts Column on Desktop

Donations

We need to be able to add alerts and messages to the form on the right side of the from as in the design here.

Mockup: https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=16214%3A53322&t=3q3lUtCjopeDjYZP-0
Screenshot: 
image

We need to create somehow a way to not make a styling mess when adding a lot of these messages. Some kind of an API for creating them would be a good idea.

One way to do it might be to encapsulate a `<Step></Step>` component with a `startMessage` which will append to the right side of the step at the start and `endMessage` at the end. 
`<Step startMessage={<Alert>Important tax messageInformation</Alert>}>
    {StepContent}
 </Step>`

There are a lot of ways to accomplish this and some of them might be better.

dimitur2204:
Closes #1286 

### Form Improvements:
* Added validation for the Login and Register forms
* Moved validations and initial value objects to each respective step and part of the form
* Moved the PaymentIntent update to run only once on submission
* Changed form value names for the amount regarding card donation to make more sense
* Added an error object on the DonationFlowContext
* Fixed `otherAmount` field bug when there is an uneven amount of prices

### Alerts column:
* Added `AlertsColumn` which takes a `sectionsRefArray` which is an array of references. It then attaches alerts to each of the section references based on their id which is mapped to the `[alerts](https://github.com/podkrepi-bg/frontend/pull/1314/files#diff-0b9f0d19b2035c89f509f8bf65f75cbc4ced3b1209c50b243ebd57ae86c2ebe9R39)` 
* Added `AnchoredAlert` which is an alert that takes a `sectionRef` and is absolutely positioned with a top equal to `section.offsetTop + spacing and titles on the section  (theme.spacing(4) + 1.5rem)`
* Added `PaymentSummaryAlert` which is the summary of the payment, taxes and total of your payment.

### Desktop
Transaction|Auth
---|---
image|image

### Mobile
Transaction|Auth
---|---
image|image

## RFC
I think that this Anchoring approach might be problematic since React does not control refs and from my experience rerendering can be unpredictable.

If another viable approach comes to mind to anyone it would be good to hear @podkrepi-bg/frontend-maintainers 
### Affected urls

`campaign/donation-v2/{slug}`


src/components/donation-flow/DonationFlowForm.tsx
```diff
@@ -4,103 +4,97 @@ import { useElements, useStripe } from '@stripe/react-stripe-js'
...
+    ...amountValidation,
+    ...loginValidation,
+    ...registerFormValidation,
   })
 
 export function DonationFlowForm() {
   const { data: session } = useSession()
-  const { campaign, stripePaymentIntent } = useContext(DonationFlowContext)
+  const { campaign, stripePaymentIntent, paymentError, setPaymentError } =
+    useContext(DonationFlowContext)
```
kachar, line 89: We can add a helper hook to get the context in a seamless way with:

```tsx
export function useDonationFlow() {
  return useContext(DonationFlowContext)
}
```

```suggestion
  const { campaign, stripePaymentIntent, paymentError, setPaymentError } = useDonationFlow()
```

src/components/donation-flow/steps/Amount.tsx
```diff
@@ -94,19 +105,19 @@ export default function Amount() {
...
             style={
-              !mobile
+              !mobile && Number(prices?.length) % 2 === 0
                 ? {
                     float: 'right',
                     marginTop: -50,
                     width: '49%',
                   }
-                : { marginTop: theme.spacing(2) }
+                : { marginTop: theme.spacing(2), width: mobile ? '100%' : '49%' }
```
kachar, line 120: Can we use `vw` instead of `%` to match the viewport size?

src/components/donation-flow/DonationFlowForm.tsx
```diff
@@ -152,32 +171,54 @@ export function DonationFlowForm() {
...
+                    <Tooltip title="Ако дарете анонимно, данните ще останат недостъпни за бенефициента.">
+                      <IconButton color="primary">
+                        <Info />
+                      </IconButton>
+                    </Tooltip>
+                  </Box>
+                }
+                name="isAnonymous"
+              />
+              {JSON.stringify(paymentError)}
```
kachar, line 205: We might need to take this off in the final version. Mentioning it so we don't forget it

```suggestion
```

src/components/donation-flow/alerts/AlertsColumn.tsx
```diff
@@ -0,0 +1,96 @@
...
+      sx: {
+        display: payment ? 'flex' : 'none',
+      },
+    },
+    'select-authentication': {
+      color: 'info',
+      children: (
+        <Box>
+          <Typography>Избирайки да се впишете. ще можете да:</Typography>
+          <List
```
kachar, line 53: We can extract this list as separate component to simplify the rendering of alerts

kachar (approved):
Yet another swift update of the payment flow @dimitur2204 

I see strong progress with this one.
Shared some ideas for code reusability and simplification.

src/components/donation-flow/alerts/AlertsColumn.tsx
kachar: We might also expand the text and reduce the right padding/margin to contain more chars on single row

image

kachar:
We might improve the width of the different sections so they match, whichever width feels right for you, just to be consistent.

image

src/components/donation-flow/steps/Amount.tsx
dimitur2204: This is not matching the viewport size, but making sure the component is 50% because the Grid is with 2 columns on desktop and 1 on mobile. It is essentially a workaround, because we can't put this field inside of the Grid.

image
image

src/components/donation-flow/alerts/AlertsColumn.tsx
dimitur2204: Removed all on the list items, but kept the one on the `X` button row

dimitur2204:
> We might improve the width of the different sections so they match, whichever width feels right for you, just to be consistent.
> 
> image

Matched the smaller width, by hardcoding it to the AnchoredAlert

src/components/donation-flow/DonationFlowForm.tsx
dimitur2204: Done in [d45154f](https://github.com/podkrepi-bg/frontend/pull/1314/commits/d45154fb0795cde4f48321470dda0b24c2940d51)

kachar:
Awesome update!

Issue #1297, opened 2023-01-23
New Donation Flow - Succes and Fail pages


dimitur2204:
### Which area(s) of Podkrepi.bg are affected? (leave empty if unsure)

Donations

### Describe the feature you'd like to request

Create the Success(Thank you) and Fail pages for the new donation flow. 

Success mockups: https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=15829%3A47727&t=3Tx9ep5pUJGc6phS-0
<img width="456" alt="image" src="https://user-images.githubusercontent.com/61479393/214135888-0cde4501-cd8f-41f8-a1a1-33908906a4d7.png">

Fail mockups: TO BE ADDED

### Describe the solution you'd like

They shouldn't be separate routes in the `pages` folder, but just 2 components which we will later render based on the `redirect_status` that Stripe return after a donation.

We should have them as just components as it currently is in the `one-time-donation` flow.

Issue #1272, opened 2023-01-04
New Version of the Donation Flow 


dimitur2204:
### Which area(s) of Podkrepi.bg are affected? (leave empty if unsure)

Donations

### Describe the feature you'd like to request

This is an epic of the tasks that need to be done for the new Donation Flow.
We have the new design and flow of creating a donation here:
https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=16163%3A49451&t=d7IRlRGBEVbNyrbG-0

### Describe the solution you'd like

Subtasks should be listed here with respective PRs:

- [x] #1269
- [x] #1287
- [x] #1274
- [x] #1276
- [x] #1286
- [x] #1292
 #1298 
- [x] #1291
- [x] #1295
- [x]  #1299 
https://github.com/podkrepi-bg/api/issues/407
- [x] #1297 
- [ ] #1308 
- [x] #1408 
- [x]  #1317

PR #1322, opened 2023-02-05
New Donation Flow - #1297 #1291 Bank Transfer flow and Success page

### LINKED ISSUE #1291: New Donation Flow - Bank Transfer Content

Donations

We need the bank transfer information to be present when the user chooses the bank transfer information payment method.
Mobile: https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=16046%3A48652&t=isUvNTPmQtgx72VE-0
Desktop:https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=16211%3A49585&t=isUvNTPmQtgx72VE-0

Would be best if we reuse most of the functionality from the current form.

There is a Stripe SEPA Direct Debit transfers that might be an option to automatically detect such payments. This suggestion is very much "in the air" though. Has to be synced with a lot of people.
https://stripe.com/docs/payments/sepa-debit

dimitur2204:
Closes #1291, partially #1297

Big PR closing 2 issues and mostly finishing off  #1272. Error handling still needs to be discussed.

## Regarding #1291:

- Added the content for bank donations to this form.
- Disabled the amount field when doing a bank payment
- Added loading states to the form.
- Moved out the `createStripePaymentIntent` request to be done server side, since it was cleaner, but error handling will now be more challenging.

### Screenshots:

Mobile|Desktop
---|---
image|image

### Affected urls
`campaigns/donation-v2/{slug}`

## Regarding #1297:

Handling of post payment done following: https://stripe.com/docs/payments/accept-a-payment?platform=web&ui=elements&client=react#web-submit-payment

- Added a `topLevelAwait` experimental flag and moved out the `loadStripe` call to a separate service so now it can be imported as `import {stripe} from 'services/stripeClient.ts'`
- Added a new `campaigns/donation-v2/consequuntur-suscipit-voluptatibusheavily-funded/status` page. It handles the `paymentIntent` status and responds by either showing the "Success" content or returning to the donation form when errored currently*.
- Added a new common component `SocialShareListButton` which renders a button to list out common share links to social medias.
- Added a call to create a wish and disable the form afterwards.

## Full flow
Desktop|Mobile
---|---
donation-v2-full-flow|donation-v2-full-flow-mobile


src/pages/campaigns/donation-v2/[slug]/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    queryFnFactory<Stripe.Price[]>(),
+  )
+
+  //create and prefetch the payment intent
+  const { data: paymentIntent } = await apiClient.post<
+    Stripe.PaymentIntentCreateParams,
+    AxiosResponse<Stripe.PaymentIntent>
+  >(endpoints.donation.createPaymentIntent.url, {
+    amount: 100,
+    currency: Currencies.BGN,
```
Comment (kachar) [inline src/pages/campaigns/donation-v2/[slug]/index.tsx:33]: Adding a note here to keep in mind we have hardcoded values for the amount.

src/components/donation-flow/DonationFlowStatusPage.tsx
```diff
@@ -0,0 +1,201 @@
...
+          </Typography>
+        </CardContent>
+      </CardActionArea>
+    </Card>
+  )
+}
+
+export default function DonationFlowStatusPage({ slug }: { slug: string }) {
+  const { data, isLoading } = useViewCampaign(slug)
+  if (isLoading || !data) return <CenteredSpinner size="2rem" />
```
kachar, line 53: We break the rules of hooks with an early return before some of the hooks. It's better to create a component wrapper that will render this component only in certain conditions.

src/components/donation-flow/DonationFlowForm.tsx
```diff
@@ -89,82 +90,102 @@ export const validationSchema: yup.SchemaOf<DonationFormDataV2> = yup
...
+          return
         }
 
+        // Confirm the payment
         const { error } = await stripe.confirmPayment({
           //`Elements` instance that was used to create the Payment Element
           elements,
           confirmParams: {
-            return_url: `${window.location.origin}/campaigns/donation-v2/${campaign.slug}`,
+            return_url: `${window.location.origin}/campaigns/donation-v2/${campaign.slug}/status`,
```
kachar, line 177: Maybe we can use `routes` here instead of hardcoded links. It will ease us when we will update the routes and remove the `v2` suffix

src/components/donation-flow/DonationFlowForm.tsx
```diff
@@ -89,82 +90,102 @@ export const validationSchema: yup.SchemaOf<DonationFormDataV2> = yup
...
       initialValues={{
         ...initialValues,
         email: session?.user?.email ?? '',
-        authentication: session?.user ? DonationFormDataAuthState.AUTHENTICATED : null,
+        authentication: session?.user ? DonationFormAuthState.AUTHENTICATED : null,
       }}
       validationSchema={validationSchema}
       onSubmit={async (values) => {
+        if (values.payment === DonationFormPaymentMethod.BANK) {
+          router.push(routes.campaigns.donationStauts(campaign.slug), {
```
kachar, line 113: Typo in the fn name `donationStauts` -> `donationStatus`

kachar (changes requested):
Awesome upgrade @dimitur2204 

It's an insightful idea to use the top-level await for the Stripe loader. :bulb: 

The overall process looks and feels very smooth :pinched_fingers: 

I've added notes on a few minor details, please check them out

src/pages/campaigns/donation-v2/[slug]/index.tsx
Comment (dimitur2204) [reply src/pages/campaigns/donation-v2/[slug]/index.tsx:33]: Yeah this value is hardcoded, because this is the initial creation of the intent. 
Stripe doesn't let us create a PaymentIntent without a value, because their assumption is that you would be creating the payment intent after the amount has been chosen. And in our case we need it beforehand.

Comment (kachar) [reply src/pages/campaigns/donation-v2/[slug]/index.tsx:33]: That's an unusual flow for the payment intents, however, we need to document it so the purpose is clear in our project.

Comment (kachar) [reply src/pages/campaigns/donation-v2/[slug]/index.tsx:33]: We might also pick a value that's closer to the initial values that we offer to the user. In this way updating the PI won't be necessary if they really pick the default value.

Comment (dimitur2204) [reply src/pages/campaigns/donation-v2/[slug]/index.tsx:33]: > We might also pick a value that's closer to the initial values that we offer to the user. In this way updating the PI won't be necessary if they really pick the default value.

We can do that if we decide to have one of the values preselected.

src/components/donation-flow/DonationFlowStatusPage.tsx
dimitur2204: That is true. I will change it just in case.
Though considering this request is prefetched this never turns out to be a problem, because this condition is never true, even on first render. Otherwise I probably would have gotten the `rules-of-hooks` error

kachar (approved):
Nice enhancement @dimitur2204 We're a few steps closer towards the epic goal.

I'm glad to see that you've updated the general donation route now.

PR #1325, opened 2023-02-06
[911] Notify all users on successful donation

### LINKED ISSUE #911: [Feature] Message to every visitor of the site when a donation is made
In order to make the site visitors more inclined to make a donation a notification for each successful donation should be made. 
Considering that currently we have a huge backlog, developing a socket server to display this information sounds like we are far away from this feature - but then again this is the beauty of open-source projects. :)

AndreyGoranov:
Closes #911
Should go together with: https://github.com/podkrepi-bg/api/pull/431

## Motivation and context

Why is this change required?
-To be able to send notifications to users when a successful donation occurs

## Screenshots:
image

## Testing
Login with different multiple users from different clients and try making a donation if it's successful all other users should receive a notification.

You can try making multiple notifications at the same time. The expected behavior, for now, is to show them one by one whenever the first one closes you should see the next one

You can try refreshing the page and see if you will receive notifications after that
### Affected urls

New or updated dependencies: Socket io


src/components/faq/helpers/utils.ts
```diff
@@ -49,3 +49,14 @@ export function filterFaqQuestions(
...
+export const dateToTime = (dateString: string | undefined) => {
+  if (!dateString) {
+    return
+  }
+
+  const date = new Date(dateString)
+  const localTime = date.toLocaleTimeString()
+
+  return localTime
+}
```
dimitur2204, line 62: There is a `formatDateString` function in `src/common/util/date.ts` you can use

src/stores/NotificationStore.ts
```diff
@@ -0,0 +1,51 @@
+import { Slide, SnackbarProps, PaperProps } from '@mui/material'
+import { NotificationLayoutData } from 'components/layout/NotificationSnackBar/DonationNotificationLayout'
+import { makeAutoObservable } from 'mobx'
+import { enableStaticRendering } from 'mobx-react'
```
dimitur2204, line 4: I personally think that we should stray away from `mobx` and class based implementations and go for something hook based, considering the whole react ecosystem seems to be moving to hooks. Something like [zustand](https://docs.pmnd.rs/zustand/getting-started/introduction) I think would be more fitting for our stores, which is way more lightweight than `mobx` too 

Let's see the opinion of other people though.

src/components/layout/NotificationSnackBar/NotificationSnackBar.tsx
```diff
@@ -0,0 +1,61 @@
...
+            data={
+              notificationStack?.getNotifications.length
+                ? notificationStack?.notifications?.[0]
+                : undefined
+            }
+          />
+        }
+        {...notificationStack?.snackbarContentProps}
+      />
+    </Snackbar>
```
dimitur2204, line 57: We already have one `<Snackbar>` for the alerts from the `AlertStore`. There will probably be some problems when showing an alert from both the `<Snackbar>`s

dimitur2204 (commented):
I have some objections about `mobx`, let's see what others have to say. Amazing addition to the platform otherwise though

src/components/layout/NotificationSnackBar/DonationNotificationLayout.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <Box sx={{ display: 'flex', alignItems: 'center', maxHeight: '50px' }}>
+      <Avatar sx={{ marginRight: '1rem' }} src={person?.picture} alt="Avatar" />
+      <Box>
+        <Typography component="h1" sx={{ fontWeight: 'bold' }}>
+          {person ? `${person?.firstName} ${person?.lastName}` : 'unknown'}
+        </Typography>
+        <Typography color={grey[600]}>{`${money(amount as number, currency)}  |  ${dateToTime(
+          createdAt,
+        )}`}</Typography>
```
dimitur2204, line 34: I think that just showing "Someone donated sth, just now" is good enough

src/components/layout/NotificationSnackBar/DonationNotificationLayout.tsx
```diff
@@ -0,0 +1,38 @@
...
+}) {
+  const { person, createdAt, currency, amount } = data || {}
+
+  return (
+    <Box sx={{ display: 'flex', alignItems: 'center', maxHeight: '50px' }}>
+      <Avatar sx={{ marginRight: '1rem' }} src={person?.picture} alt="Avatar" />
+      <Box>
+        <Typography component="h1" sx={{ fontWeight: 'bold' }}>
+          {person ? `${person?.firstName} ${person?.lastName}` : 'unknown'}
+        </Typography>
```
dimitur2204, line 31: Should have some better default than "unkown" here. Probablye "Anonymous" would be most fitting

src/pages/_app.tsx
```diff
@@ -40,6 +43,20 @@ function CustomApp({
...
+  const { publicRuntimeConfig } = getConfig()
+  const { socketClientStore, notificationStore } = stores
+
+  useEffect(() => {
+    if (session?.accessToken) {
+      socketClientStore?.initiate(publicRuntimeConfig.API_URL, {
+        transports: ['websocket'],
+        withCredentials: true,
+        query: {
+          token: session?.accessToken || null,
```
dimitur2204, line 55: Don't we want people who aren't logged in to also see the notifications?

AndreyGoranov: Not sure about that requirement and also the security problems behind it

src/components/layout/NotificationSnackBar/DonationNotificationLayout.tsx
AndreyGoranov: I don't knpw

src/stores/NotificationStore.ts
AndreyGoranov: I don't see a problem with mobx using classes and react using hooks. If zustand will improve the performance of the app ok =]

src/components/layout/NotificationSnackBar/NotificationSnackBar.tsx
AndreyGoranov: Ye this could be a potential problem depending on where those alerts are being shown

src/stores/NotificationStore.ts
dimitur2204: > I don't see a problem with mobx using classes and react using hooks. If zustand will improve the performance of the app ok =]

Technically speaking there isn't a problem. I just don't like having both paradigms when the other is an option and the concept of the passing around the `this` and general observable magic that comes with it.

In terms performance and size of `mobx` + `mobx-react` 16kb + 5.5kb and `zustand` is 1.1kb, but we are talking in kb here so it's not a huge deal

AndreyGoranov: These technologies were picked from someone before us to serve a purpose and it wasn't before React introduced Hooks. I personally don't mind mixing up paradigms. About mobx it has some powerful features and it's easy to use.

src/components/faq/helpers/utils.ts
AndreyGoranov: Unfortunately, this function doesn't meet my needs

src/components/layout/NotificationSnackBar/DonationNotificationLayout.tsx
AndreyGoranov: Thinking about it Anonymous sounds better I will change it :)

src/pages/_app.tsx
igoychev: yeah, I also think the notifications should be visible to everyone

src/components/layout/NotificationSnackBar/NotificationSnackBar.tsx
igoychev: imho since the two snackbars are shown on different locations there won't be a problem. Also they are just for info and there is no user action expected on them, so even if they overlap or hide each other the user workflow will not be affected.

src/stores/NotificationStore.ts
igoychev: I agree we should depend on least possible libs and in this case it doesn't seem we need the mobx even that it already existed in the project.
See this article - the guy is presenting a working solution without additional need of mobx or any additional state management: https://blog.logrocket.com/implementing-websocket-communication-next-js/

AndreyGoranov: Sure I will remove mobx from the socket implementation

src/pages/_app.tsx
imilchev: what would be a security concern for showing the notification to everyone?

src/stores/SocketClientStore.ts
```diff
@@ -0,0 +1,39 @@
...
+        console.log('Websocket connection established')
+      })
+
+      this.socketClient.on('successfulDonation', (notificationData: NotificationLayoutData) => {
+        this.handleDonationNotification(notificationData)
+      })
+    }
+  }
+
+  handleDonationNotification(notificationData: NotificationLayoutData) {
```
imilchev, line 36: I am wondering whether it makes sense that this is in the frontend. Shouldn't the backend be handling the logic of keeping track of donations and sending notifications?

igoychev: hey yes, this is the client side listener, and the server side is on the backend in another PR - check the sendNotification function here: https://github.com/podkrepi-bg/api/pull/431/files

src/components/layout/NotificationSnackBar/DonationNotificationLayout.tsx
```diff
@@ -0,0 +1,38 @@
...
+export default function DonationNotificationLayout({
+  data,
+}: {
+  data: NotificationLayoutData | undefined
+}) {
+  const { person, createdAt, currency, amount } = data || {}
+
+  return (
+    <Box sx={{ display: 'flex', alignItems: 'center', maxHeight: '50px' }}>
+      <Avatar sx={{ marginRight: '1rem' }} src={person?.picture} alt="Avatar" />
```
dimitur2204, line 27: `person?.picture` here will not be avaliable a lot of the time. You need some kind of fallback. Either the first letters of their names or a generic avatar picture

src/components/layout/NotificationSnackBar/NotificationSnackBar.tsx
```diff
@@ -0,0 +1,61 @@
...
+      notificationStack?.remove()
+      if (notificationStack?.notifications.length) setOpen(true)
+      clearTimeout(interval)
+    }, 2000)
+  }
+
+  return (
+    <Snackbar
+      {...notificationStack?.snackbarProps}
+      open={open && !!notificationStack?.getNotifications.length}
```
dimitur2204, line 43: Maybe use `Boolean()` instead of `!!`

src/components/layout/NotificationSnackBar/DonationNotificationLayout.tsx
AndreyGoranov: The screenshot I provided shows it when we have a missing person?.picture

dimitur2204: Yeah didn't notice that. 👍

src/components/faq/helpers/utils.ts
dimitur2204: Maybe it would make more sense to have this function to there then.

igoychev (approved):

src/components/layout/NotificationSnackBar/NotificationSnackBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+      return
+    }
+
+    setOpen(false)
+    delayOpen()
+  }
+
+  useEffect(() => {
+    console.log(notifications)
+  }, [notifications])
```
dimitur2204, line 47: Left a log here

PR #1326, opened 2023-02-06
Update FE of Campaign Details page according to the new design

ani-kalpachka:


igoychev (approved):
Nice work - lots of improvements and cleanup!

PR #1331, opened 2023-02-08
en/one-time-donation.js: Swap the places of {{totalAmount}} and {{amount}}

### LINKED ISSUE #1244: [Bug] Donation and total amount texts arrangement is different for BG and EN languages in Donation page
### Describe the Bug

Donation and total amount texts arrangement is different when switch from BG to EN in Donation page. 
Maybe the problem is in file FirstStep.tsx, component "Trans".

### To Reproduce

1. Open any campaign and navigate to Donation page. 
2. Click "Donate" button
3. Choose "Other amount" for donation and fill 100 BGN
4. Check the text below. 

**Actual result:**
In BG: 
- За вашия превод от 101,72 лв., таксата на Stripe ще е 1,72 лв., а кампанията ще получи 100,00 лв.

In EN: 
- For your donation of BGN 100.00, the fee from Stripe will be BGN 1.72, and the total charged amount will be BGN 101.72

### Expected Behavior

The text arrangement should remain the same. 

In BG: 
- За вашето дарение от 100,00 лв., таксата на Stripe ще е 1,72 лв., а общият превод ще възлиза на 101,72 лв.

In EN: 
- For your donation of BGN 100.00, the fee from Stripe will be BGN 1.72, and the total charged amount will be BGN 101.72

Chrome v.105.0.5195.102

sashko9807:
Closes #1244 

## Screenshots:

Before|After
---|---
image|image

### Steps to test
1. Switch the language

### Affected urls
{{local}}/campaigns/donation/{{slug}}

### DIFF
--- a/public/locales/en/one-time-donation.json
+++ b/public/locales/en/one-time-donation.json
@@ -52,7 +52,7 @@
     "card": "Card",
     "card-include-fees": "I want to cover transaction fees for card issued in:",
     "card-fees": "For donations by card we use the services of Stripe and for every transfer they charge a fee depending on region of your card issuer. To orient you about the net donation, after you choose the desired amount, we will show you the calculated fee below. More details for Stripe fees can be found here: ",
-    "card-calculated-fees": "For your donation of <strong>{{amount}}</strong>, the fee from Stripe will be <strong>{{fees}}</strong>, and the total charged amount will be <strong>{{totalAmount}}</strong>",
+    "card-calculated-fees": "For your transfer of <strong>{{totalAmount}}</strong>, the fee from Stripe will be <strong>{{fees}}</strong>, and the campaign will receive <strong>{{amount}}</strong>",
     "card-region": {
       "title": "region",
       "EU": "Europe",


src/components/one-time-donation/steps/FirstStep.tsx
```diff
@@ -78,11 +78,11 @@ export default function FirstStep() {
         stripeIncludeFeeCalculator(chosenAmount, formik.values.cardRegion),
       )
     } else {
+      formik.setFieldValue('amountWithoutFees', chosenAmount)
       formik.setFieldValue(
-        'amountWithoutFees',
+        'amountWithFees',
         chosenAmount - stripeFeeCalculator(chosenAmount, formik.values.cardRegion),
```
igoychev, line 84: the swap of fields makes this formula incorrect. AmountWithFees should include the fees, but they are actually substracted here

igoychev (changes requested):
Good catch, however as per the screenshot, it looks like the actual problem is in the English translation. So instead of changing the BG text, the fix should be in the English text where totalAmount and amount fields need to become the same like in the BG text.

src/components/one-time-donation/steps/FirstStep.tsx
sashko9807: As per the current implementation `amountWithFees` field should contain the total amount of money to be send  after stripe's fee.
The fee is calculated by substracting `amountWithoutFee `and `amountWithFee` fields within the `<Trans>` component [link](https://github.com/podkrepi-bg/frontend/blob/master/src/components/one-time-donation/steps/FirstStep.tsx#L268). 

The swap of fields is also consistent with the way values  are assigned to those field within the if block [link](https://github.com/podkrepi-bg/frontend/blob/master/src/components/one-time-donation/steps/FirstStep.tsx#L75)

If i am to swap the amount fields in the english translation we end up with the following string:
"For your donation of <strong>{{totalAmount}}</strong>, the fee from Stripe will be <strong>{{fees}}</strong>, and the total charged amount will be <strong>{{amount}}</strong>"

In that case <strong>{{totalAmount}}</strong> value comes from `amountWithFees` field. And <strong>{{amount}}</strong> value comes from `amountWithoutFee` field, which is inconsistent with what the string refers to.

igoychev: I still think we need to carefully read the text, because actually both EN and BG translations are correct. The formulas behind them are correct too. So to avoid the confusion, we can adjust the English text to say literally the same what the Bulgarian text says?
_"For your transfer of BGN 150 the fee from Stripe will be BGN 2,30 and the campaign will receive BGN 147,70"._

sashko9807: Confusion could be caused by the the used naming convention not correctly describing the values those variables hold. That is not something this PR aims at, and should perhaps be handled in different PR.

Lets take the  BG string as it is now for example:
 "За вашия превод от <strong>{{totalAmount}}</strong>, таксата на Stripe ще е <strong>{{fees}}</strong>, а кампанията ще получи <strong>{{amount}}</strong>"

Now lets replace the naming of those variables to something which better describes the value they hold.
<strong>{{totalAmount}}</strong> `->` <strong>{{amountOfDonationWithFee}}</strong>
<strong>{{amount}}</strong> `->` <strong>{{amountOfDonationWithoutFee}}</strong>

If we apply these changes to the current BG string we end up with the following string:
 "За вашия превод от <strong>{{amountOfDonationWithFee}}</strong>, таксата на Stripe ще е <strong>{{fees}}</strong>, а кампанията ще получи <strong>{{amountOfDonationWithoutFee}}</strong>". 

Again this is the same string, and the only thing changed is the naming convention with one better describing the data those variables hold. If I am  read the string like that I would assume it is kind of incorrect.

If we apply the requested PR change to the BG string we end up with the following:
 "За вашия превод от <strong>{{amountOfDonationWithoutFee}}</strong>, таксата на Stripe ще е <strong>{{fees}}</strong>, а кампанията ще получи <strong>{{amountOfDonationWithtFee}}</strong>". 

Now the string looks correct.

Another issue with the change you propose, is that it leaves room for inconsistence behavior which could potentially break the string for both translated languages. Here is how those values are currently assigned to the fields:

[Firststep.tsx#74](https://github.com/podkrepi-bg/frontend/blob/master/src/components/one-time-donation/steps/FirstStep.tsx#L74)
```javascript
    if (formik.values.cardIncludeFees) {
      formik.setFieldValue('amountWithoutFees', chosenAmount) //amountWithoutFees is assigned differently in else block
      formik.setFieldValue(
        'amountWithFees',
        stripeIncludeFeeCalculator(chosenAmount, formik.values.cardRegion), //amountWithtFees is assigned differently in else block
      )
    } else {
      formik.setFieldValue(
        'amountWithoutFees',
        chosenAmount - stripeFeeCalculator(chosenAmount, formik.values.cardRegion),
      ) // The field is named amountWithoutFees yet we calculate fees
      formik.setFieldValue('amountWithFees', chosenAmount) // The field is named amountWithFees yet we dont calculate any fee
    }
```
Whenever the `if block` above evaluates to` true` the fields will be assigned differently than they would, if they were assigned by the `else block` execution. If  that is to happen both translations would be wrong).

Now lets see how the <strong>{{fees}</strong> variable inside the BG/EN string get its value:
[FirstStep.tsx#265](https://github.com/podkrepi-bg/frontend/blob/master/src/components/one-time-donation/steps/FirstStep.tsx#L265)
```javascript
fees: moneyPublicDecimals2(amountWithFees.value - amountWithoutFees.value),
```
Logically speaking amountWithFees should always be lesser than amountWithoutFees, so the substracting should happen the other way around.

Incase I am to "fix" just the EN string, whenever the `if block` above evaluates to true, the  <strong>{{fees}</strong> variable will hold negative value. That is because in the way values are currently assigned inside the `if block `, amountWithoutFees will always be hold bigger value than amountWithFees.

If there is still any confusion left regarding this PR and you wish to discuss it I will be happy to continue either here or somewhere else.  if not I will commit the change just to the EN string as you proposed, but due to the reasons stated above I am not convinced this is good solution.

igoychev: > If we apply the requested PR change to the BG string we end up with the following: "За вашия превод от **{{amountOfDonationWithoutFee}}**, таксата на Stripe ще е **{{fees}}**, а кампанията ще получи **{{amountOfDonationWithtFee}}**".
> 
> Now the string looks correct.

Well, this is definitely not correct - the campaign always receives the amount without the fee, because the Fee is going to Stripe.

> Logically speaking amountWithFees should always be less than amountWithoutFees, so the substracting should happen the other way around.
I am sorry but I am totally confused by the mathematics you're trying to use. There is no way that "amount" + "fee" is less than the amount itself. The logic should be the same when you're paying in a shop: amountWithVAT = amountWithoutVAT+VAT.

sashko9807: > I am sorry but I am totally confused by the mathematics you're trying to use. There is no way that "amount" + "fee" is less than the amount itself. The logic should be the same when you're paying in a shop: amountWithVAT = amountWithoutVAT+VAT.

What the  code currently does is : `amountWithVAT = amountWithoutVAT - VAT`,  which should calculate the final price with taxes included.

Due to this substracting I thought that:
amountWithoutFee - represents the amount of money user wishes to donate
amountWithFee - represents  the amount of donation a campaign will receive after the fees.

[FirstStep.tsx#83](https://github.com/podkrepi-bg/frontend/blob/master/src/components/one-time-donation/steps/FirstStep.tsx#L83)
```javascript
chosenAmount - stripeFeeCalculator(chosenAmount, formik.values.cardRegion),
```

Either way  we can keep this part as is, but the part inside the `if block`  mentioned above, needs to be addressed along with the EN translation, otherwise both translations would end up incorrect whenever the `if block` evaluates to `true`

igoychev: > What the code currently does is : `amountWithVAT = amountWithoutVAT - VAT`, which should calculate the final amount after the taxes.

Starting to wonder, are we reading the same code, because it definitely does not do that :)

If you refer to the code here: [FirstStep.tsx#83](https://github.com/podkrepi-bg/frontend/blob/master/src/components/one-time-donation/steps/FirstStep.tsx#L83)
```
formik.setFieldValue(
        'amountWithoutFees',
        chosenAmount - stripeFeeCalculator(chosenAmount, formik.values.cardRegion),
```
This is in the situation the user chose a certain amount and does not want to cover the fees, so the fees need to be substracted from the chosenAmount and the campaign will receive: amountWithoutFees = chosenAmount - Fees

sashko9807: Now everything comes clear, as to how the idea is implemented. Pushed the new commit with just EN string change.

Sorry for the hassle.

public/locales/en/one-time-donation.json
```diff
@@ -52,7 +52,7 @@
     "card": "Card",
     "card-include-fees": "I want to cover transaction fees for card issued in:",
     "card-fees": "For donations by card we use the services of Stripe and for every transfer they charge a fee depending on region of your card issuer. To orient you about the net donation, after you choose the desired amount, we will show you the calculated fee below. More details for Stripe fees can be found here: ",
-    "card-calculated-fees": "For your donation of <strong>{{amount}}</strong>, the fee from Stripe will be <strong>{{fees}}</strong>, and the total charged amount will be <strong>{{totalAmount}}</strong>",
+    "card-calculated-fees": "For your donation of <strong>{{totalAmount}}</strong>, the fee from Stripe will be <strong>{{fees}}</strong>, and the total charged amount will be <strong>{{amount}}</strong>",
```
igoychev, line 55: All good with the swap of fields, however we also need to make the text to be literally the same like the BG translation. So it should be like this - see the words in itallic:
"For your _transfer_ of {{totalAmount}} the fee from Stripe will be {{fees}} and _the campaign will receive_ {{amount}}".

sashko9807: Done.

PR #1334, opened 2023-02-09
New Donation Flow - Translations

### LINKED ISSUE #1295: New Donation Flow - Translations for the page and components

Translations (i18n)

We need translations for the new donation flow page that are not yet done. Here are the mockups and respective content: 
https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=16163%3A49451&t=kkiHE2WvZDVhQmfL-0

We currently have most of the translations from the previos form which are on the `one-time-donation` translation object. There should be a new one created called `donation-flow` and nested inside the translations for the page in a way that makes sense.

_No response_

dimitur2204:
Closes #1295 

## Motivation and context

Add all the translations to the form and changed the anonymous checkbox to be autochecked when you are not logged in.


kachar (approved):
:clap: :books: Kudos for the update @dimitur2204

We might not need `one-time-donation` translations anymore since we don't use them.

PR #1335, opened 2023-02-10
additional field for slug updating

### LINKED ISSUE #1296: Add editable campaign shortname to be used for permalink

Campaigns

Currently if a campaign link is already shared, when editing the title the system will automatically generate a new permalink/slug and the old link will become invalid. In the same time It is normal to need to change the campaign title for more engaging words during progress of campaign.

So we need the permalink to be editable separately from the campaign title. We also can't leave the first generated permalink because with the edits of title it might become different than the title itself.

### The desired solution:
Add a field that allows editing the permalink_slug separately from the title. The field should be named "shortname" and additional "info" text should instruct that this is going to be used for permalink that is not meant to be changed.

Additional requirements:
 - When campaign is created for first time and the permalink is empty, when entering the title the permalink to be created from title. 
 - Subsequent editions of title should not change the short-name automatically
 - Editing the shortname should open "red text" explaining the old link will not be available
 - Editing the shortname should still slugify it and make it url-friendly

### Even better solution with additional complexity:
 - we can support multiple permalinks to a campaign, in the similar way github is preserving the old names of code repositories, so that old links don't break
 - if the shortnames are really short and good - we can use them for friendly Bank Reference identifiers, instead of the currently cryptic GHJP_ZCW4_I58U codes which the system generates currently

BogoCvetkov:
Closes #1296

## Motivation and context

Currently if a campaign link is already shared, when editing the title the system will automatically generate a new permalink/slug and the old link will become invalid. We need the permalink to be editable separately from the campaign title. 

I've added an additional field for updating the campaign `slug` right below the one for the `title`. When editing the slug a notification message(colored in red) shows up below the field warning the user, that changing it will invalidate the current one.

#### GitHub
- **the structure of a github url** is `/{profile}/{repo}` - if we change the name of the repo to `updated repo`  the new url will be `/{profile}/{updated-repo}` - where `{repo}` and `{updated-repo}` now both link to the same repo as long as the original first slug name (`{repo}`) is not assigned to another repo in which case only the `{updated-repo}` url will link to the current one.
#### Podkrepi.bg
- the current structure for campaign urls using slugs is `/campaign/{campaign-slug}` - from my point of view if we are to support multiple permalinks per campaign using the current structure, there is a greater chance of archived slugs (stored for example in a separate DB table) being overridden by  slugs with the same name of other campaigns that are currently active. (slugs are unique while being current - e.g inside a Campaign record and when the slug is archived the slug name will be freed to be taken by some campaign). If you don't find this is much of an issue, then I don't think it would take a lot of effort to implement it on the BE. The more sophisticated approach would be to make the slugs live in their own domain -  which in the case of github is the user profile, here it might be the organisation_name for example.

## Screenshots:

Before|After
---|---
image
image

## Testing
Try modifying the campaign slug in the update form. If you change the value from the current one a warning message should be showed. Validation should still work (the same as in  the title field)

### Affected urls

`/admin/campaigns/edit/{id}`


src/components/campaigns/grid/EditForm.tsx
```diff
@@ -229,6 +232,27 @@ export default function EditForm({ campaign }: { campaign: AdminSingleCampaignRe
               autoComplete="title"
             />
           </Grid>
+          <Grid item xs={12}>
+            <FormTextField
+              type="text"
+              label="campaigns:campaign.slug.name"
+              name="slug"
```
igoychev, line 239: it will be great to have the ability to specify the slug manually in the CreateForm too - so could you pls add the field as editable there too. May be without the onchange warning because the campaign is still not created.

Note: If a user tries to create a slug that already exists, the unique-constraint on the backend will return error and the creation will not succeed, so we are safe here.
However, would be good to make the warning and error message more descriptive to show the reason. If not easy, let's split it in separate PR.

igoychev (changes requested):
Well done - thanks for this improvement!
I have a few minor comments below.

Otherwise about the idea for the extended solution with multiple slugs for campaign with a slug table on the backend: yes, this will be great for a separate PR. I think for simplicity, the table should be populated only with slugs of campaigns that reach active state, because only those links become public and sensitive.

src/components/campaigns/grid/EditForm.tsx
```diff
@@ -1,7 +1,7 @@
 import * as yup from 'yup'
-import { FormikHelpers } from 'formik'
+import { FormikHelpers, useFormikContext } from 'formik'
 import { useRouter } from 'next/router'
-import React, { useMemo, useState } from 'react'
+import React, { ChangeEvent, FocusEvent, KeyboardEvent, useEffect, useMemo, useState } from 'react'
```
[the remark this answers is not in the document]
BogoCvetkov: I shall remove those unused imports

src/components/campaigns/grid/EditForm.tsx
BogoCvetkov: - Agree about the table with the campaign slugs on the BE. Will address it in a separate PR.
- Will add the slug field to the create form as well
- I think the unique constraint is actually not handled anywhere on the FE - it is the same on user registration. If you try to register an existing email , a general message appears and it's hard to tell why you could not register. It's definitely a must to display it from UX perspective. I'll fix that and if I find it too complex - will split it in a separate PR.

BogoCvetkov: ## UPDATE

- Added the Slug Field to the Create Form as well - the behavior is as follow
    - Field is optional
    - It contains a placeholder value instructing the user that the default value will be the one for `campaign title`, unless he/she decides to set it to a custom one
    - If left empty, the` campaign slug` is the `campaign title`, else it's the value from the field

* Regarding displaying a message for when an unique constraint is violated, I'll do this in a separate PR - to keep things organized and focused

BogoCvetkov:
There was one more Form related to campaign creation outside of the admin area.  I seem to have missed updating it's type(after adding the slug property). I've fixed that and now everything is building as is should.

igoychev (approved):

PR #1338, opened 2023-02-12
Display avatars on campaign details page on local environment

ani-kalpachka:
## Screenshots:

Before|After
---|---
image|image


dimitur2204 (approved):

dimitur2204:
We should maybe ask design or just put some defaults that are not PNGs and indicate more clearly a file that doesn't exist. Something like https://placeholder.com/ maybe

Issue #437, opened 2022-01-16
[CreateCampaign] Step process for Campaign Creation


igoychev:
As per the design here: 
https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=7690%3A23718

We need to have a step-by-step campaign creation process. Create a stepper in the same way as the Support Us registration: https://dev.podkrepi.bg/support

Task Breakdown
- [ ] Create the Stepper
- [ ] Create Organizer Step
- [ ] Select Beneficiery Step
- [ ] Fill Campaign Details and Docs/Image Upload Step

nradkova:
[https://github.com/podkrepi-bg/frontend/tree/campaign-form-beneficiary-coordinator](url)

### So far created:
- stepper structure
- page layouts for each step
- partial validation
- data state preserved through steps

### Has to be done:
- validation for details page 
- file uploading functionality
-  extracting and modeling needed data for a create request
- success or error message on submitting the request

slavcho:
Closing as duplicate of #1388

PR #1341, opened 2023-02-13
Display friendly message when an unique field constraint is violated in a form

BogoCvetkov:
DEPENDS on this PR in the backend repo [#437](https://github.com/podkrepi-bg/api/pull/437)

####  Why is this change required
Currently when form data is being send and a field with unique constraint has been violated the form responds with some generic error message without giving a clue what the reason for the failure was. For example in the registration form if you try to create a new account with email that already exists in the database (is already registered) there is no message saying that it is taken (at least while I was testing on the master branch 2 days ago). The same thing applies to all forms that have fields with unique constraints 

####  What problem are you trying to solve?
Giving better information to the user why the submission failed

#### How did you solve the problem? 
I've created a reusable function that takes the name of the violated field name (returned as 409 response from the server) and displays message that informs the user of the particular property that has been violated.  I've applied the handler to the following forms:
 - Register account - email field
 - Create/Edit Campaign Form - slug field
 - Create Company - companyNumber field
 - Benefactor  create form - The submit function in this form does not attach the JWT "Bearer token" and naturally the route for creating a benefactor(which is protected) fails every time. Maybe this functionality is not yet implemented, so I haven't touched it.
 - Create/Edit Campaign-type - name field
 - Create/Edit City - postalCode field
 - Create/Edit Country - countryCode field
 - Create Donation Admin Form - extPaymentIntentId field

## Screenshots:

Before|After
---|---
image|image

## Testing

Go to the forms mentioned above and try to enter existing values for the respective unique fields and informative  error message should pop-up that the value exists


igoychev (approved):
This was a much needed UX improvement and thanks for extending it even to registration! Users will know what is happening now!

PR #1344, opened 2023-02-14
new avatar images and resize avatar

### LINKED ISSUE #1281: New Avatars

_No response_

We would like to update avatars to: 32x32 px

Logic: Non registered user with podkrepi.bg logo in black
Registered male - blue background 
Registered female - orange background 
Randomly assigned. 

future update - Possibility to be changed to an image/photo or different avatar

_No response_

martin23535:
Resize avatars and implement new avatars

Closes #1281 

## Motivation and context

I am changing default avatar to the requested new one with the black logo of podkrepime and assigning random profile picture to each new registered user from the ones in the ticket by storing the image in the User model.

## Screenshots:

sc1
sc2
sc3


src/components/auth/register/helpers/avatarsData.tsx
```diff
@@ -0,0 +1,15 @@
...
+  const avatars = [
+    '/avatars/avatar-frog-blue.svg',
+    '/avatars/avatar-frog-yellow.svg',
+    '/avatars/avatar-lion-blue.svg',
+    '/avatars/avatar-lion-yellow.svg',
+    '/avatars/avatar-panda-yellow.svg',
+    '/avatars/avatar-sheep-blue.svg',
+    '/avatars/avatar-sheep-yellow.svg',
+    '/avatars/avatar-squirrel-blue.svg',
+    '/avatars/avatar-squirrel-blue.svg',
```
igoychev, line 11: as a best practice, the array should contain only the names, but not the path and the file extension, so that path and file to be configurable separately and changeable in future.

igoychev (changes requested):
Great contribution to make the site more engaging!

Just let's return the previous person login icon to indicate what is the purpose of that menu button. The new black logo is adding dark vibe and is also confusing.

src/components/layout/nav/PrivateMenu.tsx
```diff
@@ -61,25 +62,32 @@ export default function PrivateMenu() {
   }
 
   const title = `${session?.user?.name}\n(${session?.user?.email})`
-  const lettersAvatar = `${session.user?.given_name?.charAt(0) || session.user?.email?.charAt(0)}${
-    session.user?.family_name?.charAt(0) || session.user?.email?.charAt(1)
-  }`.toUpperCase()
+
+  const getAvatar = () => {
+    const { data: user } = getCurrentPerson(!!router.query?.register)
```
dimitur2204, line 67: I think the `Boolean` cast is way more clearer than `!!` 
```suggestion
    const { data: user } = getCurrentPerson(Boolean(router.query?.register))
```

src/components/layout/nav/PublicMenu.tsx
```diff
@@ -43,8 +43,7 @@ export default function PublicMenu() {
   const { t } = useTranslation()
   const { data: session } = useSession()
   const [anchorEl, setAnchorEl] = useState<Element | null>(null)
-  const open = Boolean(anchorEl)
-
+  const defaultAvatar = '/podkrepi-bg-logo-avatar-black.svg'
```
dimitur2204, line 46: About that black file. Is that what design has provided. I personally don't like that black color of the logo and feel like a darker gray would be more suitable, something like `#c7c7c7`

src/components/layout/nav/PublicMenu.tsx
```diff
@@ -62,7 +61,15 @@ export default function PublicMenu() {
         endIcon={
           open ? <ArrowDropUpIcon color="primary" /> : <ArrowDropDownIcon color="primary" />
         }>
-        <PersonIcon />
+        <Avatar
+          sx={{
+            bgcolor: theme.palette.success.light,
+            height: theme.spacing(4),
+            width: theme.spacing(4),
```
dimitur2204, line 68: You should not be defining width and heights with the spacing function. Hard code them as pixels and make sure to add a media query to make it smaller for smaller screens if needed.

dimitur2204 (approved):
Fun and nice addition! Some small comments.

src/components/auth/register/helpers/avatarsData.tsx
```diff
@@ -1,15 +1,19 @@
+const path = '/avatars'
+const fileExtension = '.svg'
```
dimitur2204, line 2: Change them to screaming CAMEL_CASE, because they are constants

src/components/layout/nav/PrivateMenu.tsx
```diff
@@ -73,7 +73,7 @@ export default function PrivateMenu() {
       <IconButton onClick={handleMenu} size="medium">
         {session?.user?.picture ? (
           <Avatar
-            sx={{ width: theme.spacing(4), height: theme.spacing(4) }}
+            sx={{ width: 200, height: 200 }}
```
dimitur2204, line 76: Isn't that a bit too much for an avatar? Is that width and height what is actually rendered?

dimitur2204 (changes requested):

src/components/layout/nav/PrivateMenu.tsx
MartinHristovv: I had a problem where this file and the other were not updated so i was seeing the old dimensions. Fixed it

dimitur2204 (approved):

src/components/layout/nav/PublicMenu.tsx
MartinHristovv: I removed it in my second commit. We can try to improve it to sit better and add it back in the future

dimitur2204: Resolving it then. You can contact design about that if you think they might reiterate

src/components/one-time-donation/RegisterDialog.tsx
```diff
@@ -28,6 +28,7 @@ export default function RegisterForm() {
     confirmPassword: formik.values.confirmPassword as string,
     terms: formik.values.terms as boolean,
     gdpr: formik.values.gdpr as boolean,
+    picture: formik.values.picture as string,
```
dimitur2204, line 31:
```suggestion
    
```
This seems to be failing the build and is something you can leave out since we will soon change out the `one-time-donation` flow in #1272, but be sure to include some kind of default on submission of the form!

src/components/layout/nav/PrivateMenu.tsx
```diff
@@ -61,25 +62,32 @@ export default function PrivateMenu() {
...
 
   const title = `${session?.user?.name}\n(${session?.user?.email})`
-  const lettersAvatar = `${session.user?.given_name?.charAt(0) || session.user?.email?.charAt(0)}${
-    session.user?.family_name?.charAt(0) || session.user?.email?.charAt(1)
-  }`.toUpperCase()
+
+  const getAvatar = () => {
+    const { data: user } = getCurrentPerson(Boolean(router.query?.register))
+    if (user) return user.user.picture as string
+  }
```
dimitur2204, line 69: You shouldn't be calling getCurrentPerson here. Use the hook we have `useCurrentPerson` that internally calls that function, or is there a reason you need the `getCurrentPerson` call directly?

src/components/layout/nav/AdminMenu.tsx
```diff
@@ -81,8 +81,8 @@ export default function AdminMenu() {
           <Avatar
             sx={{
               bgcolor: theme.palette.success.light,
-              height: theme.spacing(4.5),
-              width: theme.spacing(4.5),
+              height: 200,
+              width: 200,
```
dimitur2204, line 85: Also change the height here

dimitur2204 (changes requested):

src/components/layout/nav/AdminMenu.tsx
MartinHristovv: good catch

src/components/layout/nav/PrivateMenu.tsx
MartinHristovv: I was not sure about whether i will need to ever pass truthy parameter so i added getCurrentPerson because it seemed 
more configurable and saw we use it on another page as well. Now it seems i can call the hook instead.

dimitur2204: @MartinHristovv Try to use hooks whenever possible and use direct API calls only within the hooks.

src/components/layout/nav/PrivateMenu.tsx
```diff
@@ -64,7 +64,7 @@ export default function PrivateMenu() {
   const title = `${session?.user?.name}\n(${session?.user?.email})`
 
   const getAvatar = () => {
-    const { data: user } = getCurrentPerson(Boolean(router.query?.register))
+    const { data: user } = useCurrentPerson()
```
dimitur2204, line 67: Hooks are only used inside the top level of components :D.
You would have to call the hooks inside the component and then pass the person to `getAvatar(person: Person)`.

dimitur2204 (changes requested):

src/components/layout/nav/PrivateMenu.tsx
dimitur2204: Actually since getAvatar is inside the component you wouldn't need to pass it, but you need to hook to be on the topLevel of the component and not be conditionally called. Read https://reactjs.org/docs/hooks-rules.html for more information

src/components/layout/nav/PrivateMenu.tsx
```diff
@@ -52,34 +53,41 @@ export default function PrivateMenu() {
...
   const router = useRouter()
   const [anchorEl, setAnchorEl] = useState<Element | null>(null)
+  const { data: user } = useCurrentPerson()
 
   const handleMenu = (event: React.MouseEvent) => setAnchorEl(event.currentTarget)
   const handleClose = () => setAnchorEl(null)
+  const getAvatar = () => {
+    const person = user?.user
+    if (person?.picture) return person.picture as string
+  }
```
dimitur2204, line 63: At this point you can get rid of the function. Last thing here I promise :D

dimitur2204 (changes requested):

src/components/layout/nav/PrivateMenu.tsx
dimitur2204: You can provide the default directly where you use the person.picture like that:
`person.picture || {the default value for the avatar}`

MartinHristovv: yes i was wondering if i should just do it this way but the function way seemed a little better since the default value 
for the avatar should be just undefined.

dimitur2204: > yes i was wondering if i should just do it this way but the function way seemed a little better since the default value for the avatar should be just undefined.

If the default value is undefined then you don't need to do anything else other than just `user?.user?.picture` which will fallback to undefined. 
Now when I see that you can also rename the data from `useCurrentPerson` to `session` so that you don't have a `user?.user`

igoychev:
@martin23535 I was looking for what remains to merge this PR
could you confirm all comments are addressed and there are some conflicts to be resolved before the merge

src/components/layout/nav/PrivateMenu.tsx
```diff
@@ -52,34 +53,41 @@ export default function PrivateMenu() {
...
   const router = useRouter()
   const [anchorEl, setAnchorEl] = useState<Element | null>(null)
+  const { data: currentPerson } = useCurrentPerson()
+  const userPicture = currentPerson?.user.picture
 
   const handleMenu = (event: React.MouseEvent) => setAnchorEl(event.currentTarget)
   const handleClose = () => setAnchorEl(null)
+  const getAvatar = () => {
+    if (userPicture) return userPicture as string
+  }
```
kachar, line 63: This function seems to be redundant, no need to wrap it as you can use directly `userPicture` variable.

src/components/one-time-donation/RegisterDialog.tsx
```diff
@@ -28,10 +29,12 @@ export default function RegisterForm() {
     confirmPassword: formik.values.confirmPassword as string,
     terms: formik.values.terms as boolean,
     gdpr: formik.values.gdpr as boolean,
+    picture: '',
   }
   const onClick = async () => {
     try {
       setLoading(true)
+      values.picture = assignRandomProfilePicture().trim()
```
kachar, line 37: This random assignment will most probably throw hydration error as it will yield one avatar on the backend and another when it runs on the client.

We need to make sure that it is stored in memo or state before using it.

PR #1355, opened 2023-02-19
Update Donation Messages styles

ani-kalpachka:
- Fixed font discrepancy on Campaign details page.
- Updated Donation Wishes styles according to the new design.

## Screenshots:

Before|After
---|---
image|image

### DIFF
--- a/src/components/campaigns/DonationWishes.tsx
+++ b/src/components/campaigns/DonationWishes.tsx
@@ -1,12 +1,16 @@
 import React, { useRef, useState } from 'react'
+
 import { useTranslation } from 'next-i18next'
-import { Unstable_Grid2 as Grid2, Stack, Typography } from '@mui/material'
+
+import { Unstable_Grid2 as Grid2, Stack, Typography, Grid } from '@mui/material'
 import RateReviewIcon from '@mui/icons-material/RateReview'
-import { useDonationWishesList } from 'common/hooks/donationWish'
-import { getExactDate } from 'common/util/date'
-import { bg, enUS } from 'date-fns/locale'
 import AccountCircleIcon from '@mui/icons-material/AccountCircle'
 import Pagination from '@mui/material/Pagination'
+
+import { bg, enUS } from 'date-fns/locale'
+import { useDonationWishesList } from 'common/hooks/donationWish'
+import { getExactDate } from 'common/util/date'
+import theme from 'common/theme'
 
 type Props = {
   campaignId: string
@@ -14,13 +18,11 @@ type Props = {
 }
 
 export default function DonationWishes({ campaignId, pageSize = 12 }: Props) {
-  const { t, i18n } = useTranslation()
-  const titleRef = useRef<HTMLElement>(null)
+  const { t, i18n } = useTranslation('campaigns')
   const locale = i18n.language == 'bg' ? bg : enUS
-
+  const titleRef = useRef<HTMLElement>(null)
   const [pageIndex, setPageIndex] = useState<number>(0)
   const { data, isSuccess } = useDonationWishesList(campaignId, pageIndex, pageSize)
-
   const numOfPages = isSuccess && data ? Math.ceil(data.totalCount / pageSize) : 0
 
   const handlePageChange = (_e: React.ChangeEvent<unknown>, page: number) => {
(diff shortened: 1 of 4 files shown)


src/components/campaigns/DonationWishes.tsx
```diff
@@ -51,28 +62,49 @@ export default function DonationWishes({ campaignId, pageSize = 12 }: Props) {
...
+                      sx={{ fontSize: theme.typography.pxToRem(37) }}
+                    />
+                  </Grid>
+                  <Stack direction="column">
+                    <Typography
+                      variant="caption"
+                      sx={{
+                        fontWeight: 700,
+                        fontSize: theme.typography.pxToRem(16),
+                        color: '#343434',
```
dimitur2204, line 83: You can use one of the `theme.pallette.grey` colors instead of #343434

src/components/campaigns/DonationWishes.tsx
```diff
@@ -51,28 +62,49 @@ export default function DonationWishes({ campaignId, pageSize = 12 }: Props) {
...
+                      }}>
+                      {wish.person
+                        ? wish.person.firstName + ' ' + wish.person.lastName
+                        : t('donations.anonymous')}
+                    </Typography>
+                    <Typography
+                      variant="caption"
+                      sx={{
+                        fontSize: theme.typography.pxToRem(14),
+                        color: 'rgba(0, 0, 0, 0.54)',
```
dimitur2204, line 93: You can use theme grey here too probably, unless you need it to be transparent

src/pages/_document.tsx
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import React from 'react'
+// eslint-disable-next-line @next/next/no-document-import-in-page
```
dimitur2204, line 3: Was the build breaking without that?

src/components/campaigns/DonationWishes.tsx
```diff
@@ -36,12 +38,21 @@ export default function DonationWishes({ campaignId, pageSize = 12 }: Props) {
...
+            color="action"
+            sx={{
+              width: theme.spacing(1.75),
+              height: theme.spacing(1.75),
+              marginRight: theme.spacing(1),
+            }}
+          />
+          <Typography
+            ref={titleRef}
+            sx={{ color: '#343434', fontSize: theme.typography.pxToRem(16), fontWeight: 500 }}>
```
dimitur2204, line 54: Theme grey color here too :)

dimitur2204 (changes requested):

src/components/campaigns/DonationWishes.tsx
ani-kalpachka: They two greys we have in the theme are different and much lighter. We should probably introduce new colors to the theme. There are some common inlined color values.

src/pages/_document.tsx
ani-kalpachka: I was not able to commit because of this error. The 'yarn run lint' command was failing.

dimitur2204: Hmm... Seems to be something with your environment maybe. I just managed to commit [5e766bb](https://github.com/podkrepi-bg/frontend/pull/1355/commits/5e766bb632234eaf6525f74223b4aa0bd505c017)

src/components/campaigns/DonationWishes.tsx
dimitur2204: Might be a good idea to introduce a darker grey if the ones there are so light.
I thought the grey has values from 50-900, that are different :)

src/pages/_document.tsx
ani-kalpachka: Yes, this is probably an issue related to my environment - other colleagues tested and they were able to commit too. I will handle it locally. Thanks for the commit! Lets see if it going to pass the build tests.

src/components/campaigns/DonationWishes.tsx
```diff
@@ -1,26 +1,28 @@
...
   pageSize?: number
 }
 
 export default function DonationWishes({ campaignId, pageSize = 12 }: Props) {
-  const { t, i18n } = useTranslation()
-  const titleRef = useRef<HTMLElement>(null)
+  const { t, i18n } = useTranslation('campaigns')
   const locale = i18n.language == 'bg' ? bg : enUS
-
+  const titleRef = useRef<HTMLElement>(null)
```
dimitur2204, line 23: Do you need that `titleRef` for something?

dimitur2204 (approved):

PR #1360, opened 2023-02-20
 Change formula to calculate Stripe's fee

### LINKED ISSUE #1358: [BUG] Difference in Stripe's fee when user wants/doesn't want to cover the fee

_No response_

### Describe the Bug

There is a slight difference in Stripe's fee between the following usecases:
-User wants to cover the fee
-User doesn't want to cover the fee.

### To Reproduce

1. Select a campaign to donate to.
2. Choose the amount you want to donate
3. Click on "I want to cover transaction fees for card issued in" checkbox
4. Observe the difference in Stripe's fee

### Expected Behavior

Stripe's fee to remain the same, regardless of whether user wants to cover the fee or not.

_No response_

sashko9807:
Closes #1358 

## Motivation and context
This PR closes the above mentioned issue, which results to a better user experience.

### Steps to test
1. Switch between wanting to cover and not wanting to cover the fee. Stripe's fee should be the same.

### Affected urls
campaigns/donation/{slug}

## Note
In case of a merge the appropriate change should also be done to the [backend](https://github.com/podkrepi-bg/api/blob/f11d5223a991af47e68b953c16161ca6ccf1c3ad/apps/api/src/donations/helpers/stripe-fee-calculator.ts#L50)

### DIFF
--- a/src/test/helpers/stripe-fee-calculator.test.ts
+++ b/src/test/helpers/stripe-fee-calculator.test.ts
@@ -1,0 +1,51 @@
+import { CardRegion } from 'gql/donations.enums'
+import {
+  stripeFeeCalculator,
+  stripeIncludeFeeCalculator,
+} from 'components/one-time-donation/helpers/stripe-fee-calculator'
+
+describe("Calculating total charged amount with Stripe's fee included", () => {
+  //Total amount of donation in stotinki, BGN 1 = 100 stotinki
+  const donationAmount = 2000
+
+  test('Calculating total charge for EU issued cards', () => {
+    const totalCharge = stripeIncludeFeeCalculator(donationAmount, CardRegion.EU) / 100
+    const formattedResult = totalCharge.toFixed(2)
+    expect(formattedResult).toMatch('20.75')
+  })
+
+  test('Calculating total charge for UK issued cards', () => {
+    const totalCharge = stripeIncludeFeeCalculator(donationAmount, CardRegion.UK) / 100
+    const formattedResult = totalCharge.toFixed(2)
+    expect(formattedResult).toMatch('21.03')
+  })
+
+  test('Calculating total charge for Other issued cards', () => {
+    const totalCharge = stripeIncludeFeeCalculator(donationAmount, CardRegion.Other) / 100
+    const formattedResult = totalCharge.toFixed(2)
+    expect(formattedResult).toMatch('21.11')
+  })
+})
+
+describe("Calculating Stripe's fee for different regions", () => {
+  //Total amount of donation in stotinki, BGN 1 = 100 stotinki
+  const donationAmount = 2000
+
+  test("Calculating Stripe's fee for EU issued cards", () => {
+    const stripeFee = stripeFeeCalculator(donationAmount, CardRegion.EU) / 100
+    const formattedResult = stripeFee.toFixed(2)
+    expect(formattedResult).toMatch('0.74')
+  })
+
+  test("Calculating Stripe's fee for UK issued cards", () => {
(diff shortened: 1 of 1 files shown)


dimitur2204 (approved):
This is how we should do it probably. Just make sure the formula in `stripeIncludeFeeCalculator{region}` is the right one

igoychev (changes requested):
The change looks logic, however have you ensured that it returns the same amount?

dimitur2204:
Could you please add jest tests for these functions consiedring they are quite crucial @sashko9807?

sashko9807:
@igoychev 
Manually checked the amount, against cards issued in all supported regions. Stripe's fee is the exact same,  with the one shown in the deployed branch, whenever "I want to cover the fee" checkbox is checked.

@dimitur2204 
Please check if those simple unit test will suffice. Also if they are in the correct directory.

igoychev:
> Manually checked the amount, against cards issued in all supported regions. Stripe's fee is the exact same, with the one shown in the deployed branch, whenever "I want to cover the fee" checkbox is checked.

The assumption that the fees should be equal is not correct.

When you base your calculations, have in mind the following:
1/ The fees are always calculated on the total charged amount.
2/ When the user selects to pay the fee AND to ensure X amount of money is donated to campaign, this makes the charged amount bigger, hence the fee is also different.

This is the same logic if you want to calculate VAT(ДДС) taxes - try this for example: https://www.scoosh.com/vat-calculator/ and see how the VAT is different when included or not included.

And finally - these calculations were tested with Stripe and Stripe calculates the same fees on the transaction.

sashko9807:
> When you base your calculations, have in mind the following: 
>1. The fees are always calculated on the total charged amount. 
>2. When the user selects to pay the fee AND to ensure X amount of money is donated to campaign, this makes the charged amount bigger, hence the fee is also different.

> This is the same logic if you want to calculate VAT(ДДС) taxes - try this for example: https://www.scoosh.com/vat-calculator/ and see how the VAT is different when included or not included.
> 
> And finally - these calculations were tested with Stripe and Stripe calculates the same fees on the transaction.

My mistake. Completely missed out that since the total charged amount is bigger, so should be the fee. This revelation makes this change completely incorrect. 

As no changes are actually required to the currently used formula(without introducing unnecessary complexity), should I close the PR, or rebase with just the unit tests in(assuming they are fine)?

dimitur2204:
> > When you base your calculations, have in mind the following:
> > 
> > 1. The fees are always calculated on the total charged amount.
> > 2. When the user selects to pay the fee AND to ensure X amount of money is donated to campaign, this makes the charged amount bigger, hence the fee is also different.
> 
> > This is the same logic if you want to calculate VAT(ДДС) taxes - try this for example: https://www.scoosh.com/vat-calculator/ and see how the VAT is different when included or not included.
> > And finally - these calculations were tested with Stripe and Stripe calculates the same fees on the transaction.
> 
> My mistake. Completely missed out that since the total charged amount is bigger, so should be the fee. This revelation makes this change completely incorrect.
> 
> As no changes are actually required to the currently used formula(without introducing unnecessary complexity), should I close the PR, or rebase with just the unit tests in(assuming they are fine)?

Adjust the tests to the logic as explained and we will roll them out.

dimitur2204 (approved):

dimitur2204 (changes requested):
You somehow seem to have commited without having ran the formatting. It should run by itself, but you can run it manually using `yarn run lint && yarn run format`

igoychev (approved):

dimitur2204 (approved):

PR #1368, opened 2023-02-22
Split the `components` folder into `client` and `admin`

dimitur2204:
## Motivation and context
For me it was incredibly confusing and I was wasting time on scrolling through the huge amounts of files in the components folder.
There were name collisions for the `donations` or things like `documents`, `beneficiary`, `benefactor` that you might think are part of the client part of the application when really they are only on the admin panel.

## Changes
General split of the components folder to `client` - `admin`.

**I have moved everything in a way where `git` knows the files are the same so any changes in future pull requests will be tracked on that specific file!**

## Testing
Tested most of the pages and I didn't see any problems
There is nothing changed and the build process goes through.


dimitur2204:
There are still things to be improved like splitting the `apiEndpoints` and the `service` folder into admin/client, but I think this one is big enough as it is.

kachar (approved):
@dimitur2204 Awesome upgrade!!

Issue #1367, opened 2023-02-22
Clickable Partners' logos


ani-kalpachka:
Go to https://podkrepi.bg/partners.
Observe the partners' logos - they are static images.
A partner's logo should be a link that leads the user to the partner's website.

dzhaniivanov:
hello,
where can i find the all partner's websites ? :)
for 88  i don't found anything and for rinkoff only this linked profile- https://www.linkedin.com/company/rinkoff/about/

PR #1372, opened 2023-02-24
logo images now redirects to partners pages

dzhaniivanov:
issue #1367 logos image now redirects to partner's pages.
todo : some websites are hard coded to google just because i don't have the correct urls

### DIFF
--- a/src/common/routes.ts
+++ b/src/common/routes.ts
@@ -34,6 +34,29 @@ export const socialUrls = {
   discord: 'https://discord.com/invite/nZAeCb9YzP',
 }
 
+export const partnerUrls = {
+  superHosting: 'https://www.superhosting.bg/',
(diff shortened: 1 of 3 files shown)


ani-kalpachka:
Well done improvement! :rocket: Here are few comments from my side:
- I think it would be better is the links are external and open in new tabs.
- Here are the missing websites that currently lead to Google:
  - 88 Studio Design: https://www.88studiodesign.com/
  - Pulse Foundation: https://pulsfoundation.org/bg/
  - СУУНЗ "Проф. д-р Иван Шишманов": http://www.suunzvarna.com/
 - Can we add the partners routes from routes.ts file and reuse them from it?

ani-kalpachka (approved):
Can we

PR #1377, opened 2023-03-05
Clean up Lint errors and warnings

ani-kalpachka:
When running 'yarn run lint' command 1 error and 17 warnings appear.
They are mainly connected to unused variables, imports and functions.
These changes clean up all of them and the command passes successfully without showing errors and warnings.


valkirilov (approved):
Maybe, it will be a good idea if we configure a before commit/push hook so we automate this validation and don't end up with a lot of unresolved errors and warnings. Or at least add this command to the GitHub pipeline and monitor the quality of the codebase?

src/components/client/campaigns/DonorsAndDonations.tsx
```diff
@@ -74,7 +74,7 @@ export default function DonorsAndDonations({
   donations: CampaignDonation[] | undefined
 }) {
   const { t, i18n } = useTranslation()
-  const [all, setAll] = useState<boolean>(false)
+  const [all] = useState<boolean>(false)
```
dimitur2204, line 77: This probably shouldn't be a state if we are not setting it

src/components/client/one-time-donation/LoginForm.tsx
```diff
@@ -11,10 +11,6 @@ import { StepsContext } from './helpers/stepperContext'
 import { AlertStore } from 'stores/AlertStore'
 import PasswordField from 'components/common/form/PasswordField'
 
-const onGoogleLogin = () => {
-  const resp = signIn('google')
-}
-
```
dimitur2204, line 17: Did you check if the google login still works?

src/components/common/GridCellExpand.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
dimitur2204, line 1: You can probably fix that inside of the file where the probelm occurs, by type checking in some way

src/pages/_document.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @next/next/no-document-import-in-page */
```
dimitur2204, line 1: I still don't know about that one. I don't get this warning

dimitur2204 (changes requested):

src/pages/_document.tsx
ani-kalpachka: This is the only error I receive. Here it is: 

image

The others are warnings.

ani-kalpachka: When running yarn next lint --no-cache I still get the same error.

src/components/common/GridCellExpand.tsx
dimitur2204: https://github.com/podkrepi-bg/frontend/pull/1377/files/28f477a92bae5439cadd5c35f26052b9285bd755#diff-6fec82610d155962e61f92f5aee8ee2c729c116f5ef4c9ef178661ba56e502d6R24

I'm guessing this is the line that is problematic. You can create a check before using it and then return the no-non-null-assertion

igoychev (changes requested):
nice cleanup!

I'm not big fan of disabling warnings so please see below a suggestion how to avoid the need of "eslint-disable-next-line" in one of the cases

src/components/admin/expenses/Form.tsx
```diff
@@ -62,6 +62,7 @@ export default function Form() {
               const currentVault = vaults?.find((curr) => curr.id == this.parent.vaultId)
               const currentAmount =
                 Number(currentVault?.amount) - Number(currentVault?.blockedAmount)
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
               return value! < Number(currentAmount)
```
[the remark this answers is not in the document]
ani-kalpachka: Agree! Thank you! It is updated.

igoychev (approved):
Nice cleanup!

dimitur2204 (approved):

PR #1383, opened 2023-03-12
Financial reports and expenses list for a campaign

slavcho:
The organizer of the campaign should be able to report the expenses of a campaign.
We are adding functionality to add expenses and attach files to them.

The expenses become visible to the general public only of they are approved by an admin.

## Screenshots:

What the average Joe sees:
image

What the organizer sees:
image

The organizer can add expenses and files:
image
image

### Steps to test

- Login as an organizer or as an admin user.
- Open a campaign.
- Scroll down to the financial report section
- Click edit
- Then click add new expense button
- Choose amount, type of expense, etc...
- Add files if you have
- Save


src/components/client/campaign-expenses/Form.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+  const { data: person } = useViewPersonByKeylockId(session?.user?.sub as string)
+
+  const fileUploadMutation = useMutation<
+    AxiosResponse<ExpenseFile[]>,
+    AxiosError<ApiErrors>,
+    UploadExpenseFile
+  >({
+    mutationFn: useUploadExpenseFiles(),
+  })
```
igoychev, line 66: the file upload might need an onError handler, because when I was trying locally it failed due to the missing bucket, but didn't show an error

igoychev (approved):
all great and very valuable addition!!!

src/components/client/campaign-expenses/Form.tsx
slavcho: I added it. But don't think it helps.
Catching those kind of error seems to be tricky.

PR #1390, opened 2023-03-16
stripe-connection-bug - add retry logic to the donation mutation

dimitur2204:
## Motivation and context

Since there are problems with connection to Stripe and sometimes 2 or 3 requests are required before a checkout session is created successfully ([detailed issue here](https://github.com/podkrepi-bg/api/issues/463)), retry logic would be useful.

## Testing
Tested with a failed request locally and the logic works as described in the `retry` method

### Affected urls
`/campaigns/donation/{slug}`


igoychev (approved):
Looks solid!

PR #1397, opened 2023-03-27
Update Active team members

ani-kalpachka:
Added a new active team member on About us page:

image


igoychev (approved):
great to have another active member!

PR #1402, opened 2023-04-03
Fix file upload bugs in the campaign expense form

slavcho:
There were a couple of problems found during the testing: 
1. We could not edit the expense, because it did not consider the previously attached files.
2. Deleting a file from the expense should be done AFTER submit, not before that.
3. The datepicker did not allow for manual typing (only mouse selection was working properly).

## Screenshots:

image


igoychev (approved):
Looks good to me!

PR #1404, opened 2023-04-10
Updates on the font sizes and row spaces in the PDF certificate

### LINKED ISSUE #1333: PDF certificate - include full donation id

Donations

### Describe the Bug

When generating the PDF certificate, for the donation id currently the code takes only the first 2 symbols, which is incorrect. 
See this line of code:
<Text style={styles.subheading}>за дарение № {donation.id.slice(0, 2)}</Text>

This needs to be corrected so that it includes the full uuid, and because it is long the font of subtitle needs to become smaller to preserve the full line.

### To Reproduce

1. Login/Register on dev.podkrepi.bg
2. Donate to any campaign using the 4242 test card
3. Go to MyProfile and in the donations table click on Download Certificate - this will display the certificate

### Expected Behavior

see screenshot
Screenshot 2023-02-09 at 13 36 46

_No response_

mm-hstefanova:
Closes #1333 

- Shows the whole donation ID.
- The font size of the donation ID is changed to fit the screen. 
- Cut off the white space of the background image (_see the BEFORE example_). 
- The background image is placed nicely on the whole PDF document ( _see the AFTER example_ ).

## Steps to test

1. Login/Register on dev.podkrepi.bg
2. If you don't have any donations to your account, donate to any campaign using the 4242 test card
3. Go to MyProfile and in the donations table click on Download Certificate - this will display the certificate

## Screenshots:

BEFORE: 
image

AFTER:
image


ani-kalpachka:
@mm-hstefanova Nice update! :)
Do you think there is a way to avoid the negative margin values?

mm-hstefanova:
> @mm-hstefanova Nice update! :) Do you think there is a way to avoid the negative margin values?

I don't like them either, but it's necessary because of the SVG logo. It takes up much white space and pushes down the other elements. That's why only the first heading has a negative margin, and the logo itself so that I can place it on the screen.

sashko9807:
> > @mm-hstefanova Nice update! :) Do you think there is a way to avoid the negative margin values?
> 
> I don't like them either, but it's necessary because of the SVG logo. It takes up much white space and pushes down the other elements. That's why only the first heading has a negative margin, and the logo itself so that I can place it on the screen.

Try setting smaller height to the img container(styles.logo). Then you can fine tune the position of the svg in that container via the viewBox attribute. [viewBox attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox).

First number - x-axis
Second number - y-axis
Third number - min-width
Fourth number - min-height


Small suggestion, temporary set a borderWidth, to styles.logo, so you can know the exact size and placement of the image container, then once you are finished remove it.

mm-hstefanova:
> > > @mm-hstefanova Nice update! :) Do you think there is a way to avoid the negative margin values?
> > 
> > 
> > I don't like them either, but it's necessary because of the SVG logo. It takes up much white space and pushes down the other elements. That's why only the first heading has a negative margin, and the logo itself so that I can place it on the screen.
> 
> Try setting smaller height to the img container(styles.logo). Then you can fine tune the position of the svg in that container via the viewBox attribute. [viewBox attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox).
> 
> First number - x-axis Second number - y-axis Third number - width Fourth number - height
> 
> Small suggestion, temporary set a borderWidth, to styles.logo, so you can know the exact size and placement of the image container, then once you are finished remove it.

Thank you for the suggestion! It worked as expected. The PR and the screenshot "AFTER" are updated!

ani-kalpachka (approved):

ani-kalpachka:
@mm-hstefanova Thanks for the contribution!

PR #1407, opened 2023-04-13
one-time-donation/Steps.tsx: Fix isAnonymous initial value

### LINKED ISSUE #1133: [Bug] You can click next if you are both not logged and not anonymous
**Describe the bug**
You can click next on the donation flow once you are on the third step with the login.

**To Reproduce**

1. Go to '/campaigns/donation/id'
2. Choose a donation value and click next
3. You can now click the next button even though you are both not logged and haven't clicked on anonymous

**Expected behavior**
You should explicitly either click on anonymous or log in

**Screenshots**
image

**Desktop (please complete the following information):**

- OS: Mac OS Catalina
- Browser Chrome 108

sashko9807:
Closes #1133 

## Motivation and context
Due to the inverting the result of isLogged(), the initial value of isAnonymous is always true, when user is not logged in.

### DIFF
--- a/src/components/client/one-time-donation/Steps.tsx
+++ b/src/components/client/one-time-donation/Steps.tsx
@@ -67,7 +67,6 @@ export default function DonationStepper({ onStepChange }: DonationStepperProps) 
     return session && session.accessToken ? true : false
   }
 
-  initialValues.isAnonymous = !isLogged()
   initialValues.isRecurring = false
 
   const userEmail = session?.user?.email


sashko9807:
@dimitur2204 as you worked on this before, would you take a look if this change is okay?

src/components/client/one-time-donation/Steps.tsx
```diff
@@ -67,7 +67,7 @@ export default function DonationStepper({ onStepChange }: DonationStepperProps)
     return session && session.accessToken ? true : false
   }
 
-  initialValues.isAnonymous = !isLogged()
+  initialValues.isAnonymous = isLogged()
```
slavcho, line 70: this change doesn't look good to me. 
Being logged it is the opposite of being anonymous.

Yes - you are probably right - this line is most likely pointless.
I will take a look and do some tests.
I believe I introduced this bug.

sashko9807: Tell me when you are done testing, if it is safe to remove the whole line, so I can amend the commit asap.

slavcho: Yep. Just remove the line. It will be fine.

slavcho (approved):

ani-kalpachka (approved):

PR #1423, opened 2023-05-02
Add UI improvements on Homepage

ani-kalpachka:
- Update Statistics section for mobile:

Before|After
---|---
|

- Update Statistics section for desktop:

Before|After
---|---
image|image

- Updated Support section text on mobile:

Before|After
---|---
|

- Removed whitespace above Active campaigns section on mobile:

Before|After
---|---
|

- Update Completed carousel dots colors and spacings in order to match design:
image

- Updates spacings.


slavcho:
Looks good to me. 
I personally vote for removing the decimal digits from the money (on both campaigns listing page and in the stats).

image

This seems too much. We either put is as **10055.22** or **10052+**, 10055.22+ is an overkill imho.
_( I can go even further with the 93+ 😀, but lets leave this to the design team. )_

ani-kalpachka:
@slavcho Thank you for the review!
I have concerns about decimal digits too. I suggest we discuss with team on which components to remove them and on which to save them and update this as a separate issue.

mm-hstefanova (commented):
Everything looks good to me as well

igoychev (approved):

ani-kalpachka:
Update: The + sign is removed:
image

PR #1432, opened 2023-05-11
[DRAFT] Apply for new campaign flow

mm-hstefanova:
Code refactoring of the CampaignFilter component:
- ListIconButtons - list of buttons with icons that can be reused for the CampaignFilter component and when we create a new campaign
- Slightly CSS updates on the ListIconButtons - fixed the jumping issue on hover and made the hover line to be as big as the button's content

image
image
image

The approach for the Create Campaign Flow is to use the React context, this way:
- we can store the data between the pages in one place
- we can save the progress of creating a campaign to a specific step
- we can change the stepper's active step from any Form component because of the context

## Screenshots:
image
image

## NOTE:
**The new components have a prefix HS, for example, HSCreateForm, this is temporary, and the idea is to copy and modify the existing components. We still use them in the Admin panel, so it's not a good idea to modify them right away before we have a working flow.**


ani-kalpachka:
@mm-hstefanova great job! 
Could you please check the lint errors?

ani-kalpachka (changes requested):
Still in progress. Converting to a draft pr.

sashko9807:
Closing this for now.

PR #1436, opened 2023-05-16
Update password validation regex on Registration form

ani-kalpachka:
Added ), ( and : symbols as valid: 

Before|After
---|---
image|image


ani-kalpachka:
@sashko9807 Thank you! It is updated with the other special symbols.

igoychev (approved):
great job @ani-kalpachka and @sashko9807 !

Issue #1352, opened 2023-02-16
[Campaigns] Simple news on campaign page


igoychev:
### Which area(s) of Podkrepi.bg are affected? (leave empty if unsure)

Campaigns

### Describe the feature you'd like to request

Currently when there are news for the campaign there is no place to show them.

### Describe the solution you'd like

The need is for a simple news section on a separate tab on the CampaignDetails Page. The news will be written in RichTextEditor in same way as the campaign description in a freestyle text one after another.

### Describe alternatives you've considered

This is intermediate feature, until a full-fledged blog-like campaign news functionality is implemented.

swolf86:
![image](https://user-images.githubusercontent.com/75036745/236704226-55b54819-2bc6-4d17-9c39-285e35caee4f.png)
We ( design) worked on this option. At the moment we are working on the pages after you click see more

PR #1446, opened 2023-06-14
[1/2]: Add campaign news functionality

### LINKED ISSUE #1445: [Feature]: Implement a full fledged blog for campaigns related news

Campaigns

Currently when there are some news regarding a campaign, the campaign's description needs to be edited, to include the campaign's updates.  This feature's main goal is to solve this problem, by implementing separate pages, where users can read the latest news about a single campaign, or all campaigns. 

Just build the feature. Sounds easy enough...

_No response_

sashko9807:
Closes #1445 #1352 

## Motivation and context
This is a desired feature.

## Screenshots:

image
image
image

## Testing
1. News article can be added/removed by admins in admin panel
2. News article can be added/removed by admins or campaign organizer by visiting campaign's URL page
3. Only admins can change article's state(published or draft)
4. As long as news article is in draft, campaign's organizer can edit or delete it as wished(as long he or she is the one who published the article in the first place)
5. Once article is published campaign organizer needs to contact the coordinator, to make any edits to the article

### Affected urls
ADDED - `campaigns/news`
ADDED - `campaigns/:slug/news`
ADDED - `campaigns/:slug/news/admin-panel`

---|---|---|---
`dompurify`|`latest`|`latest`|To sanitize the HTML content coming from article's description.


src/components/client/campaign-news/CampaignNewsList.tsx
```diff
@@ -0,0 +1,182 @@
...
+                  xs={'auto'}
+                  style={{ maxWidth: '100%' }}>
+                  <Typography className={classes.articleHeader}>{article.title}</Typography>
+                  <Grid container item>
+                    {!isExpanded[article.id] && sanitizedDescription.length > CHARACTER_LIMIT ? (
+                      <Typography
+                        component={'div'}
+                        className={classes.articleDescription}
+                        dangerouslySetInnerHTML={{
+                          __html: sanitizedDescription.slice(0, CHARACTER_LIMIT) + '...',
```
igoychev, line 136: I assume you needed to set the html from Quill via dangerouslySetInnerHTML so that you can limit the length of the description. 
A safer would be to do with the Quill editor in readonly mode but I don't know how to limit the text size, so unitl we find how, this way is ok.

src/components/client/campaigns/CampaignNewsSection.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+  [`& .${classes.dateAndAuthorContainer}`]: {
+    marginBottom: theme.spacing(2),
+  },
+
+  [`& .${classes.timelineOppositeContent}`]: {
+    paddingTop: theme.spacing(0),
+    paddingLeft: theme.spacing(0),
+    maxWidth: theme.spacing(18),
+    marginTop: theme.spacing(-0.5),
```
igoychev, line 71: AFAIK negative margins should be avoided

igoychev (approved):

src/components/client/campaigns/CampaignNewsSection.tsx
sashko9807: Without this negative margin there is a slight offset between the beginning of the content and the beginning of the timeline component(the dot). Adding any sort of margin to the `Timeline` component breaks it, thus the need to adjust the content, through negative margin. 

I could remove the negative margin if that offset doesn't bother anyone, but so far it doesn't seem to cause any conflict with other DOM elements, in either desktop or mobile viewports.

src/components/client/campaign-news/CampaignNewsList.tsx
sashko9807: The quill editor seems to use innerHTML under the hood to display the HTML contents, which is the exact same thing as  setting `dangerouslySetInnerHTML` in react. Not quite sure if it is any safer, in regards of outputting HTML, but the HTML content is being sanitized anyway.

Using `dangerouslySetInnerHTML` also should be tad a bit faster, as the HTML content is displayed directly, rather than having to wait for the React Quill editor to load.

PR #1448, opened 2023-06-14
AnonymousForm.tsx: Disable field if user is logged in

sashko9807:
Preventive measure against users' shenannigans

Complementary to [#503](https://github.com/podkrepi-bg/api/pull/503)


igoychev (approved):
well done!

PR #1450, opened 2023-06-19
added hardcoded campaign specific prices

quantum-grit:
## Motivation and context
There was request for campaign specific default donation amounts for Peter Zhotev campaign.

To accomodate this, now the default donation amounts are hardcoded in frontend.
Previously they were read dynamically from Stripe and were valid for all campaigns. 

TODO: Need to add campaign specific config for donation amounts in database


sashko9807:
Wouldn't it be better to just add a  `minDonation` field to the campaign's table and then just filter out the prices below that min amount on the front-end?

igoychev (approved):

PR #1452, opened 2023-06-21
Update Campaign cards on Homepage

ani-kalpachka:


src/components/client/index/sections/ActiveCampaignsSection/ActiveCampaignCard/ActiveCampaignCard.tsx
```diff
@@ -63,19 +47,35 @@ export default function ActiveCampaignCard({ campaign, index }: Props) {
             : targetAmount.split('.')[0]}
         </Sum>
       </SumWrapper>
+      <Grid
+        sx={{
+          background: `url(${campaignImagesUrl})`,
```
sashko9807, line 37: Not that it is going to be issue for now, but it is better to load images through Next's `<Image>` component when possible, as it offers some in-built image optimizations, which might result in slight performance improvement. This is just something to keep in mind. 

Leaving the necessary docs to read if you are interested.
[Next.js - Image Component](https://nextjs.org/docs/pages/api-reference/components/image)
[Next.js - Image Optimization](https://nextjs.org/docs/pages/building-your-application/optimizing/images)

src/components/client/index/sections/ActiveCampaignsSection/ActiveCampaignCard/ActiveCampaignCard.tsx
```diff
@@ -63,19 +47,35 @@ export default function ActiveCampaignCard({ campaign, index }: Props) {
...
+          height: '100%',
+          backgroundSize: 'cover',
```
sashko9807, line 39: Won't this approach cause some images to be either over-stretched or cut from the frame? I think a  better approach would be to set the width to `100%` and then use the [aspectRatio CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio), to adjust the image's height.

igoychev (approved):

src/components/client/index/sections/ActiveCampaignsSection/ActiveCampaignCard/ActiveCampaignCard.tsx
igoychev: Great catch @sashko9807 ! We talked with @ani-kalpachka to make the image changes in separate PR to not delay the already ready visual improvements.

ani-kalpachka: Thanks for the review! 
I will apply this change in a separate PR.

PR #1458, opened 2023-06-30
ChangePasswordForm.tsx: Fix password validation

sashko9807:
Matches the behavior from registration and profile password change form

Closes: Reported in discord


igoychev (approved):
looks

PR #1461, opened 2023-07-01
donations/grid: Allow admins to edit billingEmail

sashko9807:
Targets [podkrepi-bg/api#500](https://github.com/podkrepi-bg/api/issues/500)

## Motivation and context
Currently billingEmail is used to link all anonymous donations to a specific user profile. Unfortunately when user donates via bank transaction, and prefers the donation to be anonymous, there is no way to link that donation to user's profile. This commits allows billingEmail to be either set or edited manually, thus linking anonymous donations to user's profile

### Steps to test
1. Open admin panel -> Дарения
2. Select specific donation, then go to Billing Email field and click the edit button
3. Select the profile to the user who will be associated with that anonymous donation
4. Click the save button

### DIFF
--- a/src/components/admin/donations/grid/RenderEditBillingEmailCell.tsx
+++ b/src/components/admin/donations/grid/RenderEditBillingEmailCell.tsx
@@ -1,0 +1,111 @@
+import React from 'react'
+import { AxiosError, AxiosResponse } from 'axios'
+import { useMutation } from '@tanstack/react-query'
+import { useTranslation } from 'next-i18next'
+import { GridRenderEditCellParams } from '@mui/x-data-grid'
+import { TextField, Tooltip, Box } from '@mui/material'
+import Autocomplete from '@mui/material/Autocomplete'
+import { Save } from '@mui/icons-material'
+import { PersonResponse } from 'gql/person'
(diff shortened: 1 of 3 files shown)


src/components/admin/donations/grid/RenderEditBillingEmailCell.tsx
```diff
@@ -0,0 +1,111 @@
...
+}
+const addIconStyles = {
+  background: '#4ac3ff',
+  borderRadius: '50%',
+  cursor: 'pointer',
+  padding: 0.7,
+  boxShadow: 3,
+}
+
+export default function RenderEditPersonCell({
```
igoychev, line 28: [naming] this should be RenderEditBillingEmailCell right?

igoychev (changes requested):
all good, just a minor naming for the new cell editor

src/components/admin/donations/grid/RenderEditBillingEmailCell.tsx
sashko9807: Yup.

igoychev (approved):

PR #1478, opened 2023-07-10
FirstStep.tsx: Improve otherAmount input field

sashko9807:
-Added an upper donation limit of BGN 1.500.000(Stripe allows up to $1M for a single transaction, so we are close enough); -
-Fixed conflicting labels with input borders;
-Fixed Firefox incompatibility, which caused the input field to accept letters and symbols

Closes: Reported in discord

### Steps to test
1. Find out a browser which doesn't completely support type="number" for inputs(such as Firefox). 
2. Input field should accept only numbers or decimal points as input.


ani-kalpachka (approved):

src/components/client/one-time-donation/steps/FirstStep.tsx
```diff
@@ -343,9 +424,74 @@ export default function FirstStep() {
...
+                  e.preventDefault()
+                  return
+                }
+                if (decimalSeparator !== e.key && (e.key === '.' || e.key === ',')) {
+                  e.preventDefault()
+                  return
+                }
+
+                if (
+                  (e.key.charCodeAt(0) >= 48 && e.key.charCodeAt(0) <= 57) ||
```
igoychev, line 447: looks like repeated code from the onKeyDown above, so definitely needs to be a reusable function

igoychev (changes requested):
great improvement, just please extract the onKeyDown code as a reusable function

src/components/client/one-time-donation/steps/FirstStep.tsx
```diff
@@ -245,9 +259,77 @@ export default function FirstStep() {
...
+                    formik.errors.otherAmount &&
+                    e.key !== 'Backspace' &&
+                    e.key !== 'Delete' &&
+                    !isInteger(formik.values.otherAmount)
+                  ) {
+                    e.preventDefault()
+                    return
+                  }
+                  if (decimalSeparator !== e.key && (e.key === '.' || e.key === ',')) {
+                    e.preventDefault()
```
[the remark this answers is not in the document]
sashko9807: Considering that the onChange and onPaste behaviors are also changed, should I just extract the whole input field to its own component?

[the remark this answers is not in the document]
igoychev: great suggestion, however I was thinking we should not spend much time on something that the browser should support out-of-the-box. So let's leave the extraction for a later stage unless you want to do it now :)

src/components/common/form/NumberInputField.tsx
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        if (
+          (e.key.charCodeAt(0) >= 48 && e.key.charCodeAt(0) <= 57) ||
+          (isInteger(meta.value) && e.key === decimalSeparator) ||
+          (e.ctrlKey && e.key === 'v') ||
+          (e.ctrlKey && e.key === 'c') ||
+          e.key === 'Backspace' ||
+          e.key === 'Delete' ||
+          (e.ctrlKey && e.key === 'a') ||
```
igoychev, line 45: I've tested after the last changes and noticed that on Mac the Copy/Paste/SelectAll combinations are not working. The ctrlKey on Mac is e.metaKey - could you include in the logic too.

sashko9807: Done, but I am unable to test it.

igoychev (approved):
great for adding the mac keys

src/components/common/form/NumberInputField.tsx
igoychev: works now!

PR #1489, opened 2023-07-15
remove characters between firstName and lastName

yg1y:
Remove characters between first and last name in user details modal window

Before:

After:


slavcho (approved):
Looks good to me.

ani-kalpachka (approved):

PR #1498, opened 2023-07-18
miu-DataGrid: Remove conflicting properties when using DataGrid

### LINKED ISSUE #1497: The rows in the Coordinators table are constantly blinking

_No response_

### Describe the Bug

In the Administration, when open "Coordinators", the rows in the table are constantly blinking

### To Reproduce

Log in as administrator and open "Administration" => "Coordinators"

### Expected Behavior

The rows in the Coordinators table should not blink

_No response_

sashko9807:
autoHeight and autoPageSize shouldn't be used at the same time. By doing so this causes the data grid to constantly rerender 
ref: https://mui.com/x/react-data-grid/pagination/#automatic-page-size

Closes #1497


igoychev (approved):
that was quick - great job1

PR #1500, opened 2023-07-18
Add UI updates on Campaign cards and Inline Donation component

ani-kalpachka:
- Updated progress bar:

Before|After
---|---
image|image

- Made the whole campaign card a link (currently on prod only the image is a link to campaign details).
- Added hover effect on the card.
- Aligned Donate button.
- Add space above active campaigns.
- Updated campaign cards' spacings.

- Fixed console error caused by button tag inside a button tag:
image

- Removed unnecessary white space above campaigns title:
image

- Separated money unit from money fraction in InlineDonation component:

Before|After
---|---
image|image

TODO: Add "лв' currency for BG locale.

The changes are discussed with the design team.


package.json
```diff
@@ -38,6 +38,7 @@
     "@mui/x-date-pickers": "^6.5.0",
     "@next/font": "^13.0.5",
     "@paypal/react-paypal-js": "^7.8.1",
+    "@ramonak/react-progress-bar": "^5.0.3",
```
sashko9807, line 41: Why is a new progress bar package needed?

ani-kalpachka: I was not able to achieve this effect: 
image

 with the Progress bar provided from MUI (https://mui.com/material-ui/react-progress/).

igoychev: I had the same question but, now that we have the answer - all good. Later we can remove it if we find a way to do it without separate library.

PR #1504, opened 2023-07-19
admin/donations: Fix not being able to link user to donation if billingEmail is null

sashko9807:
This PR provides a fix for an issue, which made associating a certain donation to user impossible, when billingEmail field is null, as well as extending the autocomplete filter of the donor field, to look for both firstname and email.

Closes Reported in discord

### DIFF
--- a/src/components/admin/donations/grid/RenderEditPersonCell.tsx
+++ b/src/components/admin/donations/grid/RenderEditPersonCell.tsx
@@ -11,6 +11,7 @@ import { DonationResponse, UserDonationInput } from 'gql/donations'
 import { useEditDonation } from 'service/donation'
 import { ApiErrors } from 'service/apiErrors'
 import { AlertStore } from 'stores/AlertStore'
+import { personFilter } from 'components/common/person/PersonAutoCompleteFilter'
 
 interface RenderEditCellProps {
   params: GridRenderEditCellParams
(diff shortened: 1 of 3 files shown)


src/components/admin/donations/grid/RenderEditPersonCell.tsx
```diff
@@ -86,12 +87,18 @@ export default function RenderEditPersonCell({
...
-        filterOptions={createFilterOptions<PersonResponse>({
-          matchFrom: 'any',
-          limit: 5,
-          ignoreCase: true,
-          trim: true,
-        })}
+        filterOptions={(options, state) => {
+          const displayOptions = options.filter(
+            (option) =>
+              option.firstName
```
slavcho, line 93: can we extract this to a function like:

```
function contains(value: string, pattern: string): boolean {
  const valueCleaned: string = value.toLowerCase().trim();
  const patternCleaned: string = pattern.toLowerCase().trim();
  return valueCleaned.includes(patternCleaned);
}
```

and then do:
`    (option) => contains(option.firstName, state.inputValue) || contains(options.email, state.inputValue)`

slavcho (approved):
Looks good to me. 
Just a small refactoring idea.

src/components/admin/donations/grid/RenderEditPersonCell.tsx
sashko9807: Extracted the whole filtering function, as its own module. Looks good now at least to me.

PR #1509, opened 2023-07-21
actions buttons - moved to the left side

yg1y:
Moved the action buttons to the left side in all pages in the administration


igoychev (approved):
great to have all grids aligned

PR #1512, opened 2023-07-21
info-requests - added styles

yg1y:
Paging was not visible

Before:

After:

### DIFF
--- a/src/components/admin/InfoRequestGrid.tsx
+++ b/src/components/admin/InfoRequestGrid.tsx
@@ -44,7 +44,6 @@ export default function InfoRequestGrid() {
     <DataGrid
       style={{
         background: 'white',
-        position: 'absolute',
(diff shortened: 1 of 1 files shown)


src/components/admin/InfoRequestGrid.tsx
```diff
@@ -42,14 +42,23 @@ export default function InfoRequestGrid() {
 
   return (
     <DataGrid
+      style={{
+        background: 'white',
+        height: 'calc(100vh - 300px)',
+        border: 'none',
+        width: 'calc(100% - 48px)',
+        left: '24px',
```
ani-kalpachka, line 50: Is left: '24px' going to be applied if the position is not absolute?

yg1y: 'left' and 'width' removed

ani-kalpachka (approved):

PR #1520, opened 2023-07-26
[TEMP] Don't show successfull tag if campaign has not reached 100%

sashko9807:
## Motivation and context
In some cases campaigns are marked as completed, due to their end date being reached, while the target amount has not been raised. 
This commit checks if the target amount has been raised, before showing the successfull tag.

Note:
This is a temporary solution until, a more descriptive campaign statuses are implemented.


igoychev (approved):
great!

PR #1535, opened 2023-08-01
ViewCampaignPage: Show page not found on wrong slug

sashko9807:
In some cases, the url of the campaign, is invalid due to some typos being made. And since no data is coming from the server, the only thing the user sees, is a spinner going on forever. 
Fix this by rendering NotFoundPage, if the server has returned an error

Closes N/A

## Motivation and context
Improves UX


ani-kalpachka (approved):

igoychev (approved):
yep, better with the NotFound error than the loading forever

PR #1540, opened 2023-08-02
ViewCampaignPage.tsx: Refactoring

### LINKED ISSUE #1536: Video load slowly

_No response_

### Describe the Bug

As a user when a land on the campaign page my video is not loaded so I see an empty blank space. I would like to see some indication that the video is loading 

### To Reproduce

 image

### Expected Behavior

video to load quickly or some indication video will load 

_No response_

sashko9807:
The current ViewCampaignPage, is structured in a kind of an anti-pattern way, for react. Will give better explanation later.

Closes #1536 Hopefully

### DIFF
--- a/src/common/hooks/campaigns.ts
+++ b/src/common/hooks/campaigns.ts
@@ -12,7 +12,6 @@ import {
 } from 'gql/campaigns'
 import { DonationStatus } from 'gql/donations.enums'
 import { apiClient } from 'service/apiClient'
-import { useCurrentPerson } from 'common/util/useCurrentPerson'
 import { isAdmin } from 'common/util/roles'
 
 // NOTE: shuffling the campaigns so that each gets its fair chance to be on top row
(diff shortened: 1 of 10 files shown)


igoychev:
are you working on this one?
Just thought to test it and saw the text description is wider than before.
Screenshot 2023-08-08 at 09 33 33

ani-kalpachka:
@igoychev @sashko9807 This style issue is most probably caused by removing <StyledGrid item xs={12} md={8}> in CampaignDetails.tsx.

sashko9807:
> are you working on this one? Just thought to test it and saw the text description is wider than before. Screenshot 2023-08-08 at 09 33 33

@igoychev did you by any chance, use the mxeditor to insert the text of the campaign from the image above, because I am not able to reproduce this in local environment(image below).

Its good thing to keep in mind when , that when inserting an image or video using React Quill, the editor adds its own CSS classes to the elements, thus switching to MxEditor, is not exactly backwards compatible, with campaigns whose texts were added using React Quill.

That's the main  reason I implemented the [QuillStyleWrapper](https://github.com/podkrepi-bg/frontend/blob/master/src/components/common/QuillStyleWrapper.tsx), though it might still need some adjustments.(I haven't tried it with text inputs from mxeditor)

image

sashko9807:
Seems like removing unnecessary re-renders doesn't actually solve the issue, as I thought. 
Closing for now, as there isn't any visible performance benefits, and could be re-implemented later on.

PR #1541, opened 2023-08-02
pages/campaigns: Fix wrong prefetch url on donation page

### LINKED ISSUE #1496: Loading the donation page executes a request to http://localhost:5010/api/v1/[object Object]

Donations

### Describe the Bug

While on localhost, having started the frontend with 'yarn dev':
After clicking the donation button for any campaign, the donation page is executing request to http://localhost:5010/api/v1/[object Object]

and in the terminal you can see a long error finishing with:
```
data: {
      statusCode: 404,
      message: 'Cannot GET /api/v1/[object%20Object]',
      error: 'Not Found'
    }
```
also on the backend in the terminal window this is logged as:
```
[Nest] 43551  - 07/25/2023, 5:14:07 PM    WARN [HTTP] GET /api/v1/[object%20Object] 404 Not Found
```

The problem has no visible or functional effect on the completion of the donation.

### To Reproduce

same as above

### Expected Behavior

no error in console and no request with invalid url

_No response_

sashko9807:
We've been sending the object, rather than the url.

Closes #1496

### DIFF
--- a/src/common/hooks/campaigns.ts
+++ b/src/common/hooks/campaigns.ts
@@ -14,7 +14,7 @@ import { DonationStatus } from 'gql/donations.enums'
 import { apiClient } from 'service/apiClient'
 import { useCurrentPerson } from 'common/util/useCurrentPerson'
 import { isAdmin } from 'common/util/roles'
-import { AxiosError, AxiosResponse } from 'axios'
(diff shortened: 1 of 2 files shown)


igoychev (approved):
nice catch!

PR #1551, opened 2023-08-13
Updated admin/users grid to server-side pagination

### LINKED ISSUE #1474: Server-side paging, filtering and search for Registered users on Admin Site

User Management

Currently when opening the Admin UI for Registered Users, all of them are loaded and paged on the frontend side. 
This was ok, when the user count was low, but now we need to have the paging working on the server side to not load unnecessary data.

Use the recommended approach from MUI https://mui.com/x/react-data-grid/pagination/#server-side-pagination.
The users grid is located on https://podkrepi.bg/admin/persons - please also rename persons to users.

To make this work see how the backend implements Donations paging, search and filtering:
https://github.com/podkrepi-bg/api/blob/d032fcf1fa62f34c8851a764d19893a2370ee541/apps/api/src/donations/donations.controller.ts#L124

On the frontend https://podkrepi.bg/admin/donations the Donations filtering is implemented with custom controls, but it will be great for Users grid to use the standard MUI grid capabilities.

tongo-angelov:
Changed admin/users data grid to use server-side pagination and sorting

Closes #1474

Since I've updated usePersonList hook I had to update each place that uses it


src/service/apiEndpoints.ts
```diff
@@ -237,7 +237,15 @@ export const endpoints = {
       <Endpoint>{ url: `/campaign-types/${id}`, method: 'DELETE' },
   },
   person: {
-    list: <Endpoint>{ url: '/person', method: 'GET' },
+    list: (paginationData?: PaginationData, sort?: SortData, searchData?: string) => {
```
igoychev, line 240: the list function might need default paging values

src/pages/admin/users/index.tsx
```diff
@@ -4,7 +4,7 @@ import { endpoints } from 'service/apiEndpoints'
 
 export const getServerSideProps = securedAdminProps(
   ['common', 'auth', 'person', 'admin', 'validation'],
-  () => endpoints.person.list.url,
+  () => endpoints.person.list().url,
```
igoychev, line 7: without default values it seems this will load all users - could you check and confirm if the list function needs default paging?

igoychev (commented):
all good, just please confirm if default params are needed on the list function

igoychev (approved):
perfect!

PR #1564, opened 2023-08-18
pages/index.tsx: Dehydrate query client

sashko9807:
Needed to update the query state from prefetchQuery on the client. Improves loading time significantly.

-hook/campaigns:Remove staleTime from the useCampaignList.
 Showing stale data, might confuse donors, as they expect their donation to be shown immediately.


sashko9807:
> this is indeed a huge speedup!
> the only slight concern I have is that after refreshing the campaigns are reshuffled and their progress bars animations run again and go up or down to the new amounts. 
> 
> Could you check if we can use a default value of 0 amount when caching so that the animations run smootly always from 0 to the collected amount?

The issue was caused, due to `index` being used as key. 
Switched to `campaign.id`. Seems to be fixed now.

PR #1568, opened 2023-08-22
hooks/campaigns.ts: Disable useCampaignList on client

sashko9807:
Necessary data is prefetched on the server, and since campaign's data is no longer considered stale, it results in 2 queries to the server being made.
 Due to the call of useCampaignList, the campaigns are being reordered twice - once on the server, once on client load, which makes it unpleasant for the users.


igoychev (approved):
nice!

PR #1575, opened 2023-08-30
fix: disable source editor to avoid SWC bug

petyosi:
Due to https://github.com/swc-project/swc/issues/7640, the editor fails in production. The change here removes the plugin that causes the error.

There are two other alternatives: 
- disable the swc minification
- upgrade to the latest Next.js version

Given the non-technical audience, I think that disabling the diff/source feature is the best option. Let me know if you think otherwise, though.


ani-kalpachka (approved):

igoychev:
closing as it was fixed by updating next as per your recomendation in this PR: https://github.com/podkrepi-bg/frontend/pull/1576

Issue #1549, opened 2023-08-09
Images aren't loading after navigating through the site a bit


MartinaMitz:
### Which area(s) of Podkrepi.bg are affected?

Images

### Describe the Bug

After a while, when navigating through the website, all images start showing as broken links even when clicking on different pages (e.g. homepages, all campaigns). It only resolves when one refreshes the website.
<img width="1052" alt="Screenshot 2023-08-09 at 19 43 17" src="https://github.com/podkrepi-bg/frontend/assets/132821177/c9c9c153-0b48-43da-b57c-14f57d1116fd">

### To Reproduce

when navigating through the website, all images start showing as broken links

### Expected Behavior

Images should load even with lower bandwidth (11 Mbps)

### Which browser are you using? (if relevant)

Chrome Version 114.0.5735.248 (Official Build) (arm64)

MartinaMitz:
Unfortunately, I can't tell a particular sequence... it appears to be rather at random, but happens at least once every day.

slavcho:
@MartinaMitz  are you by any chance using a VPN or something of sort?

MartinaMitz:
No @slavcho , I am not.
Will upload a screen recording in the front-end channel too, as it's too big to add here.

slavcho:
Since we cannot reproduce this.
Please ping me on discord if you experience it again. We will do debugging in real time.

ani-kalpachka:
I tested on Chrome and Firefox browsers on desktop and mobile device and I could not reproduce such issue.

igoychev:
closing as fixed after the latest improvements, pls open new one if it happens again

PR #1580, opened 2023-09-04
Use next/image component for campaign images

sashko9807:
Use next/image component for campaign card images.

Closes #{issue number}

## Motivation and context
As of now we don't have a strict policy on image sizes for campaign's cards. This result in scenarios where a 1MB images are used, which affects both the user experience, and web vitals stats tremendously, as it takes some time before they are downloaded by the client.

The easiest way to solve this, is to use Next's Image component which provides, out of the box optimizations to reduce image size(such as automatically converting it to webp, and compressing it).


igoychev (approved):
great to use the automatic optimizations of nextjs/image

igoychev:
this change is also helping in reducing chances of dissapearing images like in this issue: https://github.com/podkrepi-bg/frontend/issues/1549

PR #1581, opened 2023-09-04
Transform otherAmount select to input field

### LINKED ISSUE #1561: Other amount field for donation custom sum

_No response_

Please allow users to enter sum in the exact filed which they entered so that they are not confused by adding separate field 

on click  this field
image
 should transform to active state of the field which they can enter a sum 
 
image

_No response_

dphilipov:
Closes #1561

## Motivation and context

* When clicking the `otherAmount` option, the input field renders in it's place instead of next to it, to avoid confusion for the users
* The input field is auto focused
* Since the issue didn't specify, I added a fade transition (200ms)

### Before

before

### After

after

### Steps to test

1. Initiate a donation to a cmpaign
2. Choose the `Other Amount` field


src/components/common/form/NumberInputField.tsx
```diff
@@ -18,6 +20,10 @@ export default function NumberInputField({
   const [, meta, { setValue, setError }] = useField(name)
   const decimalSeparator = (1.1).toLocaleString(i18n.lang).charAt(1)
 
+  useEffect(() => {
+    setValue(1)
```
dphilipov, line 24: I'm setting the value to 1, because currently there is brief moment where the validation get's triggered and the `Минималната сума която може да дарите с карта е 1лв.` error message gets shown. This was kinda hidden by the `Collapse` animation.

dphilipov:
Not related to the issue, but I believe this translation does not exist `setError(t('first-step.only-numbers'))`. Is there a translation text that I can use, or should I remove the setter?

Currently it shows this:
image

igoychev:
> Not related to the issue, but I believe this translation does not exist `setError(t('first-step.only-numbers'))`. Is there a translation text that I can use

Great find! Let's add the missing one - please search for 'transaction-limit' in the code(Ctrl+Shift+F) and it will show you the translation files with translations and feel free to add the "only-numbers" text messages.

dphilipov:
> > Not related to the issue, but I believe this translation does not exist `setError(t('first-step.only-numbers'))`. Is there a translation text that I can use
> 
> Great find! Let's add the missing one - please search for 'transaction-limit' in the code(Ctrl+Shift+F) and it will show you the translation files with translations and feel free to add the "only-numbers" text messages.

I added the missing translation. Hope the text is ok.

igoychev (approved):
great change and very professional side improvements for the translations!

PR #1593, opened 2023-09-14
Reuse main colors from theme.ts, remove inlined color values

ani-kalpachka:
Reused main colors from the theme and removed inlined hex color values - #4ac3ff, #294E85 and #FFFFFF.


igoychev (approved):
great unification of color scheme

PR #1600, opened 2023-09-18
Wrap Completed campaign card's content with link, update hover effect

ani-kalpachka:
- Wrap the whole content of completed campaign's card with a link. Currently on production only the image is a link.
- Update the hover effect to cover the whole card, not only the image.


igoychev (approved):
great for usability!

PR #1601, opened 2023-09-19
Reuse common colors from the theme

ani-kalpachka:


igoychev (approved):
great to have the colors theme unified!

PR #1616, opened 2023-10-10
Upgrade of translation libraries and related

quantum-grit:
## Motivation and context
The initial goal was to upgrade NextJS 13.5.4 to fix dependabot vulnerability warnings in dependent libraries.
However, things didn't go well and I am not proud of the process, yet the end result is good so decided to push it forward ;)

Problems and status after all changes:
 - Left NextJS at 13.4.12, because from 13.4.13 up to 13.5.4 we have a problem with translations on the CampaignDetails Page
    - resorted to setting the libraries from dependabot to fixed versions in Resolutions section
 - Attempting to fix the translations led to:
   - upgade of MUI 
   - upgrade of all i18n libraries and some side fixes there
   - fixes in regressions tests were needed too in the logic for how language is changed - tests are more stable now
   - the donations are now changed to integer amounts only, because the DecimalSeparator for custom amounts was also problematic
    - all affected regression tests with decimal amounts were updated to integer amounts
- the upgrades turned to require nodejs > 18
   - so took the opportunity to update NodeJS in Docker to node:20-alpine
   - which revealed a problem with the github actions that test the docker build and added a fix for using the proper context

All changes were good and while the PR is hard to read, I think it is all worth it - reviews can be done easier on the commits which group the changes well.


quantum-grit:
seems like the translations are broken in 13.5.4 and work well up to 13.4.12

ani-kalpachka (approved):

Issue #1261, opened 2022-12-18
Registration as an organisation/company/corporate


swolf86:
### Describe the feature you'd like to request

As a organisation I would like to donate as an organisation so that I can take advantage of tax relief and I can see that I donate as an organisation (company) 

![157932fb-3413-4e15-8b4c-a33c8c51bd8d](https://user-images.githubusercontent.com/75036745/208321405-bb747274-e60e-4bf6-855c-7bfc1318924a.png)
![c9144024-a3d8-40b5-880d-e86b0317461d](https://user-images.githubusercontent.com/75036745/208321406-6648f6c6-935f-4754-ac39-9d8138b4f6b5.png)
![a7659a0f-15b0-4c69-8459-e422646e4538](https://user-images.githubusercontent.com/75036745/208321407-a371ad83-53d6-46a4-80a8-cd85fca75834.png)
![496b7c8d-5afe-4195-b27b-03405d2d3852](https://user-images.githubusercontent.com/75036745/208321420-9233e440-f2b5-4841-98ee-6f5e343b5448.png)

### Describe the solution you'd like

Acceptance criteria: 
- organisation need to have special user profile where their name is shown as an organisation on donations and comments instead of first last name 
- not sure whether Google and Facebook can be used for organisation to register?

swolf86:
Linked with 
https://app.zenhub.com/workspaces/design-61667129a41297002437fb37/issues/gh/podkrepi-bg/design/91

Nnachevvv:
I would like to take this ticket

swolf86:
@Nnachevvv super, any questions on the requirements let me know. Technical questions - please direct to @igoychev

PR #1627, opened 2023-10-21
[1/2] Affiliate program Integration v1

sashko9807:
Closes: #1261

- Changed registration flow, to allow for corporate registrations
-Shows company name on donation list,  when donation is made through corporate profile
-Included guaranteedAmount to the reachedAmount calculation
-Match with changes made to the backend APIs

Depends on [backend#564](https://github.com/podkrepi-bg/api/pull/564)


igoychev (changes requested):
looks great so far!
please check if the change in the register endpoint could be avoided as per the backend logic

igoychev (approved):
looks good to go so far, and we can continue with improvements in separate PRs

PR #1631, opened 2023-10-26
Upgrade next version to 13.5.4

sashko9807:
-Upgrade next version to 13.5.4
-Resolve issues caused after upgrading


igoychev (approved):
great find on how to fix translations in the middleware file

PR #1635, opened 2023-10-28
[Stripe] Add refund functionality 

### LINKED ISSUE #1585: Add a refund donation button in the admin panel

Donations, Administration

Sometimes we need to return the money from a specific donation.
There could be many reasons for it.
Currently the only way to do this is manually via direct db changes.

We need a "refund" button in the admin panel on the donations list.
The refund should mark the donation as refunded and should remove the money from the campaign's vault.
This should be done in a single transaction.

It would be best if we have an explicit "right" and this access only to some admins - not all.
But that could be extracted as a different ticket later.

Nnachevvv:
Depends to https://github.com/podkrepi-bg/api/pull/556

Closes #1585 and https://github.com/podkrepi-bg/api/issues/518

## Motivation and context

Sometimes we need to return the money from a specific donation.
There could be many reasons for it.
Currently the only way to do this is manually via direct db changes.

In order to fix this I opened a new endpoint in the backend -  **/api/v1/donation/refund-stripe-paymet/<extPaymentId>**

## Screenshots:
Screenshot 2023-10-28 at 23 02 24
Screenshot 2023-10-28 at 23 02 40

Before|After
---|---

### Steps to test
1. Setup stripe test env
2. Make a donation
3. Login as admin
4. Go in donation tab
5. Refund your donation


src/components/admin/donations/modals/RefundModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+        </Grid>
+        <DialogTitle style={{ display: 'flex', justifyContent: 'center', width: '100%' }}>
+          {t('refund.title')}
+        </DialogTitle>
+        <Grid container direction="column" component="section">
+          <GenericForm onSubmit={onSubmit} initialValues={initialValues}>
+            <CardContent>
+              <Typography variant="body1">{t('refund.confirmation')}</Typography>
+              <Typography variant="body1">
+                {t('refund.number')} {data?.id}
```
igoychev, line 73: let's show here the payment intend id, so that it can be easily double-checked in the Stripe UI

igoychev (changes requested):
all great! 
Just found two typo's and also please let's display the paymentIntentId in the confirmation modal.

igoychev (approved):
cool, all good!

Issue #1434, opened 2023-05-15
User account, created via Google SSO, cannot be deleted


donangel:
### Describe the Bug

When the user tries to delete their account created through Google SSO, they get a password prompt:

<img width="658" alt="image" src="https://github.com/podkrepi-bg/frontend/assets/4919119/6b11e88b-61d6-41f9-a596-e3208b4b70ea">

Since there's no assigned password for the account, they can never delete their profile from the system.

### To Reproduce

1. Go to https://podkrepi.bg/register
2. Choose "Вход с Google"
3. Go to [My Profile](https://podkrepi.bg/profile/donations), [Лична информация](https://podkrepi.bg/profile/personal-information)
4. Click on "[изтриване на акаунт/ профил]"

### Expected Behavior

The user does not get a password prompt, just "Изтрий моя акаунт", maybe with some additional "Are you sure" checkbox to prevent deleting the account by mistake.

PetarDimitrov91:
I will take a look on this ticket.

PetarDimitrov91:
@igoychev after PR #1662 this issue can be closed.

igoychev:
The initial problem is solved in #1662 , however while fixing we discovered that not all personal information is deleted. 
So we need an extension of the fix that enables "The right to be forgotten" which removes all personal information - this will be a new ticket.

sashko9807:
Doesn't seem to be the case anymore.

Tuscann:
https://github.com/podkrepi-bg/frontend/assets/16596102/19923df1-b876-4865-8ca3-05b8bbd7a49d

Please open this bug @sashko9807.

sashko9807:
Opened a new issue instead, as this issue isn't Google SSO specific. 
Link to new issue [api#641](https://github.com/podkrepi-bg/api/issues/641)

PR #1662, opened 2023-11-10
Bugfix/google auth del acc

PetarDimitrov91:
Fixed the issue displaying the password input field for users using Google SSO by adding a token mapper in Keycloak. Manual configuration is required for both dev and prod environments. Introduced a 'selfReg' field in the ServerUser interface to indicate if the user is registered normally. In the DisableAccountModal, this new prop is used to determine whether to display the password field. Additionally, adjusted the position of the used templates.

### DIFF
--- a/src/components/client/auth/profile/DisableAccountModal.tsx
+++ b/src/components/client/auth/profile/DisableAccountModal.tsx
@@ -93,7 +93,7 @@ function DisableAccountModal({
   const [loading, setLoading] = useState(false)
 
   const { data: session } = useSession()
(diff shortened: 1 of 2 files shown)


PetarDimitrov91:
For some reason, I cannot add the 'run tests' label,

sashko9807 (approved):

igoychev (approved):

igoychev:
part 1 of fixing https://github.com/podkrepi-bg/frontend/issues/1434

PR #1667, opened 2023-11-14
Added campaign donation charts

### LINKED ISSUE #1532: [Donations] Add Statistics Dashboard for Donations

Campaigns, Donations

Currently users of Podkrepi.bg can see the donations log, but cannot see an overall impression of how the donations were collected during time, what is the distribution of amounts and who are the biggest donors. Adding this information in a Statistics Dashboard would improve user engagement, will focus the marketing efforts and will increase the transparency of the platform.

Donations Statistics Dashboard should appear as a separate tab on every campaign page and should be visible by all users of the web site. 

#### Performance considerations:
- the data should be loaded only when the tab is visible
- load only the needed data fields from the backend and if needed create /statistics API endpoints
- the backend endpoints serving the data should be cached for at least 30min with a notice on the page saying "The Donations Statistics are intentionally showing cached data delayed to max 30min of your first load."

#### List of charts needed for now:
- [ ] Cumulative Donations Timeline(vertical area) - shows the accumulation of the total amount over time. The impression the chart makes is always growing with flat periods where no donations are made 
- [ ] Donations Amounts Timeline (vertical barchart) - shows the sum of the donated amounts as they came in the selected time interval (daily, weekly, monthly) without addition with the previous interval. If possible on a second Y axis we can show the count of donations too or add it as a separate graph. The chart will show 0 amounts when no donations are made during that interval.
- [ ] Donations Amount Distribution (vertical barchart) - shows the count of donations of the same amount. On X axis are the unique amounts like 10, 20, 50, 100, 1000 etc and on the Y axis is the count of donations of the same amount. The goal of this chart is to show which are the preferred amounts, as well as the min and max amounts donated.
- [ ] Donations During Day Distribution (vertical barchart) - shows how users are donating during the day. On X axis we show only the time of the donation, on Y axis is the count of donations done. Some charts allow grouping in time intervals like 5, 15, 30min or 1h automatically - please enable it for selection

The ticket can be broken in subtickets as each chart brings huge value and it's better to deliver them in iterations.

A main cross-campaign dashboard can be added too, however that will be a separate effort as it will probably require preprocessing of the huge amount of data and preparing summary reports.

tongo-angelov:
Added campaign statistics charts.
The server returns only dates with donation, so the charts generate the empty data to fill the graphs

Closes #1532 

## Screenshots:

Screenshot from 2023-11-14 19-29-43

Screenshot from 2023-11-14 19-28-23
Screenshot from 2023-11-14 19-28-32
Screenshot from 2023-11-14 19-28-42
Screenshot from 2023-11-14 19-28-59
Screenshot from 2023-11-14 19-29-16

"chartjs-plugin-annotation": "^3.0.1"


src/components/client/campaigns/CampaignInfo/CampaignInfoStatus.tsx
```diff
@@ -55,6 +56,16 @@ export default function CampaignInfoStatus({ campaign, showExpensesLink }: Props
           size="small"
           sx={{ backgroundColor: theme.palette.primary.light }}
         />
+        {!!campaign.summary.reachedAmount && (
+          <Chip
+            component="a"
+            label={t('campaigns:statistics.button')}
```
igoychev, line 62: Good idea to show the stats only after having donations

src/components/client/campaigns/CampaignStatistics/CumulativeDonationsChart.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  statistics.forEach((donation) => {
+    while (previousDate < new Date(donation.date)) {
+      finalData.push({ date: previousDate.toISOString(), count: 0, sum })
+      previousDate = addHours(previousDate, 24)
+    }
+    sum += donation.sum
+    finalData.push({ ...donation, sum })
+    previousDate = addHours(previousDate, 24)
+  })
```
igoychev, line 100: well done with retrieving only the donations data and filling the flat no-donation periods on the frontend!

igoychev (approved):
looks great!

PR #1676, opened 2023-11-29
Implement fullscreen image slider when image is clicked

sashko9807:
Currently we don't have a gallery component, and when image is uploaded with role==gallery those images remain unusable on larger screens, due to their small size This commit makes it so a full-screen gallery like is opened whenever an image with role==gallery is clicked.

gallery-gif


ani-kalpachka (approved):

ani-kalpachka:
I think this looks really cool! :rocket:

PR #1687, opened 2023-12-14
CampaignInfo.tsx: Use category of  type to translate campaign type

sashko9807:
New campaign types are added dynamically through the admin panel, thus some translation could be missing in the file. Use name of campaignType until we figure out a way to translate them dynamically.

### DIFF
--- a/src/components/client/campaigns/CampaignInfo/CampaignInfoStatus.tsx
+++ b/src/components/client/campaigns/CampaignInfo/CampaignInfoStatus.tsx
@@ -69,9 +69,7 @@ export default function CampaignInfoStatus({ campaign, showExpensesLink }: Props
       </Box>
       <InfoStatusWrapper>
         <Grid item xs={12} md={6}>
-          <StatusLabel>
-            {t(`campaigns:campaignTypesFields.${campaign.campaignType.slug}`)}
-          </StatusLabel>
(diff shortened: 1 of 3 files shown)


src/components/client/campaigns/CampaignInfo/CampaignInfoStatus.tsx
```diff
@@ -69,9 +69,7 @@ export default function CampaignInfoStatus({ campaign, showExpensesLink }: Props
       </Box>
       <InfoStatusWrapper>
         <Grid item xs={12} md={6}>
-          <StatusLabel>
-            {t(`campaigns:campaignTypesFields.${campaign.campaignType.slug}`)}
-          </StatusLabel>
+          <StatusLabel>{campaign.campaignType.name}</StatusLabel>
```
slavcho, line 72: We will loose the translation this way.
Can't we use campaignType.category?

slavcho (changes requested):

src/components/client/campaigns/CampaignInfo/CampaignInfoStatus.tsx
sashko9807: Done.

slavcho (approved):

PR #1704, opened 2024-01-16
Invalidate a donation from the admin panel

slavcho:
This gives certain users with the proper credentials the right to mark a donation as invalid.
We need this in the rare cases when a manual fix is need due to some problem with stripe or the bank.

image


ani-kalpachka:
There is a formatting error in Grid.tsx which is blocking the build. @slavcho could you please check this?

ani-kalpachka (approved):

src/components/admin/donations/grid/Grid.tsx
```diff
@@ -166,6 +182,14 @@ export default observer(function Grid() {
                 <RestoreIcon />
               </IconButton>
             </Tooltip>
+            <Tooltip title={t('donations:invalidate.icon')}>
```
sashko9807, line 185: Invalidating donation should be outside of `params.row?.provider === PaymentProvider.stripe` condition, as we might need to invalidate bank donations as well.

sashko9807 (changes requested):

src/components/admin/donations/grid/Grid.tsx
slavcho: Fixed.

sashko9807 (approved):

PR #1709, opened 2024-01-22
Incorrect restriction in  withdrawls edit form

slavcho:
The amount of the withdrawals can be a decimal number. 
It should not be limited to integers only.

Before:

image


sashko9807:
If I remember correctly ,we save the amount of a withdrawal record in stotinki as an integer, and conversions to decimals are only done on the front end for data visualization.

slavcho:
> If I remember correctly ,we save the amount of a withdrawal record in stotinki as an integer, and conversions to decimals are only done on the front end for data visualization.

this is handled in:
```
  function handleSubmit(values: WithdrawalInput) {
    const data: WithdrawalData = {
      amount: toMoney(values.amount),
      

```

Actually we can save the form, but we cannot edit it later. It should be the same logic.
And this restriction is not in the CreateForm.

slavcho:
@ani-kalpachka Can you please review this since Diana requested it urgently?

src/components/admin/withdrawals/EditForm.tsx
```diff
@@ -47,7 +47,7 @@ const validationSchema: yup.SchemaOf<WithdrawalData> = yup
   .defined()
   .shape({
     status: yup.string().trim().min(1).max(10).required(),
-    amount: yup.number().positive().integer().required(),
+    amount: yup.number().positive().required(),
```
sashko9807, line 50: Not that it is an issue considering the amount field can't be edited in EditForm, but maybe for consistency reason we should at least have the number().positive(), rules for amount validation.

sashko9807 (approved):

PR #1711, opened 2024-01-30
Fix translations on Contact page

ani-kalpachka:
Before|After
---|---
image|image

### DIFF
--- a/src/pages/contact.tsx
+++ b/src/pages/contact.tsx
@@ -5,7 +5,13 @@ import ContactPage from 'components/client/contact/ContactPage'
 
 export const getStaticProps: GetStaticProps = async ({ locale }) => ({
   props: {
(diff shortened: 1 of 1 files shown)


sashko9807 (changes requested):
Moving email subscription related translations to common.json, would be the better and more scalable solution, as the `common` namespace is used in almost every page.

slavcho (approved):

ani-kalpachka:
@sashko9807 I agree. I am going to merge this in order to fix the problem and extract and reuse the common translations when available in another pr. Thank you for noticing!

sashko9807 (approved):

sashko9807:
> @sashko9807 I agree. Thank you for noticing! Lets merge this in order to fix the problem and extract and reuse the common translations in another pr? I will do it when available.

Just keep in mind that the issue, affects and other pages, in which campaigns namespace is not included(e.g. campaign news page), not just the reported one.

Issue #1824, opened 2024-05-28
After click on checkbox moving page one pixel on register page


Tuscann:
### Which area(s) of Podkrepi.bg are affected?

Desktop layout, Mobile layout

### Describe the Bug

https://github.com/podkrepi-bg/frontend/assets/16596102/01c8551f-bcfb-475c-954e-7780b7902b61

Why page moving with one pixel after click checkbox?

### To Reproduce

1. Open [https://podkrepi.bg/en/register](https://podkrepi.bg/en/register)
2. Click F5 from keyboard
3. Click checkbox for "I understand and I agree to the [General Data Protection Regulation (GDPR)]"
4. Why whole page move with one pixel?

### Expected Behavior

After click checkbox no moving of bottom of the page

### Which browser are you using? (if relevant)

Floorp

ani-kalpachka:
At first sight, the issue should be resolved if we set the margin-top of the text error to 0:

![image](https://github.com/podkrepi-bg/frontend/assets/14351733/70873521-4631-40d0-959d-ecead90ce715)

PR #1827, opened 2024-05-30
After click on checkbox moving page one pixel on register page #1824

iliyan90:
After click on checkbox moving page one pixel on register page #1824

Closes #{1824}

## Motivation and context
margin-top is set to 0


ani-kalpachka (approved):

ani-kalpachka:
Well done! 
We have this problem on Contacts page too. This global solution would fix the issue everywhere.

Issue #1816, opened 2024-05-27
Translate to bulgarian table in tab "Доброволци" in admin panel


Tuscann:
### Which area(s) of Podkrepi.bg are affected?

Desktop layout

### Describe the Bug

![image](https://github.com/podkrepi-bg/frontend/assets/16596102/641e48b1-872e-4d77-9856-7f1261d33a27)

Need translation to bulgarian name of columns

### To Reproduce

1.Open [https://podkrepi.bg/admin/supporters](https://podkrepi.bg/admin/supporters)
2. Login with admin user
3. Check all columns names.
4. 

### Expected Behavior

Name of columns to be in bulgarian.

### Which browser are you using? (if relevant)

Vivaldi

ani-kalpachka:
The issue is fixed here: https://github.com/podkrepi-bg/frontend/pull/1838
I will close the ticket as completed.

PR #1838, opened 2024-05-31
Translate to bulgarian table in tab Доброволци in admin panel #1816

iliyan90:
Translate to bulgarian table in tab Доброволци in admin panel #1816

Closes #{1816}
afterTranslate

### DIFF
--- a/src/components/admin/SupportersGrid.tsx
+++ b/src/components/admin/SupportersGrid.tsx
@@ -19,51 +19,51 @@ const columns: GridColDef[] = [
   { field: 'id', headerName: 'ID' },
   {
     field: 'name',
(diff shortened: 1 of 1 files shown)


ani-kalpachka (changes requested):
We need to have both Bulgarian and English translations in order to be able to switch the language. You can check how we store the BG and EN translations in public/locations folder and use them with useTranslation which comes from 'next-i18next' library.

ani-kalpachka:
I will merge this change. We are not going to need EN translations on the Admin panel.
Note: I am not sure if the comments are needed in this case.

ani-kalpachka (approved):

PR #1840, opened 2024-05-31
Add translation updates

ani-kalpachka:
- Add missing EN translations on FAQ page and Profile page.
- Add punctuation updates on Register form and news articles text.


src/components/common/form/AcceptPrivacyPolicyField.tsx
```diff
@@ -17,7 +17,7 @@ export default function AcceptPrivacyPolicyField({ name }: AcceptGDPRFieldProps)
       label={
         <Typography variant="body2">
           {t('validation:informed-agree-with')}{' '}
-          <ExternalLink href={routes.privacyPolicy}>{t('validation:gdpr')}</ExternalLink>
+          <ExternalLink href={routes.privacyPolicy}>{t('validation:gdpr')}</ExternalLink>.
```
sashko9807, line 20: Is there a reason why the full-stop is not part of the translated string?

src/components/common/form/AcceptTermsField.tsx
```diff
@@ -20,6 +20,7 @@ export default function AcceptTermsField({ name }: AcceptTermsFieldProps) {
           <ExternalLink href={routes.termsOfService}>
             {t('validation:terms-and-conditions')}
           </ExternalLink>
+          .
```
sashko9807, line 23: Same question as above.

sashko9807 (changes requested):

src/components/common/form/AcceptPrivacyPolicyField.tsx
ani-kalpachka: If I insert it as a part of the string it is included as a link in blue. This is why I thought it is better to be this way.

sashko9807 (approved):

PR #1860, opened 2024-06-21
Update Subscription modal translations

ani-kalpachka:
- Update Subscription modal translations on the pages that they were missing.
- Update punctuation on validation.json.


sashko9807 (changes requested):
GeneralSubscriptionModal is accessible on every page through the footer, so any translations to it must be in the common namespace.
 Please put the untranslated modal texts inside the common.json, within the [notifications](https://github.com/podkrepi-bg/frontend/blob/master/public/locales/bg/common.json#L104) object, and make the necessary changes to the `AcceptNewsletterField.tsx` and `GeneralSubscribeModal.tsx` components.

ani-kalpachka:
@sashko9807 I agree. Thanks for the review. I will close this PR and create a new one.

Issue #1842, opened 2024-06-05
[New Epic Feature] Create a new campaign request from website


Juicebgs:
### Which area(s) of Podkrepi.bg are affected? (leave empty if unsure)

Campaigns

### Describe the feature you'd like to request

**First things first**: In order to better understand this feature, you will need to know the main stakeholders and terms. You can find information [here ](https://docs.podkrepi.bg/general/architecture/glossary). The text is in Bulgarian, you can user Google translate if you need help with the translation 

 **Current workflow**: If somebody wants to start a campaign on podkrepi.bg website, he or she starts by filling in a Google form. The form can be found [here](https://docs.google.com/forms/d/1qHCnuR-1Fe44ZZ7hlbIkGtjoVd6j6LiAXfSw-ugO1ZM/viewform?edit_requested=true#response=ACYDBNiiizeOU37Jq0XDMYAHEvARvp7xrJBQ-L3pewVdRyfwx1CIofM5au9Sa7j4njvs36c). 

The form is than assigned to a member of campaigns "department" and the volunteer initiates all the necessary steps to ensure that the form is correctly filled out, there is enough information about the campaign, finances, gathers expert opinion (if needed). This is done by updating the information in a Google Sheet - a template can be found [here](https://docs.google.com/spreadsheets/d/1g-ifDIzIhJJpetiypHxrC-IJNgDiPABPgqKfIn3yiw0/edit?usp=sharing), adding documentes to Google Drive and exchanging emails, phone calls and/ or meeting with the organizer of the campaign or experts on the campaign subject. Tasks are tracked via Trello Board. 

After all the information has been gathered and checked (which can take multiple iterations with the Organizers), three members of the Podkrepi.bg Board should approve the start of the campaign. 

**Our task**: Our task here is to enable Organizers to create a new Campaign Request from podkrepi.bg website and be able to edit the Request following off-system communication with podkrepi.bg volunteers. We will need to create a MVP of the process (basic features only). In time with additional feature requests we will expand this feature to include additional user stories. The main task, though, is to have a working Campaign Request form in the platform and Campaigns team to be able to oversee the Campaign approval process through the platform's admin panel. 

  

### Describe the solution you'd like

**Design**: Campaign Request creation prototype can be found [here](https://www.figma.com/file/MmvFKzUv6yE5U2wrOpWtwS/Podkrepi.bg?node-id=7690%3A23718)
[...]

PR #1863, opened 2024-06-24
feat: create campaign application organizer form (Epic #1842)

teodorazhelyazkova:
Create campaign application organizer form connected to [Epic #1842](https://github.com/podkrepi-bg/frontend/issues/1842)
I have used for reference the `src/components/client/support-form/SupportForm.tsx`

### Steps to test

1. Login as normal user
2. Visit `http://localhost:3040/campaigns/application`


src/components/client/campaign-application/helpers/campaignApplicationFormActions.styled.tsx
```diff
@@ -28,8 +27,7 @@ export const ActionButton = styled(Button)(() => ({
   padding: theme.spacing(1, 5),
```
gparlakov, line 27: It looks like a lot of repetition here. Can we do something about that? Have a button and variations for the button with only the changed styles?

src/components/client/campaign-application/helpers/validation-schema.ts
```diff
@@ -0,0 +1,30 @@
...
+import {
+  CampaignApplicationFormDataSteps,
+  CampaignApplicationOrganizer,
+  Steps,
+} from './campaignApplication.types'
+
+const organizer: yup.SchemaOf<CampaignApplicationOrganizer> = yup
+  .object()
+  .shape({
+    firstName: name.required(),
```
gparlakov, line 14: Organizer name will be a single string. Checkout the https://dev.podkrepi.bg/swagger#/campaign-application/CampaignApplication.create schema

gparlakov: Also - these values (name, email and phone ) should be prefilled from the current user. Please create a task for this over at https://github.com/users/gparlakov/projects/1/views/1

src/components/client/campaign-application/helpers/campaignApplicationFormActions.styled.tsx
teodorazhelyazkova: @gparlakov No, because all of the styled components wrap a different component underneath (Button, LinkButton, SubmitButton) and the styled() utility cannot take as params more than one component.

src/components/client/campaign-application/helpers/validation-schema.ts
teodorazhelyazkova: > Organizer name will be a single string. Checkout the https://dev.podkrepi.bg/swagger#/campaign-application/CampaignApplication.create schema

Done

src/components/client/campaign-application/helpers/campaignApplicationFormActions.styled.tsx
gparlakov: @teodorazhelyazkova would it not work if we were to 
```ts
const common = {
  padding: theme.spacing(1,5),
  color: 'black'
 }
 
 export const ActionButton = styled(Button)(() => ({ 
    ...common,
    my-different: style-here
   })
   ```

Comment (teodorazhelyazkova) [reply src/components/client/campaign-application/helpers/campaignApplicationFormActions.styled.tsx]: You are right, it works. I've changed it.

**Review by gparlakov — COMMENTED** [review]
Looks good to me

**Review by sashko9807 — APPROVED** [review]
On the style repetitiveness topic, there is no rule saying that each component we want to style should be a styled component. You can have single single styled wrapper, and define the styles of all child components , through CSS selectors(elements, classes etc..).  


For example you could have something like
```tsx
export const OrganizerDataStepStyled = styled(Grid)(() => ({
 '& .header': {
     fontSize: theme.typography.pxToRem(18)
     ....
  },
 'input': {
     fontSize: theme.typography.pxToRem(18)
     borderRadius: 10
    .....
  },
 'button': {
  padding: theme.spacing(1, 5),
  border: `1px solid ${theme.palette.common.black}`,
  borderRadius: theme.borders.round,
  color: theme.palette.common.black,
  fontSize: theme.typography.pxToRem(15),
  fontWeight: 800,
  }
})
```

Then you use it like this
```tsx
//OrganizerDataStepStyled inherits all properties of the Grid component
<OrganizerDataStepStyled container>
<Header className={'header'}>Something nice</Header>
<Button variant='outline'>Go back</Button>
<Button variant='contained'>Save and continue</Button>
</OrganizerDataStepStyled>
```

You can find this pattern used in the codebase to get the idea. Quick example I found is [CampaignNewsList.tsx](https://github.com/podkrepi-bg/frontend/blob/9552937a1e39f1f006af87c417403881f1c780ca/src/components/client/campaign-news/CampaignNewsList.tsx#L33)

P.S. This comment was just to give you an idea of alternative solution on how to work with MUI's styled components. We don't have a strict policy(at least for now), when it comes to building patterns for react component, thus feel free to use the pattern which comes more natural/logical to you.

PR #1867, opened 2024-06-30
feat: campaign application admin init

gparlakov:
## Motivation and context
- issue #1842
- add a admin space for the campaign applications
- only skeleton - will be fleshed out iteratively
- add admin edit campaign application
  - it will show all the steps of the stepper (create campaign application) with addition of the admin-only editable stuff

## Screenshots:

image
image

Before|After
---|---

---|---|---|---


teodorazhelyazkova:
@gparlakov What are the credentials for the admin test account?

sashko9807:
> @gparlakov What are the credentials for the admin test account?

User:`admin@podkrepi.bg`
Password:`$ecurePa33`

sashko9807:
Perhaps it would be good to hide this section for the production environment, until everything to this is finished, to prevent confusion of non-tech people.
What do you think?

gparlakov:
That might be a problem but since I'm not well acquainted with the team I'll defer to you for this. 

I can say that there are 2 main goals of this:

- get some early feedback from the team i.e. this is how I imagine it being laid out and working is it good for you or would you expect something else (this one might be harder to achieve with the page being hidden) 
- allow parallel execution of the sub-components of the page (this one is not affected by hiding the page)

And finally - we might be trying to solve too many issues with this one PR. I suggest taking it step by step and if there is some negative feedback we can always hide the menu.

sashko9807 (approved):

PR #1882, opened 2024-07-10
chore: add admin campaign applications

gparlakov:
- add a login fixture
- add a test that verifies the current admin list of campaign applications

## Motivation and context

Authentication of a user in e2e tests context 
Verification of Admin Campaign Application pages 
## Testing
run the e2e tests 

- [x] `PODKREPI_EMAIL`: email for user auth
- [x] `PODKREPI_PASSWORD`: password for user auth

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -43,3 +43,5 @@ PAYPAL_CLIENT_ID=sb
 ###########
 GHOST_API_URL=https://blog.podkrepi.bg
 GHOST_CONTENT_KEY=86ec17c4b9660acd66b6034682
+PODKREPI_EMAIL=admin@podkrepi.bg
+PODKREPI_PASSWORD=$ecurePa33
--- a/README.md
+++ b/README.md
@@ -100,7 +100,9 @@ Watch releases of this repository to be notified about future updates:
 ## Contributors ✨
 
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
+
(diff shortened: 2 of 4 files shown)


sashko9807:
E2E job seems to fail.
Though im not sure if any auth-based e2e test would ever pass, as we don't have a job to setup keycloak, which is needed for auth(or at least I don't see one).

gparlakov:
Let me see if I can add a step for that keycloak

gparlakov:
> E2E job seems to fail.
I can't seem to find the run of those (neither unit-tests.yaml or playwright.yaml-s workflow logs show a failed run) @sashko9807 was that run locally or via the actions?

gparlakov:
@sashko9807 sorry - it looks like the latest playwright.yaml (addition of keycloak) was ignored - is this a cache thing? Or did I miss something?

image

Notice the command (on the left) was `docker-compose up -d pg-db` while the update (on the right) includes the keycloak service as well...

sashko9807:
> @sashko9807 sorry - it looks like the latest playwright.yaml (addition of keycloak) was ignored - is this a cache thing? Or did I miss something?
> 
> image
> 
> Notice the command (on the left) was `docker-compose up -d pg-db` while the update (on the right) includes the keycloak service as well... 

Probably the keycloak step should be added and merged from a separate commit first.

 Though I am not sure whether it would work, because the API container still requires a secret to connect with keycloak. Will have to check whether we can preconfigure it from docker compose first.

slavcho (approved):

slavcho (approved):

PR #1883, opened 2024-07-10
Update 'See more' button on Active campaigns (Homepage)

ani-kalpachka:
Updated 'See all' button on Active campaigns (Homepage) as per the new design:

image


swolf86:
@ani-kalpachka  button text to be "see more" виж още - "All" is not correct

ani-kalpachka:
Thanks @swolf86!
It is updated:
image

sashko9807 (approved):

PR #1891, opened 2024-07-22
Fix id after choosing a campaign (Admin panel)

ani-kalpachka:
Fixed id after choosing a campain in recurring donations in Admin panel:

Before|After
---|---
image|image

The id is set in handleCampaignSelected function when the user chooses a campaign. The issue was caused by setting an initail value on sourceVault when the id is already present, so I reset it to '' in the object.

### DIFF
--- a/src/components/admin/recurring-donation/Form.tsx
+++ b/src/components/admin/recurring-donation/Form.tsx
@@ -74,7 +74,7 @@ export default function EditForm() {
       extCustomerId: data?.extCustomerId,
       money: fromMoney(data?.amount || 0),
       currency: data?.currency,
-      sourceVault: data?.sourceVault.id,
+      sourceVault: '',
       campaign: data?.sourceVault.campaign.id,
     }
   }


slavcho (commented):
I don't understand how this fixes the issue?
Is the "data.sourceVault.Id" incorrect?

What if we open the form, but do not change the campaign? Don't we end up without a sourceVault set to "" if we click save without changing the campaign?

ani-kalpachka:
As I understand it, sourceVault: `data?.sourceVault.id `sets the inital value if the id is present.
The same id is set for the second time when a campaign is chosen from the dropdown in this block of code:
 ```
handleCampaignSelected={(campaign) => {
    campaigns?.forEach((c) => {
        if (c.id === campaign && c.vaults) {
            setFieldValue('sourceVault', c.vaults[0].id)
         }
    })
}}
```
I thought we dont need the first initail setting of the value and it should be handled in the handleCampaignSelected() function.
This approach fixes the issue and seems not to cause unwanted side effects. It may be better approuches though. If you have some concerns and other better suggestions how to fix the issue, please advice.

Issue #1912, opened 2024-09-02
When show error message on support page


Tuscann:
### Describe the Bug

When show error message on support page

### To Reproduce

1. Open https://podkrepi.bg/en/support
2. Choose company and click button next
3. Choose company sponsor and click button next
4. **Click button Submit ** without add any information
5. Show error only for Given name

![image](https://github.com/user-attachments/assets/f0952eb9-4405-407f-8476-ff4d3ce8604c)

6. **Click button Submit**
7. Show error for all fields
![image](https://github.com/user-attachments/assets/a57f0835-4508-47b0-9217-c1ab11167689)

**Why first time show only one error but next time show all errors?**

 

### Expected Behavior

Show errors on first time click Submit

### Which browser are you using? (if relevant)

Vivaldi

velnachev:
Could I please get this issue assigned to me?

PR #1961, opened 2024-10-17
bugfix/1912-Support-Page-Forms-Validation

velnachev:
Implemented dynamic form validation for each step of the support form

Closes [#1912](https://github.com/podkrepi-bg/frontend/issues/1912)

## Motivation and context

On blur validation has been disabled due to an issue caused by the initial focus on the first input field of a form step

Logic is that each step will validated on change only when it was submitted with an error from the validation schema for that step. If a user goes back to a previous step and tries to submit it again with an error that logic will again be triggered

### Steps to test

1. Navigate to https://podkrepi.bg/en/support
2. Try to submit each step without selecting/filling anything
3. Observe validation behavior for each step
4. Go back to previous steps and try to submit without selecting/filling anything and observe validation behavior

---|---|---|---


gparlakov:
@velnachev hi. This "ValidationObserver" pattern does not seem to have been used before in the project. Help me understand where it comes from and what makes it the right fit for the project that uses `formik`

velnachev:
> @velnachev hi. This "ValidationObserver" pattern does not seem to have been used before in the project. Help me understand where it comes from and what makes it the right fit for the project that uses `formik`

Hi! The pattern has not been used in the project since I've developed it to solve the particular issue and improve the UX of the form flow by displaying error states and triggering the onChange validation of the form only once an erroneous submission attempt has been made for the particular step, instead of it being immediately triggered once a field is touched. 

Formik is certainly being used, and the role of the ValidationObserver component is to simply gain access to the formik context, so the logic of tracking submission attempts for each separate step in a state variable can be achieved, which is then used in toggling the onChange validation of formik on and off.

If this is not a desired behavior - we can just do validateOnBlur={false} on the form and the described issue would be solved.

src/components/client/support-form/SupportForm.tsx
```diff
@@ -222,21 +222,22 @@ export default function SupportForm() {
...
-              <StepLabel error={isStepFailed(index)} StepIconComponent={StepIcon}>
-                {t(step.label)}
-              </StepLabel>
-            </Step>
-          ))}
-        </Stepper>
-      </Hidden>
+      innerRef={formRef}
+      validateOnBlur={false}
+      validateOnChange={currentValidatedField === activeStep}>
```
gparlakov, line 227: `currentValidateField` sounds like it pertains to a field, but in here it is used to toggle on/off of individual field validation on a step. 

How about `validateInidividualFieldsOnStep` ?

src/components/client/support-form/ValidationObserver.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useEffect, useRef } from 'react'
+import { Steps } from './helpers/support-form.types'
+
+const STEPS_VALIDATION_MAP = {
+  [Steps.ROLES]: ['roles'],
+  [Steps.QUESTIONS]: ['benefactor', 'partner', 'volunteer', 'associationMember', 'company'],
+  [Steps.PERSON]: ['person'],
+}
+
+const ValidationObserver = ({
```
gparlakov, line 11: Let's add some comment stating the need to turn off individual field validation first time the step is submitted but then turning it back on so the user gets immediate feedback when filling the correct/expected value

gparlakov: And finally - this new pattern might benefit greatly from a few unit tests. Benefit in terms of readability.

src/components/client/support-form/SupportForm.tsx
gparlakov: I'd suggest to make the pattern more explicit and readable by making the value that comes out of the ValidationObserver a map such that individual field validation is turned on/off explicitly 
```tsx 
<GeneralForm<...>
      ...
      validateOnBlur={validateIndividualFields[activeStep]}
      validateOnChange={validateIndividualFields[activeStep]}>
      
```

gparlakov (commented):
Thank you for your response,

I believe I understand the intent now after reading the comment a few times and also trying to fix the issue using alternative means. So now that I understand the code I have a few suggestions that only aim at making the logic more readable and understandable. 

See them in the individual files' comment

Issue #1960, opened 2024-10-15
Donation bugs on dev env


Tuscann:
### Which area(s) of Podkrepi.bg are affected?

Desktop layout

### Describe the Bug

Bugs

### To Reproduce

1. Missing dot
![image](https://github.com/user-attachments/assets/7ac79aef-f313-43d8-95cd-76436acdf87f)

2. Why this rectangle is smaller ?
![image](https://github.com/user-attachments/assets/5c908567-5b53-4453-9b82-f6c508b81f6c)

3. What is meaning of this green tick? 
![image](https://github.com/user-attachments/assets/734733d8-805a-40f3-bed3-2588f564f66f)

4,  1.Click button Donate
      2. Fill all data without amount money
     
     Why is not red color?
     
![image](https://github.com/user-attachments/assets/a0f149ba-cc6c-420a-85b0-1f8680b57f78)

5.  1. Select BGN 20.00
      2. Click back button
      3. On new screen click close button.
      4. **What this button close need to do ?**    (Currently after click close return to campaign not close this new screen)
      
      
![image](https://github.com/user-attachments/assets/2c6a0ac7-83d0-4335-8226-249678683d46)

6. Missing "имейл" after donation but see
![image](https://github.com/user-attachments/assets/c71870f2-7858-4b88-baec-f33eb0cd58c8)

7. Not working shareable link to Facebook

https://youtu.be/VJWwGqe1qKw

8. What is Field suggestion ?

![image](https://github.com/user-attachments/assets/0f96dd17-69db-4c7d-a738-25f39dd5969a)

9. Status code 400 if  I want to be anonymous is not checked ?

![image](https://github.com/user-attachments/assets/7dc27686-5887-4185-a572-281a6527c95a)

https://youtu.be/-LS6TrlymLo

10. Add fullstop on end sentance

![image](https://github.com/user-attachments/assets/fed7a978-c1e8-43d9-8765-8e0956c8b104)

11. Why show this twice?

![image](https://github.com/user-attachments/assets/4a3b8c44-73ec-4074-9fc8-5a4c962a6956)

12. Why open https://dev.podkrepi.bg/privacy-policy not https://dev.podkrepi.bg/en/privacy-policy
when using english.

![image](https://github.com/user-attachments/assets/8a37c8a6-1e1d-4737-b48e-9a664082e812)

13. Why show errors messages on Bulgarian when choosen language is English?

[...]

PR #1962, opened 2024-10-19
Bugfix/new donation flow issues

nikolay-yankov:
## Motivation and context
Fixes several issues from ticket #1960 related to the new donation flow, discovered during QA testing.

## Screenshots:

Issue #|Before|After
-|---|---
1|image|image
2|image|image
3|Cannot reproduce - message sent to QA for details| N/A
4|image|imageimage
5|image| Yes, it should work exactly the opposite way. Now when the user selects the `X` button or `Cancel` button, the user remains on the same page. If the user selects `Continue` button, then it will be redirected to the Campaign preview page.
10|image|image
11|image|image
12|image|image

### Additional change:
IMO we should not disable the `Amount` field when bank payment option is selected as this creates several side effect. IMO we should either hide it completely when Bank options is chosen or leave it always available. This is because currently, if the user selects an amount and then selects bank payment type, the `Amount` field becomes disabled. This could create some confusion because the user cannot change the amount anymore and they cannot use the calculator for amount + fees at the bottom. Also if validation is triggered and the field, they cannot select an option to continue with the donation.
Here I removed the option for disabling in commit https://github.com/podkrepi-bg/frontend/commit/2ceeef55f2826e3ac9e43e81b25c8d28021b1d9e

### DIFF
--- a/src/components/client/donation-flow/DonationFlowForm.tsx
+++ b/src/components/client/donation-flow/DonationFlowForm.tsx
@@ -226,15 +226,15 @@ export function DonationFlowForm() {
               <ConfirmationDialog
                 isOpen={showCancelDialog}
                 handleCancel={() => {
-                  cancelSetupIntentMutation.mutate({ id: setupIntent.id })
-                  router.push(routes.campaigns.viewCampaignBySlug(campaign.slug))
+                  setShowCancelDialog(false)
                 }}
                 title={t('cancel-dialog.title')}
                 content={t('cancel-dialog.content')}
                 confirmButtonLabel={t('cancel-dialog.btn-continue')}
                 cancelButtonLabel={t('cancel-dialog.btn-cancel')}
                 handleConfirm={() => {
-                  setShowCancelDialog(false)
+                  cancelSetupIntentMutation.mutate({ id: setupIntent.id })
+                  router.push(routes.campaigns.viewCampaignBySlug(campaign.slug))
                 }}
               />
               <Button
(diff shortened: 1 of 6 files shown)


src/components/client/donation-flow/DonationFlowForm.tsx
```diff
@@ -247,10 +247,7 @@ export function DonationFlowForm() {
               </Button>
               <Grid2 container size={{ xs: 12, md: 7 }} direction={'column'}>
                 <StepSplitter content="1" active={Boolean(values.amountChosen)} />
-                <Amount
-                  disabled={values.payment === DonationFormPaymentMethod.BANK}
-                  error={Boolean(errors.finalAmount) && Boolean(submitCount > 0)}
-                />
+                <Amount error={Boolean(errors.finalAmount) && Boolean(submitCount > 0)} />
```
sashko9807, line 250: Can we please revert this change for now? 
I understand the reasoning for this, but since this Amount Picker has no effect on bank transactions, in my opinion it will cause more confusion to keep it enabled.
I've already contacted the design team to look at this. If there are any changes regarding this step, lets handle them in a separate PR.

sashko9807 (changes requested):

src/components/client/donation-flow/DonationFlowForm.tsx
nikolay-yankov: Yes, I just reverted this one. Also reverted the change in `Amount` validation to not take into account the bank type payment with this commit https://github.com/podkrepi-bg/frontend/pull/1962/commits/ef26ae382f30aa2ee2449d33615c90645c090128
Here is a visual what I meant that the user cannot select another amount when Bank is selected and the calculation at the bottom cannot be used if the user wants to see what the fees would be.
cc @swolf86

chrome_WnBgUR0833

swolf86: @nikolay-yankov 
I think the transactions fees should not be displayed it should be ---- until we have the integration with paysera. 

Why the amount cannot be changed, after selecting a bank transfer?
I saw that too

sashko9807 (approved):

PR #1976, opened 2024-11-16
fix: campaign application

gparlakov:
- fix order of fields (make the goal the last field on the 3-rd step)
- remove "Membership" from the list of campaign types (it's a system type)


sashko9807 (approved):
In my opinion it would be better to add some flag to the BE to manage campaignTypes visibility, so if a type is renamed for some reason, we won't need to make changes to FE. Though I suppose this would do for now

PR #1989, opened 2024-12-04
fix: Donation flow more fixes

sashko9807:
Addresses bugs reported by #1960


src/components/common/SocialShareListButton.tsx
```diff
@@ -25,8 +25,9 @@ export default function SocialShareListButton({
 }) {
   const { t } = useTranslation('common')
   const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(null)
-  const serializedUrl = new URLSearchParams(url).toString()
+
   const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    console.log(event.currentTarget)
```
gparlakov, line 30: probably dont need that one

src/components/common/SocialShareListButton.tsx
```diff
@@ -56,26 +57,27 @@ export default function SocialShareListButton({
           <ListItemButton
             onClick={() => {
               navigator.clipboard.writeText(url)
+              console.log(url)
```
gparlakov, line 60: console log again - ditto

gparlakov (approved):
lgtm - just a couple of console.log-s that need to go

We should probably add a eslint rule to prune these

PR #1990, opened 2024-12-06
Lint out console.log-s

gparlakov:
This eslint rule will help catch console.log-s and similar in the codd


sashko9807 (changes requested):
Lets warn for now, instead of error-ing it.

sashko9807:
Closing due to inactivity.

PR #1991, opened 2024-12-06
Donation flow more fixes

sashko9807:


src/components/client/donation-flow/steps/payment-method/TaxesCheckbox.tsx
```diff
@@ -50,15 +50,17 @@ export const TaxesCheckbox = () => {
...
-      <Trans
-        t={t}
-        i18nKey="step.payment-method.alert.calculated-fees"
-        values={{
-          amount: moneyPublicDecimals2(amountWithoutFees.value),
-          fees: moneyPublicDecimals2(amountWithFees.value - amountWithoutFees.value),
-          totalAmount: moneyPublicDecimals2(amountWithFees.value),
-        }}
-      />
+      {!!amountWithFees.value && (
```
gparlakov, line 53: Is this to assertain that the value is a non-zero number?  As in `!isNaN(amountWithFees.value) && amountWithFees.value > 0` 

It's hard to read. And also negative numbers give true value as well => `!!-1 === true`

gparlakov (approved):
LGTM. I've one question - see in comments

gparlakov:
Yes I get that. What I'm saying is 1 it is unreadable, and 2 it will show
the value when it's negative number or a string or an object etc

PR #2024, opened 2025-04-19
2022 align text icon

tsanev96:
image


tsanev96:
@ani-kalpachka Actually do we want this centered to the text or just the icon to be aligned with the text on the first row only?

sashko9807 (approved):

PR #2049, opened 2025-07-24
improve container build to reduce image size

imilchev:
Changes the way we build the frontend container so it results in a smaller size image. The current one is 2.38GB, when building with this change it results in an image of 521MB. 

The changes were mostly AI generated


ani-kalpachka (approved):

sashko9807:
Closing this as similar thing was done in different PR.

